PR #17, opened 2020-07-17
fix(editors): add saveOutputType to finally have proper save format

ghiscoding:
- we now have 3 available types to use with the Date Editors formatting
1. `type` (data input format),
2. `outputType` (picker display format)
3. `saveOutputType` (save format)

### DIFF
--- a/packages/common/src/editors/dateEditor.ts
+++ b/packages/common/src/editors/dateEditor.ts
@@ -109,9 +109,7 @@ export class DateEditor implements Editor {
         dateFormat: inputFormat,
         closeOnSelect: false,
         locale: (currentLocale !== 'en') ? this.loadFlatpickrLocale(currentLocale) : 'en',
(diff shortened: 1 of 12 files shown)


packages/common/src/editors/dateEditor.ts
```diff
@@ -105,10 +106,10 @@ export class DateEditor implements Editor {
         defaultDate: this.defaultDate as string,
         altInput: true,
         altFormat: outputFormat,
-        dateFormat: outputFormat,
+        dateFormat: inputFormat,
         closeOnSelect: false,
         locale: (currentLocale !== 'en') ? this.loadFlatpickrLocale(currentLocale) : 'en',
-        onChange: (selectedDates: Date[] | Date, dateStr: string, instance: any) => {
```
ghiscoding, line 111: remove all arguments and make it a 1 liner

PR #74, opened 2020-08-12
fix(core): fix some problems found with AutoComplete

ghiscoding:


packages/common/src/interfaces/autocompleteOption.interface.ts
```diff
@@ -51,11 +51,11 @@ export interface AutocompleteOption {
...
-  focus?: (e: Event, ui: any) => boolean;
+  focus?: (e: Event, ui: { item: any; }) => boolean;
 
   /** Triggered when user enters a search value */
-  search?: (e: Event, ui: any) => boolean;
+  search?: (e: Event, ui: { item: any; }) => boolean;
 
   /** Triggered when a value is selected from the autocomplete list */
-  select?: (e: Event, ui: any) => boolean;
+  select?: (e: Event, ui: { item: any; }) => boolean;
```
ghiscoding, line 60: this should result `void`

PR #155, opened 2020-11-07
feat(core): add "Empty Data" warning message when grid is empty

ghiscoding:
- this is in fact a div element that is created and reposition inside the grid, the reason we do that is because we cannot put that text inside a cell because it has its cell boundary, so creating a div and position it over the grid is the best approach


packages/vanilla-bundle/src/components/slick-empty-warning.component.ts
```diff
@@ -0,0 +1,62 @@
...
+   * @param isShowing - are we showing the message?
+   * @param options - any styling options you'd like to pass like the text color
+   */
+  showEmptyDataMessage(gridSelector: string, isShowing = true, emptyMessage?: string, options?: EmptyWarning) {
+    const gridUid = this.grid.getUID();
+    const mergedOptions = { ...this.gridOptions.emptyDataWarning, options };
+    const emptyDataClassName = mergedOptions.class || 'slick-empty-data-warning';
+    const finalClassNames = [gridUid, emptyDataClassName];
+    this._warningElement = document.querySelector<HTMLDivElement>(`.${finalClassNames.join('.')}`);
+    const gridElm = document.querySelector<HTMLDivElement>(gridSelector);
```
ghiscoding, line 31: need to nullify `gridElm`

PR #227, opened 2021-01-05
fix(build): import Flatpickr Locale on demand via regular imports

ghiscoding:
- we shouldn't use require(locale) because they all end up being part of the final bundle, it's better to let the user import whichever Flatpickr Locale he wants to use and that will end up being a much smaller bundle with only the locale we really need

### DIFF
--- a/packages/common/src/editors/__tests__/dateEditor.spec.ts
+++ b/packages/common/src/editors/__tests__/dateEditor.spec.ts
@@ -434,10 +434,24 @@ describe('DateEditor', () => {
     });
 
     describe('with different locale', () => {
-      it('should display text in new locale', (done) => {
+      it('should display a console warning when locale it not previously imported', (done) => {
+        const consoleSpy = jest.spyOn(global.console, 'warn').mockReturnValue();
+
         gridOptionMock.translater = translateService;
 
-        translateService.use('fr-CA'); // will be trimmed to "fr"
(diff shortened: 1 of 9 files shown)


packages/common/src/editors/__tests__/dateEditor.spec.ts
```diff
@@ -434,10 +434,24 @@ describe('DateEditor', () => {
     });
 
     describe('with different locale', () => {
-      it('should display text in new locale', (done) => {
+      it('should display a console warning when locale it not previously imported', (done) => {
```
zewa666, line 437: Locale it(is)

zewa666:
This is definitely a better approach since you aint really gonna need all locales bundled all the time.

Tbh this actually is a breaking change, because users depending on a different language would now have to touch code as their CI/Tests might break. It feels relatively small but essentially it still is one ;) dunno If you just want to document this which might be enough given the en fallback

ghiscoding:
hmm yeah it might be a small breaking change but I certainly won't do a version 4.x lol
I will document it in Wikis and the Release, to be fair there's a warning message so they will eventually see what to do and there's probably a low number of developers using multiple locales.

zewa666:
Well you asked me and you got an answer ;)
I'm by god no overzealos semver fanatic so I'd tend to go with what you said that the benefit and clear/simple fix should be ok

ghiscoding:
haha I do appreciate your feedback since you're a big user of the lib 😉

PR #261, opened 2021-02-09
feat(perf): huge filtering execution speed improvements

ghiscoding:
The problem was that previous code was parsing the search values at the same time as processing the filtering check  (on every single row) and that is totally unnecessary, the search value once provided will never change when comparing to each cell. This PR is rewriting the Filter Service to process the search values only once (before looping through all cells). Parsing Dates are the biggest performance killer because we take dates, that could be in any format, and we parse (convert) them to MomentJS object and that is expensive on 50,000 or more rows (see below for perf logs before/after). 

### Avg Gain on a set of 50k rows
- single date filter: 6.1x faster
- date range filter: 4.4x faster

below is a quick test with perf logs performing date filtering before & after code change with 50k rows and 8 columns grid
_note: the date are all in ISO format (yyyy-mm-dd), there might be an even bigger perf boost with non-ISO format_

--- 
#### with a single date
searching: 1471.196044921875 ms
searching: 1438.469970703125 ms
searching: 1363.123046875 ms
avg: ~1424.26 ms

#### with a date range
searching: 2695.779052734375 ms
searching: 2751.52490234375 ms
searching: 2727.10888671875 ms
avg: ~2724.80 ms

#### with a single date
searching: 246.010009765625 ms
searching: 236.799072265625 ms
searching: 217.127197265625 ms
avg: ~233.31 ms

#### with a date range
searching: 591.408935546875 ms
searching: 695.953857421875 ms
searching: 585.989990234375 ms
avg: ~624.45 ms


gotstu:
Awesome work Ghislain!

PR #282, opened 2021-03-15
feat(filters): add new IN_COLLECTION operator to allow searching cell value as Array

ArnaudVanHerck:
We are having issues with our implementation of SlickGrid (through Aurelia-SlickGrid) where the search terms sometimes can contain whitespaces. When the testFilterCondition function runs with the operator 'inContains' the terms get split on whitespaces. This issue got solved for us when using camelCasing or kebabCasing because this removes whitespaces from the search terms and cellValue before sending them to the testFilterCondition function. Because the 'toKebabCase' function needs a string and we know that multiple-select.js also provides a string, a toString() was added to the searchTerm without the need of worrying about possible type conficts (as far as my understanding of the code).


ghiscoding:
hmm I'm not sure I like this approach, it seems the goal of this is simply to remove white spaces, why don't we fix the source instead? Because if I look at your code change, I'm quite sure this will have a perf impact and I'm not willing to add that for that reason. Where do you see that it gets split by white space? Can't we just add a trim somewhere?

ArnaudVanHerck:
Yes, I understand your concern about the performance hit. I just assumed you'd prefer me reusing existing functions. 😃
From my understanding the problem occurs when using a multiselect filter on an column that contains an array of strings, each string can contain white spaces. The strings get comma-separated as expected but the array of selected options seem to get whitespace separated. When these values get send to the testFilterCondition function the 'inContains' operator splits the search string coming from the multiselect filter on white spaces. This is where our issue lies. Our search string, e.g. Group ABC, gets parted in Group and ABC. This makes that is will never get a hit

ghiscoding:
Are you using the `IN` or the `IN_CONTAINS` operator? Because I did a change recently (PR #262) on that latter operator because it wasn't working as it should have. 
https://github.com/ghiscoding/slickgrid-universal/blob/e44145d897da570bf6ea15b156c7961ce96ce6f0/packages/common/src/filter-conditions/filterUtilities.ts#L74-L78

The `IN_CONTAINS` does a split on comma and it expects the search terms to have comma but typically the `multiple-select.js` uses the `IN` or `NOT_IN` operators. I rarely ever use this `IN_CONTAINS` operator, the only time I used it in all of the demos was on this [Example 25](https://ghiscoding.github.io/aurelia-slickgrid/#/slickgrid/example25) and I don't think I ever used it with the `multiple-select`

What are you using that with? Is that a local grid (JSON dataset) or is it with a Backend Service (OData/GraphQL)? It's not provided in your issue, and perhaps show a print screen of the issue

image

ArnaudVanHerck:
We're using SlickGrid with json (local grid) and we're using the `IN` operator on columns where the cells only contain a string and we're using the `IN_CONTAINS` operator on columns where the cells contain an array of strings.
When I test the code you quoted with a simple string with white space, the string gets split on the white space.
E.g. : 'Group ABC'.split(/[\s,]+/) becomes an array containing ['Group', 'ABC'] which is our issue right now.
I see now that the filter items from `multi-select.js` seem to come as an array of strings and that the `IN_CONTAINS` operator splits the cellValue not the select options (my bad, understood that part wrong).
So it seems that the issue is the REGEX used in the split. Because the cellValue from the array gets comma separated it shouldn't split on whitespaces. When I changes the REGEX to '.split(/[,]+/)' it seems to work as expected (which it seems you changed it to i PR #262 but that's not how it looks like in master).

ghiscoding:
hmm it seems that the reason you are using the `IN_CONTAINS` is not why I created this Operator, I mainly created it to do a split on comma (and/or spaces in some cases) on an input text filter, but in your use case you just want it to compare against an array. I'd be ok with removing the space from the regex but this will not be bullet proof because in your use case if your string as a comma then you'll be back to square one and have problems again because it will split again... so I think a better approach would be to simply add a new Operator, I'm thinking of `IN_COLLECTION` and that would be the same code as the `IN_CONTAINS` but without the split because your cell is already an array.
Something like this
```ts
case 'IN_CONTAINS':
      if (value2 && Array.isArray(value2) && typeof value1 === 'string') {
        return value2.some(item => value1.split(/[\s,]+/).includes(item));
      }
      return false;
case 'IN_COLLECTION':
      if (value1 && value2 && Array.isArray(value1) && Array.isArray(value2)) {
        return value2.some(item => value1.includes(item));
      }
      return false;
```

> When I changes the REGEX to '.split(/[,]+/)' it seems to work as expected (which it seems you changed it to in PR #262 but that's not how it looks like in master right now).

I'm not sure what you mean here? The code in that PR for that switch case is what is in master right now and it was shipped in last version `0.11.x`. If you don't have that code on your side it might be because you didn't update to the latest version? You can see what was shipped on the [releases](https://github.com/ghiscoding/slickgrid-universal/releases) page or in the [changelog](https://github.com/ghiscoding/slickgrid-universal/blob/master/CHANGELOG.md)

ArnaudVanHerck:
An new `IN_COLLECTION` operator could work. But then the cell value shouldn't be combined to one string as it does today in the `executeCollectionSearchFilterCondition` file. Do we need a new search filter condition then maybe too? Or are we adding a check to cell value to see if it's an array or not before sending it to the test function?

What I meant is that the commit you're referring to in PR #262 doesn't split on whitespaces just comma but that got added in PR #267 😄

ghiscoding:
Oh I didn't know it was using a string every time, do you mean this line (on the last part)?
https://github.com/ghiscoding/slickgrid-universal/blob/3174c86e011b4927510b99a348e8019adb4baa00/packages/common/src/filter-conditions/collectionSearchFilterCondition.ts#L10

I think I made it a string mainly to convert numbers to string, we probably need to find a better way so that it works with an array. Maybe just casting as a string when it's a number should be sufficient
```ts
const cellValue = (options.cellValue === undefined || options.cellValue === null) ? '' : typeof options.cellValue === 'number' ? `${options.cellValue}` : options.cellValue; 
```

> What I meant is that the commit you're referring to in PR #262 doesn't split on whitespaces just comma but that got added in PR #267

Thanks for the find and so I would rather keep that then because we use the Composite Editor extensively in our project and I don't want to create new bugs by changing the regex. So I think the `IN_COLLECTION` and `NOT_IN_COLLECTION` is probably the best way to go. 

Would you like to contribute and test it out with your code? If so, you'll have to add both Operators in the following 2 files [operatorType.enum.ts](https://github.com/ghiscoding/slickgrid-universal/blob/master/packages/common/src/enums/operatorType.enum.ts) and [operatorString.type.ts](https://github.com/ghiscoding/slickgrid-universal/blob/master/packages/common/src/enums/operatorString.type.ts) and modify the other files which you know about.

ArnaudVanHerck:
Yes, that's the code that makes it a combined string. I'm just wondering how other types get affected by the change in this string conversation you're proposing.
``const cellValue = (options.cellValue === undefined || options.cellValue === null) ? '' : typeof options.cellValue === 'number' ? `${options.cellValue}` : options.cellValue;``
In case we'd be looking for e.g. a boolean, would it need to be parsed to a string too?. Maybe some extra testing should done to make sure nothing else stops working?

Thank you for pointing me in the right direction, I'll try the proposed changes our project and update the PR if everything works as expected. 👍

ghiscoding:
ah yeah I forgot about the boolean values, it's probably better to include them as well because the `multiple-select.js` only works with string if I recall correctly and also when you type a character on a regular text input filter it will always be a string and I still want number/boolean to work and that is why that string casting is there. 

Maybe we can simply go the inverse and only check for arrays

```ts
const cellValue = (options.cellValue === undefined || options.cellValue === null) 
  ? '' 
  : Array.isArray(options.cellValue) ? options.cellValue : `${options.cellValue}`;
```
You might need to put it all in 1 line though because I think ESLint will complain if it's not

It's a good discussion, I'm always happy to have contributions ;)

ArnaudVanHerck:
I updated the PR now, and did the following changes:

- Added the InCollection and NotInCollection operator types
- Added these types to the isCollectionOperator function
- Changed the `executeCollectionSearchFilterCondition` function so that it doesn't make a string of the array but of the items in the array
- Added tests to check if the `executeCollectionSearchFilterCondition` function behaves as expected with types other than string.
- Added tests to check if the `testFilterCondition` behaves as expected

ghiscoding:
Wow that is pretty good, I got nothing to say on the PR, it's perfect 😄 
Have you had a chance to try it in your own project?

Were you done with the PR, I would be ready to merge as is.

EDIT

Actually the Example 7 Cypress E2E tests keeps failing on the last column "Prerequisites" which is a multiple-select dropdown. I'm not sure what makes it fail though, it could be the `Array.isArray(options.cellValue)` or the switch/case

ArnaudVanHerck:
Yes, I saw that 😢 
I'm trying to get cypress to work but it's fighting me a bit 😉

ghiscoding:
I'm not sure why Cypress dashboard doesn't me those tests but if I look at the GitHub Actions, the failing test is the following (which is the last test on 

```bash
1) should open the "Prerequisites" Filter then choose "Task 3", "Task 4" and "Task 8" from the list and expect to see 2 rows of data in the grid

  30 passing (33s)
  1 failing
```
The Cypress test code has the following steps, I suggest to try the steps manually on that Example 7
https://github.com/ghiscoding/slickgrid-universal/blob/3174c86e011b4927510b99a348e8019adb4baa00/test/cypress/integration/example07.spec.js#L587-L611

ArnaudVanHerck:
Yes, I'm experiencing a similar issue, I cannot seem to run the cypress tests locally.
Do you have the possibility to test the steps manually? I'm not sure how to get that part working either 😕

ghiscoding:
hmm I sent you the code for the Cypress tests, I thought it was readable but the title of the test should tell you what to do in a general way 
```ts
it('should open the "Prerequisites" Filter then choose "Task 3", "Task 4" and "Task 8" from the list and expect to see 2 rows of data in the grid')
```
Note that a few months back, we can no longer open Cypress directly from VSCode, it throws some kind of memory heap size error of some kind. It was reported on VSCode but they never really addressed it, if that is what you're doing then simply open a separate terminal and run cypress from that terminal (outside of VSCode) and that is how I got it to work.

ghiscoding:
@ArnaudVanHerck I saw you closed/reopened the PR (maybe to trigger another test) and it still fails on the exact same Cypress test, not sure what it is yet but I'll definitely take a look at it a bit later today. At least what we know so far is that there's really something on that test and it's not a flaky test because it's always the same one that fails.... to be continued ;)

ArnaudVanHerck:
@ghiscoding I tried to rename the branch in order for it to reflect the PR better but apparently that closed the PR 😋
I finally got the project running locally (I got stuck after apparently running `yarn run bootstrap` multiple times). I figured out what was the issues with the test and added a check on the operator when setting cellValue to a string. I'm running the tests now to see if everything still works as expected.

packages/common/src/filter-conditions/collectionSearchFilterCondition.ts
```diff
@@ -7,7 +7,7 @@ import { testFilterCondition } from './filterUtilities';
  */
 export const executeCollectionSearchFilterCondition: FilterCondition = (options: FilterConditionOption) => {
   // multiple-select will always return text, so we should make our cell values text as well
-  const cellValue = (options.cellValue === undefined || options.cellValue === null) ? '' : `${options.cellValue}`;
+  const cellValue = (options.cellValue === undefined || options.cellValue === null) ? '' : ((options.operator === 'IN_COLLECTION' || options.operator === 'NOT_IN_COLLECTION') && Array.isArray(options.cellValue)) ? options.cellValue.map(value => `${value}`) : `${options.cellValue}`;
```
ghiscoding, line 10: I find this line is starting to be way too lengthy, could we split this into multiple lines to increase readability. Actually, I rewrote the entire function and the unit tests are still good, could you please use the code below, I find it more readable. Thanks

```ts
export const executeCollectionSearchFilterCondition: FilterCondition = (options: FilterConditionOption) => {
  // multiple-select will always return text, so we should make our cell values text as well
  const filterOperator = options.operator;
  let cellValue: string | string[];
  if (Array.isArray(options.cellValue) && (filterOperator === 'IN_COLLECTION' || filterOperator === 'NOT_IN_COLLECTION')) {
    cellValue = options.cellValue.map(value => `${value}`);
  } else {
    cellValue = (options.cellValue === undefined || options.cellValue === null) ? '' : `${options.cellValue}`;
  }

  return testFilterCondition(filterOperator || 'IN', cellValue, options.searchTerms || []);
};
```

ghiscoding:
Alright, so I ran all the unit tests and I also wanted to test it out in Aurelia-Slickgrid before approving this PR and all the Aurelia-Slickgrid Cypress tests are passing (all 400 tests), so I'll be happy to merge after you do the code change I asked in previous comment, the rest is all great, nice PR. Thanks

packages/common/src/filter-conditions/collectionSearchFilterCondition.ts
ArnaudVanHerck: Haha, yes that line got quite long now. I'm happy to change it, I'll update it soon 👍

ArnaudVanHerck:
I added a nullcheck to the array given that the cellValue got typed as `string `or `string[]`. Otherwise it feels like, in case of an empty array, it would return `null` and not an empty array. 

```ts
export const executeCollectionSearchFilterCondition: FilterCondition = (options: FilterConditionOption) => {
  // multiple-select will always return text, so we should make our cell values text as well
  const filterOperator = options.operator;
  let cellValue: string | string[];
  if (Array.isArray(options.cellValue) && (filterOperator === 'IN_COLLECTION' || filterOperator === 'NOT_IN_COLLECTION')) {
    cellValue = (!!options.cellValue.length ? options.cellValue.map(value => `${value}`) : []);
  } else {
    cellValue = (options.cellValue === undefined || options.cellValue === null) ? '' : `${options.cellValue}`;
  }

  return testFilterCondition(filterOperator || 'IN', cellValue, options.searchTerms || []);
};
```

What do you think?

ghiscoding:
sure that'd be ok

ghiscoding:
Thanks for the great contribution, good PR, good feedback, all good 😄 
Expect a release by end of this week or early next week 🚀

ArnaudVanHerck:
Thank you for the feedback and your patience. And thank you for you hard work on these packages 🥇

ghiscoding:
@ArnaudVanHerck 
This is now live with a few other things, see the new Slickgrid-Universal [changelog](https://github.com/ghiscoding/slickgrid-universal/blob/master/CHANGELOG.md) and the new Aurelia-Slickgrid version [Release 3.5.0](https://github.com/ghiscoding/aurelia-slickgrid/releases)

Cheers ⭐  and thanks again for the contribution

PR #292, opened 2021-03-20
feat(state): add Pinning (frozen) to Grid State & Presets

ghiscoding:
#### TODOs
- [x] get Pinning from Grid State
- [x] use Pinning from grid options `presets`
- [x] add `clearPinning` and `setPinning`
- [x] add Cypress E2E tests to
  - [x]  save text search Filters with operators and redisplay them after changing Grid View dropdown
  - [x] save & redisplay Pinning after changing Grid View dropdown
- [x] rename grid menu to "Unfreeze Columns/Rows" and in French "Dégeler les colonnes/rangées"


test/cypress/integration/example11.spec.js
```diff
@@ -561,4 +561,174 @@ describe('Example 11 - Batch Editing', () => {
...
+
+    cy.get('.slick-header-menu')
+      .should('be.visible')
+      .children('.slick-header-menuitem:nth-child(1)')
+      .children('.slick-header-menucontent')
+      .should('contain', 'Freeze Column')
+      .click();
+  });
+
+  it('should have a frozen grid with 4 containers on page load with 3 columns on the left and 6 columns on the right', () => {
```
ghiscoding, line 607: the test should say 3 left, 8 right

PR #342, opened 2021-05-17
fix(presets): loading columns presets should only be done once

ghiscoding:
- loading of Columns Grid Presets should only be done once (before resizing) and loading Filters Grid Presets could be done one or more times, so to fix this I have separated `loadPresetsWhenDatasetInitialized` into 2 new methods and made sure to call the columns grid preset only once.
- this issue was identified while porting the previous PR #341 into Angular-Slickgrid


AnnetteZhang (approved):
Thank you for adding detailed comments for this PR and I like it!

PR #352, opened 2021-05-26
fix(demo): we should be able to move row(s) and keep selections

ghiscoding:
- grid state should only be allowed to change row selection only when Pagination is enabled
- refactor some part of the code to use more optional chaining (?.) syntax for cleaner and smaller code
- move around some of the Cypress tests so that it makes sure that moving rows will keep row selections

### DIFF
--- a/packages/common/src/services/gridState.service.ts
+++ b/packages/common/src/services/gridState.service.ts
@@ -257,8 +257,8 @@ export class GridStateService {
    * @return current pagination state
    */
   getCurrentPagination(): CurrentPagination | null {
-    if (this._gridOptions.enablePagination) {
-      if (this._gridOptions?.backendServiceApi) {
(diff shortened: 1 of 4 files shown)


packages/common/src/services/gridState.service.ts
```diff
@@ -258,9 +258,9 @@ export class GridStateService {
    */
   getCurrentPagination(): CurrentPagination | null {
     if (this._gridOptions.enablePagination) {
```
AnnetteZhang, line 260: looks like _gridOptions can be null sometimes. If you check null on line 260 with "?.", you can remove the "?." on line 261.

AnnetteZhang (changes requested):
One minor comment

AnnetteZhang (approved):

PR #381, opened 2021-06-15
feat(footer): add option to customize right footer text

ghiscoding:
- add option to provide optional right footer text
- also reworked the Custom Footer DOM tree to remove an extra `<div>` that wasn't necessary, we can just use the same right footer div to show/hide the metrics

### DIFF
--- a/packages/vanilla-bundle/src/components/__tests__/slick-footer.spec.ts
+++ b/packages/vanilla-bundle/src/components/__tests__/slick-footer.spec.ts
@@ -1,3 +1,4 @@
+import 'jest-extended';
 import { CustomFooterOption, GridOption, SlickGrid } from '@slickgrid-universal/common';
 import { SlickFooterComponent } from '../slick-footer.component';
 import { TranslateServiceStub } from '../../../../../test/translateServiceStub';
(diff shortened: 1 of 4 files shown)


packages/vanilla-bundle/src/components/__tests__/slick-footer.spec.ts
```diff
@@ -85,7 +86,7 @@ describe('Slick-Footer Component', () => {
 
       const footerContainerElm = document.querySelector('div.slick-custom-footer.slickgrid_123456') as HTMLDivElement;
       const leftFooterElm = document.querySelector('div.slick-custom-footer.slickgrid_123456 > div.left-footer') as HTMLSpanElement;
-      const rightFooterElm = document.querySelector('div.slick-custom-footer.slickgrid_123456 > div.metrics') as HTMLSpanElement;
+      const rightFooterElm = document.querySelector('div.slick-custom-footer.slickgrid_123456 > div.right-footer') as HTMLSpanElement;
```
AnnetteZhang, line 89: Should the lookup be 'div.right-footer.metrics'?

AnnetteZhang (commented):
I have one comment on the test, but looks like the tests passed anyway.

packages/vanilla-bundle/src/components/__tests__/slick-footer.spec.ts
ghiscoding: no not in this case because the `metrics` css class name is only added if it's not defined as hidden, that was refactored in this PR is different compare to previous code.

Since this unit test is to hide it, it won't add css class because of this code
```ts
// ...
else if (!this.customFooterOptions.hideMetrics) {
      rightFooterElm.classList.add('metrics');
```

AnnetteZhang: Thank you for the clarifications. It makes senses now.

PR #403, opened 2021-07-02
feat(services): move Resizer Service w/common services folder for reuse

ghiscoding:
- for reusability purposes, we can move the Resizer Service from the vanilla bundle into the common Services folder
- also move all the Resizer code into the Resizer Service instead of passing by the external SlickGrid/Slick.Resizer 3rd party plugin


packages/vanilla-bundle/src/components/__tests__/slick-vanilla-grid.spec.ts
```diff
@@ -51,7 +51,7 @@ import { TextExportService } from '../../services/textExport.service';
 import { TranslateServiceStub } from '../../../../../test/translateServiceStub';
 import { HttpStub } from '../../../../../test/httpClientStub';
 import { MockSlickEvent, MockSlickEventHandler } from '../../../../../test/mockSlickEvent';
-import { ResizerService } from '../../services/resizer.service';
+import { ResizerService } from '../../../../common/src/services/resizer.service';
```
ghiscoding, line 54: this should be coming from `common` package of Slickgrid-Universal instead

AnnetteZhang (approved):

PR #407, opened 2021-07-13
fix(events): use nullish coalescing in slickgrid event prefix

ghiscoding:


packages/vanilla-bundle/src/components/slick-vanilla-grid-bundle.ts
```diff
@@ -5,7 +5,6 @@ import 'slickgrid/slick.core';
 import 'slickgrid/slick.grid';
 import 'slickgrid/slick.dataview';
 import 'slickgrid/slick.groupitemmetadataprovider';
-import 'slickgrid/plugins/slick.resizer';
```
ghiscoding, line 8: this is no longer used, it should have been removed in previous PR from last week

AnnetteZhang (approved):

PR #483, opened 2021-09-13
feat(sorts): option to ignore accent while sorting text

mcallegario:
@ghiscoding 

add a Grid Option to sort text by ignoring any accent by normalizing the text, the option is set to false by default

### DIFF
--- a/packages/common/src/interfaces/gridOption.interface.ts
+++ b/packages/common/src/interfaces/gridOption.interface.ts
@@ -462,6 +462,12 @@ export interface GridOption {
   ignoreAccentOnStringFilter?: boolean;
 
   /**
+   * Defaults to false, should we ignore any accent while sorting text?
+   * For example if our text is "José" and we type "Jose" then it won't return unless we use this flag because "é" is not equal to "e"
+   */
+  ignoreAccentOnStringSort?: boolean;
+
+  /**
    * When using custom Locales (that is when user is NOT using a Translate Service, this property does nothing when used with Translate Service),
    * This is useful so that every component of the lib knows the locale.
    * For example, not providing this will make the Date Filter/Editor use English by default even if we use different "locales",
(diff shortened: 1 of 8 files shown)


packages/common/src/interfaces/gridOption.interface.ts
```diff
@@ -461,6 +461,12 @@ export interface GridOption {
    */
   ignoreAccentOnStringFilter?: boolean;
 
+  /**
+   * Defaults to false, should we ignore any accent while sorting text?
+   * For example if our text is "José" and we type "Jose" then it won't return unless we use this flag because "é" is not equal to "e"
+   */
+  ignoreAccentOnStringSort?: boolean;
```
ghiscoding, line 468: hmm I think we should probably rename the previous flag and keep only 1, maybe `ignoreAccentOnStringFilterAndSort`

ghiscoding:
We're also missing the test unit coverage, I want to keep my 100% coverage, the test is located under the `__tests__` folder [here](https://github.com/ghiscoding/slickgrid-universal/blob/master/packages/common/src/sortComparers/__tests__/stringSortComparer.spec.ts). Do you think you add a Jest test to cover that? The info on how to run them are in the [readme](https://github.com/ghiscoding/slickgrid-universal#tests)

mcallegario:
I made the changes you suggested, make sure it's ok, please.

ghiscoding:
@mcallegario thanks for the update, it seems all good now. Have you had a chance to test it? Also I'm not expecting any release in a while, there isn't much more commits done, the easiest way if you want to use it (and test it) would be to run the `bundle` npm script and then go under the `/packages/common/dist` and copy the `dist` and overwrite your `node_modules/@slickgrid-universal/common` of your project

ghiscoding:
@mcallegario 
This is now released under the new version [0.18.0](https://github.com/ghiscoding/slickgrid-universal/releases/tag/v0.18.0)
Thanks again for the contribution 🚀

PR #489, opened 2021-09-21
fix(resizer): only bind autoresize when enabled

jr01:
This fixes the bug/exception mentioned here https://github.com/ghiscoding/Angular-Slickgrid/issues/836 and is "PR 1".

Writing this comment I realize that just the `clearTimeout` would have been enough to fix the exception, but since `bindAutoResizeDataGrid` has this comment `Bind an auto resize trigger on the datagrid, if that is enable then it will resize itself to the available space` I think the added check on `this.gridOptions.enableAutoResize` should be done in `init` and is good.  Also when not using auto resize `resizeGrid` is called from `bindResizeHook` in Vanilla, Angular and Aurelia, so this change shouldn't cause noticeable side effects.

### DIFF
--- a/packages/common/src/services/resizer.service.ts
+++ b/packages/common/src/services/resizer.service.ts
@@ -117,7 +117,7 @@ export class ResizerService {
       this._fixedWidth = fixedGridSizes.width;
     }
 
-    if (this.gridOptions) {
+    if (this.gridOptions && this.gridOptions.enableAutoResize) {
       this.bindAutoResizeDataGrid();
     }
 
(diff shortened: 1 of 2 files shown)


packages/common/src/services/resizer.service.ts
```diff
@@ -251,7 +252,7 @@ export class ResizerService {
 
   /**
    * Provide the possibility to pause the resizer for some time, until user decides to re-enabled it later if he wish to.
-   * @param {boolean} isResizePaused are we pausing the resizer?
+   * @param {boolean} isResizePaused are we pausing the resizer?@C
```
ghiscoding, line 255: is that `@C` a typo?

jr01: Nice catch. No idea how that got there...but I'll fix it.

packages/common/src/services/resizer.service.ts
```diff
@@ -117,7 +118,7 @@ export class ResizerService {
       this._fixedWidth = fixedGridSizes.width;
     }
 
-    if (this.gridOptions) {
+    if (this.gridOptions && this.gridOptions.enableAutoResize) {
```
ghiscoding, line 121: but isn't this make the grid to not resize even with static sizes? 
Have you tried running the code in this lib project? 
We need to make sure that even if `enableAutoResize` is disabled, this won't break grids with static sizes like the Example 1 in this webpack demo.

PR #490, opened 2021-09-21
feat(resizer): add option to detect container resize

jr01:
This PR builds on top of https://github.com/ghiscoding/slickgrid-universal/pull/489 and adds the `ResizeObserver` mentioned here https://github.com/ghiscoding/Angular-Slickgrid/issues/836 and is the "PR 3" mentioned there.

This PR adds the `gridOptions.autoResize.resizeDetection`. It's `'window'` by default, when set to `'container'` it observes the set `gridOptions.autoResize.container` element for resize events. When a resize event is observed it resizes the grid with the available height/width based on that container's dimensions.

The `gridOptions.autoResize.container` got changed from `string` to `string | HtmlElement` so that it's possible to provide an element directly which saves me in my project for having to create a unique id. I think this change is backwards compatible. 

Changed example-14 in the Slickgrid universal demo vanilla to use the `ResizeObserver`. I used a simple approach for letting the (resize) container div take all available space in the y-axis.

BTW I tested from my project by:
* creating a symlink:
`C:\repos\MyProject\src\node_modules\@slickgrid-universal\common>mklink /J dist C:\repos\slickgrid-universal\packages\common\dist`
* running a build watch from @slickgrid-universal/common: `C:\repos\slickgrid-universal>lerna run build:watch --stream`
* and then running `ng serve` / `ng test` from my project.

 I first tried the `npm link` route, but couldn't get that to work. Would love to hear if there is a better solution than symlinks...

### DIFF
--- a/packages/common/src/interfaces/resizerOption.interface.ts
+++ b/packages/common/src/interfaces/resizerOption.interface.ts
@@ -2,10 +2,7 @@ export interface ResizerOption {
   /** Defaults to false, do we want to apply the resized dimentions to the grid container as well? */
   applyResizeToContainer?: boolean;
 
-  /**
-   * Defaults to 'window', which DOM element are we using to calculate the available size for the grid?
-   * When {@link useResizeObserver}=true the {@link container} is used and this option is ignored.
-   */
+  /** Defaults to 'window', which DOM element are we using to calculate the available size for the grid? */
   calculateAvailableSizeBy?: 'container' | 'window';
 
   /** bottom padding of the grid in pixels */
@@ -41,16 +38,12 @@ export interface ResizerOption {
   rightPadding?: number;
 
   /**
-   * Use a [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) to detect resizes on the {@link container} element instead of window resize events.
+   * Defaults to 'window', how are resizes detected?
    *
-   * Remarks:
+   * When set to 'container':
    * * Requires {@link container} to be set.
-   *
-   * * ResizeObserver is not supported on older browsers like Internet Explorer. If you need the support you need to install a
-   * polyfill like [resize-observer-polyfill](https://www.npmjs.com/package/resize-observer-polyfill) yourself.
-   *
    * * If you get 'ResizeObserver loop limit exceeded' errors in automated tests take a look
-   * [here](https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded).
+   *   [here](https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded).
    */
-  useResizeObserver?: boolean;
+  resizeDetection?: 'container' | 'window';
 }
--- a/packages/common/src/interfaces/resizerOption.interface.ts
+++ b/packages/common/src/interfaces/resizerOption.interface.ts
@@ -2,14 +2,20 @@ export interface ResizerOption {
   /** Defaults to false, do we want to apply the resized dimentions to the grid container as well? */
   applyResizeToContainer?: boolean;
 
-  /** Defaults to 'window', which DOM element are we using to calculate the available size for the grid? */
+  /**
+   * Defaults to 'window', which DOM element are we using to calculate the available size for the grid?
+   * When {@link useResizeObserver}=true the {@link container} is used and this option is ignored.
+   */
   calculateAvailableSizeBy?: 'container' | 'window';
 
   /** bottom padding of the grid in pixels */
   bottomPadding?: number;
 
-  /** Page Container selector, for example '.page-container' or '#page-container', basically what element in the page will be used to calculate the available space */
-  container?: string;
+  /**
+   * Page Container. Either selector (for example '.page-container' or '#page-container'), or an HTMLElement.
+   * Basically what element in the page will be used to calculate the available space.
+   */
+  container?: string | HTMLElement;
 
   /**
    * Grid Container selector, for example '.myGrid' or '#myGrid', this is provided by the lib internally.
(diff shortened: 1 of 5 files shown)


packages/common/src/interfaces/resizerOption.interface.ts
```diff
@@ -33,4 +39,18 @@ export interface ResizerOption {
...
+   * Remarks:
+   * * Requires {@link container} to be set.
+   *
+   * * ResizeObserver is not supported on older browsers like Internet Explorer. If you need the support you need to install a
+   * polyfill like [resize-observer-polyfill](https://www.npmjs.com/package/resize-observer-polyfill) yourself.
+   *
+   * * If you get 'ResizeObserver loop limit exceeded' errors in automated tests take a look
+   * [here](https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded).
+   */
+  useResizeObserver?: boolean;
```
ghiscoding, line 55: is this resize observer only here because of legacy browser? If that is the only reason, I would prefer to drop the flag and simply make the Resize Observer the default because I'm already no longer supporting legacy browser, not even IE11 because I'm bundling for ES2018 (as shown in [tsconfig.json](https://github.com/ghiscoding/slickgrid-universal/blob/master/tsconfig.json) and so IE11 and lower are excluded. So if this Resize Observer is really better than the window resize event, I would prefer to only keep that one and get rid of the old window event. What do you think?

ghiscoding:
after running the project, I see that all the grids go outside their container, not sure what happened there but it's a regression

image

packages/common/src/services/resizer.service.ts
```diff
@@ -196,22 +223,25 @@ export class ResizerService {
...
       const coordOffsetTop = this._gridDomElm.offset();
-      gridOffsetTop = (coordOffsetTop !== undefined) ? coordOffsetTop.top : 0;
+      const gridOffsetTop = (coordOffsetTop !== undefined) ? coordOffsetTop.top : 0;
+      gridHeight -= gridOffsetTop;
     }
 
-    const availableHeight = gridHeight - gridOffsetTop - bottomPadding;
-    const availableWidth = this._pageContainerElm.width() || window.innerWidth || 0;
+    const availableHeight = gridHeight - bottomPadding;
+    const availableWidth = gridWidth;
```
ghiscoding, line 244: the new `width` seems to be problematic as can be seen in the print screen

ghiscoding (changes requested):

packages/common/src/interfaces/resizerOption.interface.ts
jr01: The reason for using a `ResizeObserver` is that it detects resizes on a specific element it observes (which I set to be the `resizerOption.container`). Whereas the window resize only detects entire window resizes. So for example if I just change the container's width (and not the window width) the window resize wouldn't be triggered (and jQuery wouldn't detect a resize) , but `ResizeObserver` does.

I'll change `useResizeObserver` to `resizeDetection?: 'container' | 'window'`,  similar to `calculateAvailableSizeBy`, that should make it more clear.
And I'll remove the comment about old IE and polyfill.

When `resizerOption.container` is set, probably `resizeDetection='container'` would be a good default, but I don't want to change current behavior and break things.

packages/common/src/services/resizer.service.ts
jr01: My bad, sorry!

I think I got a little too enthusiastic with removing jQuery here without understanding exactly how jQuery determines width/height. This change was working fine in example-14 and in my application, but I didn't check all the other examples. 
I now looked into how jQuery determines width/height ... and it takes padding etc into account and has some more subtleties ... but in hindsight I don't really need this change for detecting resizes on elements, so I'll revert the code.

packages/common/src/services/resizer.service.ts
```diff
@@ -110,7 +115,13 @@ export class ResizerService {
 
     const containerNode = grid?.getContainerNode?.() as HTMLDivElement;
     this._gridDomElm = $(containerNode);
-    this._pageContainerElm = $(this._autoResizeOptions.container!);
+
+    if (typeof this._autoResizeOptions.container === 'string') {
+      this._pageContainerElm = $(this._autoResizeOptions.container);
+    } else {
+      this._pageContainerElm = $(this._autoResizeOptions.container!);
+    }
```
jr01, line 123: I don't like this `if` here, but the typing on `$` doesn't support a parameter with `X | Y`.

packages/common/src/interfaces/resizerOption.interface.ts
ghiscoding: yeah as long as we don't have regression I'm fine with that, I was just thinking in a way that might end up with less lines of code.

ghiscoding:
@jr01 
I changed the title to "WIP" so that I don't merge the PR by mistake, just rename it and remove that word once you're done

jr01:
Hey @ghiscoding, I'm done with this PR I think. Let me know if it needs more work.

ghiscoding:
@jr01 
I tried in our Salesforce environment without changing anything and I got an error thrown saying `ResizeObserver is not a constructor`. I didn't investigate neither change any settings in my environment. Would you have any clue why this happen?

jr01:
@ghiscoding That's interesting. What browser/version is used in that 'environment' ? https://www.caniuse.com/resizeobserver

But regardless, I've just pushed a commit where I moved the `new ResizeObserver...` to `bindAutoResizeDataGrid`, so it's only called when `resizeDetection='container'`

ghiscoding:
I'm using the new Edge browser (Chromium) at work, so it shouldn't matter but even then I wouldn't want it to work only for some users. It might be the DTS bundle I do with [DtsBundleWebpack](https://github.com/ghiscoding/slickgrid-universal/blob/master/packages/vanilla-bundle/webpack.config.js#L45), but anyway just a flag to disable it should be sufficient (like you did in your last PR), I'll give that a try on Monday on the work laptop. It seems you're working late, kinda like me lol ;)

ghiscoding (approved):

ghiscoding:
seems all good now, merging

PR #513, opened 2021-10-01
fix(sorts) option to ignore accent while sorting the numeric text 

mcallegario:
@ghiscoding 

Identifies a bug in the routine that ignores the accent when sorting the numeric text, in this case it generates an error as shown in the images below:

image

image

image

### DIFF
--- a/packages/common/src/services/utilities.ts
+++ b/packages/common/src/services/utilities.ts
@@ -883,7 +883,7 @@ export function parseUtcDate(inputDateString: any, useUtc?: boolean): string {
  * @returns
  */
 export function removeAccentFromText(text: string, shouldLowerCase = false) {
-  const normalizedText = text.normalize('NFD').replace(/[\u0300-\u036f]/g, '');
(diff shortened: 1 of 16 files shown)


packages/common/src/services/utilities.ts
```diff
@@ -883,7 +883,7 @@ export function parseUtcDate(inputDateString: any, useUtc?: boolean): string {
  * @returns
  */
 export function removeAccentFromText(text: string, shouldLowerCase = false) {
-  const normalizedText = text.normalize('NFD').replace(/[\u0300-\u036f]/g, '');
+  const normalizedText = text.normalize?text.normalize('NFD').replace(/[\u0300-\u036f]/g, ''):text;
```
ghiscoding, line 886: I think it's incorrect syntax isn't it? I think it should always include a dot `?.` instead of `?`

mcallegario (commented):
redid the code, see if it's better this way.

packages/common/src/services/utilities.ts
mcallegario: redid the code, see if it's better this way.
image

ghiscoding:
yeah that's better thanks

PR #569, opened 2021-12-08
feat(filters): change-filter-element-Container

mcallegario:
Dear @ghiscoding 
Here is my suggestion for solving problem  #512, please check and make your comments as soon as possible.
My solution was to create a new method that would create the filter on a given container (HTML DOM element).
Check mainly the "test" routine, as I don't know this type of configuration.


ghiscoding:
@mcallegario that's interesting, I have a few questions before I start reviewing (I'm in the middle of a PR so I can't test your code on my local until I finish my own PR). 
- could you maybe provide a print screen of what that looks like in the UI ?
- ideally it would be better to either modify an existing Example or create a new one to demo to show everyone but also to add the possibility to add Cypress E2E tests to avoid future breaking changes
   - if you don't have time to do that, perhaps provide some piece of code so that I understand how to live test it

Also note that I already merged a few breaking change PRs (mostly all about styling) and so even if I merge your code, you will only be able to use it with the upcoming major `1.0.0` release (I'm expecting to do that by the end of this week or next Monday). It would be nice to include your PR but we could also release it under version `1.1.0` later in January. Anyway, I just wanted to let you know that you'll have to upgrade to use this code, it won't be released under `0.x.x` anymore.

packages/common/src/services/__tests__/filter.service.spec.ts
```diff
@@ -1611,6 +1611,47 @@ describe('FilterService', () => {
...
+      gridStub.onHeaderRowCellRendered.notify(mockArgs2 as any, new Slick.EventData(), gridStub);
+      await service.updateFilters(mockNewFilters);
+
+      const filterContainerElm:HTMLDivElement = document.querySelector('#'+DOM_ELEMENT_ID);
+      const output1 = service.drawFilterTemplate('firstName','#'+DOM_ELEMENT_ID);
+      const output2 = service.drawFilterTemplate('isActive',filterContainerElm);
+      const output3 = service.drawFilterTemplate('selector',filterContainerElm);
+      const output4 = service.drawFilterTemplate('name',filterContainerElm);
+            
+      //expect(output).toBe(output.columnDef.id);
```
ghiscoding, line 1651: so I see that you're not actually testing anything (you probably get good coverage because you're calling the function but you're not testing anything though), we'll have to fix that. We can probably query the DOM for that element and expect something to be rendered on that filter element, something like `document.body.querySelector('#'+DOM_ELEMENT_ID);` (it would also be better to use string template literal with backtick ` `` ` which is more readable, instead of string concatenation with +)

mcallegario: Exactly, I tried to make the test routine observing the other routines, as I had never done this before, I was only able to execute the call for the coverage test, but I couldn't implement the return analysis routine.
I'm really going to need help with this.

mcallegario:
@ghiscoding 
As for the implementation I did, I tried to develop something that solves the problem of the filter's external interface, with minimal changes to the original code.
For that, I created a new method called "drawFilterTemplate" (I accept suggestion for the name) in the class "filter.service.ts" which was a "copy" of "addFilterTemplateToHeaderRow". The new method takes two parameters (Column ID and HTML element) and returns the filterMetadata object of the created filter, moreover it creates the HTML DOM element inside the HTML element received as parameter.
The biggest change I made to the existing methods was to remove from the classes of each filter the routine that takes the HTML DOM element from the filter column "getHeaderRowColumn" and put it in the "addFilterTemplateToHeaderRow" routine, passing the HTML DOM element as parameter (filterArguments) to the classes of each type of filter.
Basically that was it.

As soon as possible (time) I will change an example to carry out the implementation, as requested.

packages/common/src/services/__tests__/filter.service.spec.ts
ghiscoding: just try to get the DOM element from the body with what I wrote in previous paragraph and then at least check that it's truthy, something like this `expect(filterElm).toBeTruthy();`, just try to follow the other tests, copy a test and add whatever you need to it

ghiscoding:
Yes I really need to see a live demo, else I can't test it out and I prefer to merge something that I can have unit tests and E2E tests (Cypress). I'm aiming for releasing the next major version `1.0.0` today, so we'll look at your PR to go in a future version `1.1.0`

mcallegario:
@ghiscoding 
Ok, I implemented in "Example 7" the modal filter feature.
image

image

ghiscoding:
@mcallegario 
Oh nice I'll check that a bit later today, maybe 1 comment would be to maybe add a Close button for the modal and maybe a Search button!? I assume the search is already active when you start typing in these filters, is that correct? If so, we could still add the Search button that would just close the modal with these search filters, it's less confusing for the user, even if we know it's not needed it's still better as a UI perspective, we could also have the Close button to reset the search filters before closing the modal.

examples/webpack-demo-vanilla-bundle/src/examples/example07.html
```diff
@@ -54,6 +54,10 @@ <h3 class="title is-3">
     <span class="icon mdi mdi-close"></span>
     <span>Clear Filters</span>
   </button>
+  <button class="button is-small" data-test="clear-filters-btn" onclick.delegate="allFilters()">
```
ghiscoding, line 57: you need to change the `data-test="clear-filters-btn"` to something else, this is used in the Cypress E2E tests and this is what makes Cypress fail now. I guess `data-test="modal-filters-btn"` would be ok

mcallegario:
Correct, the search is already active when you start typing these filters.
Done. I've already implemented the buttons.
image

examples/webpack-demo-vanilla-bundle/src/examples/example07.ts
```diff
@@ -315,6 +315,73 @@ export class Example7 {
     this.sgb.filterService.clearFilters();
   }
 
+  
+  allFilters() {
+    const grid = this.sgb;
+    $(`<div id="modal-allFilter" class="modal is-active" >
```
ghiscoding, line 321: hey I just saw this while relooking at this new code sample, and that worries me a bit. I know it's for demo purpose but I'm not a fan of using that kind of html mostly because it's not sanitized and you're using jQuery to do it. You can maybe change it to something like 
```ts
import * as DOMPurify from 'dompurify';

const modalHtml =`<div id="modal-allFilter" class="modal is-active" >...</div>`;
document.body.appendChild(DOMPurify.sanitize(modalHtml));
```
this would be better since it's both sanitized and isn't using jQuery. We could also create a new package in the future for this, similar to what I've done with [Composite Editor](https://github.com/ghiscoding/slickgrid-universal/tree/master/packages/composite-editor-component) and use similar CSS Classes so that it works the same in all framework, it would also be in plain JS code like the Composite Editor one (but probably more as a future idea)

mcallegario: Ready! 
I added the DOMPurify.

packages/common/src/services/__tests__/filter.service.spec.ts
ghiscoding: hey @mcallegario so I looked at adding full unit test coverage, here's the lines you can add the following tests 

```ts
expect(filterContainerElm).toBeTruthy();
expect(output1).toBeTruthy();
expect(output2).toBeTruthy();
expect(output4).toBeTruthy();
```
Then I can see that the `output3` with `selector` won't return anything because that is not a valid Column name, so perhaps we can use that for another test. We can start by making it more obvious that it's not a valid column name, let's rename `selector` to `invalid-column` with this line
```ts
const output3 = service.drawFilterTemplate('invalid-column', filterContainerElm);
```
and add test to check for null
```ts
expect(output3).toBeNull();
```
but then we're still missing a test when passing a string as the second argument, so we can do something like this
```ts
const output5 = service.drawFilterTemplate('firstName', '.not-exists');
expect(output5).toBeNull();
```
... and there you go full 100% unit test coverage 😉

examples/webpack-demo-vanilla-bundle/src/examples/example07.ts
```diff
@@ -315,6 +315,76 @@ export class Example7 {
...
+        $('#modal-allFilter-table').append(`
+        <div class="row slick-headerrow-columns">
+          <div class="column">${columnFilter.name}</div><div id="${filterElm}" class="column ui-state-default slick-headerrow-column"></div>
+        </div>
+        `);
+        grid?.filterService.drawFilterTemplate(columnFilter,`#${filterElm}`);
+      }
+    }
+    if (grid?.slickGrid.getOptions().showHeaderRow){
+      grid?.filterService.toggleHeaderFilterRow();
```
ghiscoding, line 383: I was testing the UI and I could see the Filter bar being toggled (hidden) in the background because of this line, are there any particular reason you're doing this? or is it mostly for your use case? I'm just curious to know if it's related to how it works or not.

packages/common/src/services/__tests__/filter.service.spec.ts
mcallegario: Ready!
I made some adjustments to the test routine.

ghiscoding:
Your code styling is a bit off but I can fix that after, are you using Visual Studio Code? If so you can enable this setting "Format on Save", if you're not then no big deal and I'll fix it later. So it's getting closer to be merged, can you just maybe answer this question I asked [here](https://github.com/ghiscoding/slickgrid-universal/pull/569#discussion_r768275654) before proceeding with the merge.

examples/webpack-demo-vanilla-bundle/src/examples/example07.ts
mcallegario: Yes, there is a reason for this, synchronization of the two filter inputs (modal/title column).
When I type a filter in the modal, for performance reasons, I do not update the column filter, if in this case it was visible the same filter field would have different values, the update (synchronization) only occurs after closing the modal when it is displayed again the filter line.

mcallegario:
Yes, i use Visual Studio Code

examples/webpack-demo-vanilla-bundle/src/examples/example07.ts
ghiscoding: oh ok I see and that works because looking at the code I can see that calling `showHeaderRow` will recreate all the columns and their filters because it triggers this `onHeaderRowCellRendered` event which re-renders the filters. That's probably the only option you have then... or actually I think if you call `this.sgb.showHeaderRow(true)` it would probably work too without having to toggle the bar at all, you can give that a try and see if it makes any difference

ghiscoding: so I gave this a try and it works, probably better than toggling, I replaced all 3 toggle code with `this.sgb?.showHeaderRow(true);` (you'll have to replace `function` with fat arrow `() =>`. 

The end result is the same, calling the toggle or the `showHeaderRow` will both call a re-render of all filters after closing the modal which will keep filter values in sync. So did you want to give that a try or do you want me to merge as it is?

ghiscoding:
@mcallegario 
I'm not sure if you saw my [last comment](https://github.com/ghiscoding/slickgrid-universal/pull/569#discussion_r769164343), let me know if you want me to merge or if you still wanted to test a few things.

mcallegario:
Oops! I haven't really read the last two comment.
I'll make your suggestion to set the "showHeaderRow".

mcallegario:
@ghiscoding 
About merge, my suggestion is to do with this implementation and in the future perform another implementation to incorporate the modal, as is currently done with the modal editor.

ghiscoding:
@mcallegario 
This is now released, thanks a lot for another great contribution 😉 
[Release 1.2.0](https://github.com/ghiscoding/slickgrid-universal/releases/tag/v1.2.0)

PR #582, opened 2021-12-17
fix(demo): latest change with Filter container breaks other demos

ghiscoding:
- after opening Example 7 all filters modal and then going to another Example breaks the multiple-select filter (probably other filters too). The issue was caused by the DOMPurify option of `RETURN_DOM` which was caused by the creation of a second `body` within the current `body`
- also nit pick on code linting/styling

### DIFF
--- a/examples/webpack-demo-vanilla-bundle/src/examples/example07.ts
+++ b/examples/webpack-demo-vanilla-bundle/src/examples/example07.ts
@@ -10,9 +10,11 @@ import {
 } from '@slickgrid-universal/common';
 import { ExcelExportService } from '@slickgrid-universal/excel-export';
 import { Slicker, SlickVanillaGridBundle } from '@slickgrid-universal/vanilla-bundle';
-import { TranslateService } from '../translate.service';
 import * as DOMPurify from 'dompurify';
+
+import { TranslateService } from '../translate.service';
 import { ExampleGridOptions } from './example-grid-options';
+import './example07.scss';
 
 export class Example7 {
   private _bindingEventService: BindingEventService;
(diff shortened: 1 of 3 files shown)


examples/webpack-demo-vanilla-bundle/src/examples/example07.ts
```diff
@@ -318,69 +320,56 @@ export class Example7 {
...
+        </section>
+        <footer class="modal-card-foot">
+          <button class="button btn-close">Close</button>
+          <button id="btn-clear-all" class="button">Clear All Filter</button>
+          <button class="button btn-close is-success">Search</button>
+        </footer>
+      </div>
     </div>`;
 
-    document.body.appendChild(DOMPurify.sanitize(modalHtml, { RETURN_DOM: true }));
```
ghiscoding, line 359: @mcallegario I think you used `RETURN_DOM` because `appendChild` didn't accept the element otherwise but please note that it was causing issues and I had to remove that and I had to revert to using jQuery instead for appending to the real body, see animated gif included

mcallegario: Oops! I'm very sorry!
I didn't know (and didn't even imagine) that DOMPurify adds an extra BODY to the html document.

mcallegario: I use slickgrid-universal on a page within Sharepoint. I started having problem designing the selection filter menu after these code changes were made:
image

To work 100% and I have to always comment out these lines(170,226) which calls the method "subscribeToOnHeaderRowCellRendered".

Otherwise, it starts to happen:
image

ghiscoding: that code is there for a reason, to delete all filters and avoid leaks, it's not a good idea to remove it and is meant to destroy/remove the filter. Did you really remove the DOM Sanitize `RETURN_DOM`? I can't really help with a repro

mcallegario: So, in the original version, this problem occurred after the commit of 08-29-2020 "[fix(filters): remove Filters from DOM after header row gets destroyed](https://github.com/ghiscoding/slickgrid-universal/commit/b08d4ba070ec9d9d131d6830e4625e6ef950ac09)"
But don't worry, I believe it's something specific to my Sharepoint environment (webpart) with vanilla version, unfortunately I didn't have free time to investigate the problem further (debug)

ghiscoding: that code change was because Filters were not being removed, they became orphans in the DOM. Every time I was clicking on a select dropdown filter, it was creating a new dropdown filter without removing it from the DOM after closing it, so the more I click the filter, the more orphan dropdown were being created, so it was a big memory leak, not visible to the user but visible in the DOM. The Multiple/Single Filter is special, it create and append the dropdown to the `body` every time you open it and when you close the dropdown, it destroys it (removes it) from the DOM, in your use case it doesn't seem to find the dropdown (perhaps because of the latest filter PR you made, I'm not sure)

PR #588, opened 2021-12-23
feat(odata): $select and $expand

jr01:
As discussed here: https://github.com/ghiscoding/slickgrid-universal/discussions/580

This feature adds support for `$select` and `$expand` in the oData query. 

It can be enabled by:
```typescript
  const oDataOptions: OdataOption = {
     enableSelect: true;
     enableExpand: true;
  }
```

With `oDataOptions.enableSelect=true` only the grid column's fields are queried from the backend. For example `columns: [{ id: 'col1', field: 'field1' }, { id: 'col2', field: 'field2' }]` results in the query: `?$select=id,field1,field2`. 

Note that `id` is always added to `$select`. The name can be changed by setting `gridOptions.datasetIdPropertyName`.

With `oDataOptions.enableExpand===true` navigations are queried from the backend. A navigation is identified by having a field with `'/'` in it's name. For example `columns: [{ id: 'col1', field: 'nav1/field1' }, { id: 'col2', field: 'nav2/field1' }]` results in the query  `?$expand=nav1,nav2`. 

When both `oDataOptions.enableSelect===true` and `oDataOptions.enableExpand===true` and `oDataOptions.version===4` we also select only the navigation fields from the backend. For example `columns: [{ id: 'col1', field: 'nav1/field1' }, { id: 'col2', field: 'nav2/field1' }]` results in the query  `?Select=id,$expand=nav1($select=field1),nav2($select=field2)`. 

Navigations within navigations are also supported. For example `columns: [{ id: 'col1', field: 'nav1/subnav1/field1' }]`.

The dataset from the backend is extracted and navigation fields are flattened so the grid can display them and sort/filter just work. So for example if the backend responds with `{ value: [{ id: 1, nav1: { field1: 'x' }, { nav2: { field2: 'y' } } ] }` we'll flatten this to `{ value: [{ id: 1, 'nav1/field1': 'x', 'nav2/field2': 'y' } ] }`
The exact property that is used for the dataset depends on the oData version specified with `oDataOptions.version` and some defaults are set, if needed it can be overriden by setting/implementing the `oDataOptions.datasetExtractor` option.

Another addition is when `oDataOptions.enableCount=true` we now extract the count from the backend's response and automatically set `pagination.totalItems`. Also here there is a `oDataOptions.countExtractor` option to implement if needed.


packages/odata/src/interfaces/odataOption.interface.ts
```diff
@@ -7,6 +7,21 @@ export interface OdataOption extends BackendServiceOption {
...
   enableCount?: boolean;
 
+  /**
+   * Query fields using $select. The row identifier field is always added.
+   * E.g.: columns [{ field: 'date' }] results in $select=id,date
+   */
+  enableSelect?: boolean;
+
+  /** Which field to use as the row identifier when using enableSelect */
+  rowId?: string;
```
ghiscoding, line 17: I'm not sure about this name here, I find it's too much related to a row and maybe not enough to the data. Can it be `idProperty`, `idPropName`, `dataId`, `entityId` or anything else

packages/odata/src/services/odataQueryBuilder.service.ts
```diff
@@ -7,6 +7,11 @@ export class OdataQueryBuilderService {
   _filterCount = 0;
   _odataOptions: Partial<OdataOption>;
 
+  private _columnDefinitions: Column[] = [];
```
ghiscoding, line 10: I using more `protected` than `private` nowadays so it can be extended (that was actually a request from you last year lol)

ghiscoding:
@jr01 
I also sent you a collaborator invite for both libs so that you could update the [OData Wiki](https://github.com/ghiscoding/slickgrid-universal/wiki/OData) with these new features. Thanks

packages/odata/src/services/grid-odata.service.ts
```diff
@@ -94,12 +94,70 @@ export class GridOdataService implements BackendService {
...
+
+    this._odataService.columnDefinitions = this._columnDefinitions;
   }
 
   buildQuery(): string {
     return this._odataService.buildQuery();
   }
 
+  postProcess(processResult: any): void {
+    if (!processResult) {
```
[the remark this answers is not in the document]
jr01: Actually, thinking about it. It can't be null/undefined at this point because of the check in `backendUtility.service.ts`:
```typescript
// allow the backend service to change the result.
    if (processResult && backendApi.service.postProcess) {
      backendApi.service.postProcess(processResult);
    }
```

So I'll just remove the check here.

packages/odata/src/services/grid-odata.service.ts
```diff
@@ -94,12 +94,70 @@ export class GridOdataService implements BackendService {
...
+      return;
+    }
+    const odataVersion = this._odataService?.options?.version ?? 2;
+
+    if (this.pagination && this._odataService?.options?.enableCount) {
+      const countExtractor = this._odataService?.options?.countExtractor ??
+        odataVersion >= 4 ? (r: any) => r['@odata.count'] :
```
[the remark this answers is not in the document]
jr01: good point! I'll fix.

packages/odata/src/interfaces/odataOption.interface.ts
jr01: Yes this name is not good. It made me think why I needed it in the first place:
- if different entities don't use the same approach, then a single `rowId` setting doesn't work. E.g. `customer.customerId` 
 and `order.orderId`.
- a `rowId` setting doesn't work if an entity has a compound key.
- this is really specific from my code base where all entities have an `id` column. I can also just add `id` as a hidden column to my grids.

So I'll remove it.

jr01:
Hey @ghiscoding, I have done the rework and accepted the invites. 
And yes I will update the OData wiki. I guess it's best to wait until after you merge & release a new version?

test/cypress/integration/example09.spec.js
```diff
@@ -832,7 +832,7 @@ describe('Example 09 - OData Grid', { retries: 1 }, () => {
 
       cy.get('[data-test=odata-query-result]')
         .should(($span) => {
-          expect($span.text()).to.eq(`$top=10&$orderby=Name desc&$select=name,gender,company,id&$expand=category($select=name)`);
+          expect($span.text()).to.eq(`$top=10&$orderby=Name desc&$select=name,gender,company&$expand=category($select=name)`);
```
ghiscoding, line 835: I see you removed the `rowId` and `id` is no longer part of the `$select`, but will you still get a unique `id` in your dataset? Because you know that SlickGrid will throw an error if you don't have an `id` defined in your data

ghiscoding:
Just 1 last comment to verify before merging, thanks for all your great contributions, it is greatly appreciated. 
Happy Holidays to you and your family and also a big thanks for the coffee 🎄 🎁 🎅

test/cypress/integration/example09.spec.js
jr01: Yes, I mentioned that here https://github.com/ghiscoding/slickgrid-universal/pull/588#discussion_r774921405 

So the user (me) needs to add a column with a field `id`.  That column can be hidden of course if he doesn't want to have it in the UI. 

But it does not have to be `id` and can be anything like `customerId`, or multiple fields and if so in his `backendServiceApi.postProcess` the user can change what the backend send and make sure each item gets an `id`.

ghiscoding: hmm in GraphQL Service I add it automatically, so I still wonder if it wouldn't be better to do it in here as well, see below. 
If you still think it's unnecessary then let me know and I'll merge as it is. 

Thanks

https://github.com/ghiscoding/slickgrid-universal/blob/37c6ef68e802e634f2ed2ef70800f753a4b2d8eb/packages/graphql/src/services/graphql.service.ts#L72

https://github.com/ghiscoding/slickgrid-universal/blob/37c6ef68e802e634f2ed2ef70800f753a4b2d8eb/packages/graphql/src/services/graphql.service.ts#L109-L113

jr01: Aha. I didn't know about `gridOptions.datasetIdPropertyName`. Then it's probably smart to align the odata/graphql implementations. 

I'm not sure how to best forward the `grid` or `gridOptions` or perhaps just `datasetIdPropertyName` to `OdataQueryBuilderService`?  I'm leaning towards adding `init(grid) { ... }` to `OdataQueryBuilderService`. Do you have some guidance here? :-)

ghiscoding: yeah `gridOptions.datasetIdPropertyName` is what you use when you want something else than `id`, I'm not sure if I have it anywhere in the Wikis but it's in a few examples. The `OdataQueryBuilderService` wasn't meant to know about the grid, all it knows and has access is the `odataOption`, it's probably better to keep it that way, so I think you can just add a new option into the `odataOption` and you can let the user provide it and/or use `datasetIdPropertyName` when not. Below I'm using `entityId` but it could be something else `dataId`, `datasetIdProp`, ... most people won't fill that property and it will automatically fall to `gridOptions.datasetIdPropertyName` (that grid option is initialized in global options)

https://github.com/ghiscoding/slickgrid-universal/blob/37c6ef68e802e634f2ed2ef70800f753a4b2d8eb/packages/common/src/global-grid-options.ts#L101

```ts
init(serviceOptions?: Partial<OdataOption>, pagination?: Pagination, grid?: SlickGrid, sharedService?: SharedService): void {
    this._grid = grid;
    const mergedOptions = { ...this.defaultOptions, ...serviceOptions };
    this._datasetIdPropName = serviceOptions?.entityId ?? this._gridOptions.datasetIdPropertyName ?? 'id';
    // ...
}
```

ghiscoding: BTW, have you already upgraded to Angular-Slickgrid 4.0.0 that I released 2 weeks ago? Because that would fall under that version only

jr01: OK. I now forward the `_gridOptions.datasetIdPropertyName` to the `OdataQueryBuilderService`. I didn't extend the `odataOptions` because that doesn't add much I think and it can always be done later.

And yes, we're on v4!

Can you merge? :-)

ghiscoding: yes I can, awesome work 👍

Issue #604, opened 2022-01-13
Range filter .. no longer works


jr01:
### Your Environment
| Software               | Version(s) |
| --------------------- | ---------- |
| Slickgrid-Universal  | 1.2        |
| TypeScript               | 4.4        |

### Describe the Bug
Searchterm `a..e` no longer queries the data from a to e.

### Steps to Reproduce
- https://ghiscoding.github.io/slickgrid-universal/#/example09
- type a..e
- notice `$filter=(substringof('a..e', Name))`  expected is `$filter=(Name ge a and Name le e)`

### Expected Behavior
A search term containing `..` should be split into 2 parts and a range query should be executed with those parts.

### Current Behavior
A query is executed with a literal `a..e` contains (for a string field).

### Possible Solution
It seems search terms are still parsed correctly, for example [here](https://github.com/ghiscoding/slickgrid-universal/blob/master/packages/common/src/filter-conditions/stringFilterCondition.ts#L50), but the parsed values are not used later on. 

I'm trying to find out where and how this got broken, but haven't been successful yet :(

### Code Sample
Not needed I think.

ghiscoding:
@jr01 do you know if it was working with previous 1.1.0 version? The only PR that might have an impact might be this PR #569

ghiscoding:
The code that you pointed out which is in the `executeStringFilterCondition` will only work for local grids (without Backend Service) and so that is why that is never called. 

The problem really seem to be in the OData Service, I'm not sure if it was ever implemented correctly!?

Looking at this code, it does the split of `..` but it won't override the `operator` since it's already set to `Contains` and with that it executes the regular text contains filtering. 
https://github.com/ghiscoding/slickgrid-universal/blob/586311533a11a7bc747efd2fa610ea36c633dc09/packages/odata/src/services/grid-odata.service.ts#L351-L361

So was that ever working? Shouldn't the split of `..` automatically override the `operator` to a range operator?

jr01:
Ah yes that's the problem! 💯

I'm pretty sure it worked a year ago. Perhaps `columnFilter.operator` was `undefined` to begin with. But it doesn't really matter. 

Anyway, yes, I'll create a PR and definitely it will get a cypress test for this!

PR #608, opened 2022-01-18
fix(odata): fix range filtering with ".."

jr01:
Fixes https://github.com/ghiscoding/slickgrid-universal/issues/604 for the odata backend.


ghiscoding:
@jr01 
both libs got update with patch release. 
Cheers

PR #609, opened 2022-01-18
fix(graphql): fix range filtering with ".."

ghiscoding:
- range filter was no longer working for GraphQL backend Service


jr01:
looks good 👍

PR #642, opened 2022-02-15
fix(utilities): check if the normalize function exists

mcallegario:
dear @ghiscoding ,

I verified that after the compilation of the "slickgrid-vanilla-bundle.js" version, the line of code responsible for checking the existence of the "normalize" function is generated with wrong logic causing an error:
2022-02-14 08_49_47-DevTools - ingty sharepoint com_sites_pwa_dev_Project%20Detail%20Pages_DashProj

Original Code:
image

Generated code:
image

Code generated after correction:
image

### DIFF
--- a/packages/common/src/services/utilities.ts
+++ b/packages/common/src/services/utilities.ts
@@ -917,7 +917,7 @@ export function parseUtcDate(inputDateString: any, useUtc?: boolean): string {
  * @returns
  */
 export function removeAccentFromText(text: string, shouldLowerCase = false) {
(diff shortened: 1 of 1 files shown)


ghiscoding:
ok thanks, looks good to me. 
note that I released a new version yesterday so I'm not sure when I'll release this one though

ghiscoding:
@mcallegario this should be released now, I think since I had an issue with auto-release but I think it's still release, try it out. Cheers and thanks again

PR #737, opened 2022-08-09
fix(collectionAsync): hidden column does not load edit field selection

mcallegario:
Bug: Hidden column does not load edit field with selection options

The "this.slickGrid.getColumns()" method only returns the visible fields of the grid, so if the field that was configured to return a list of elements (collectionAsync) is hidden in the grid, it does not receive the list of elements.


packages/vanilla-bundle/src/components/slick-vanilla-grid-bundle.ts
```diff
@@ -1434,7 +1434,7 @@ export class SlickVanillaGridBundle {
 
     if (this.slickGrid) {
       // find the new column reference pointer & re-assign the new editor to the internalColumnEditor
-      const columns = this.slickGrid.getColumns();
+      const columns = this._columnDefinitions;
```
ghiscoding, line 1437: can you use the GETTER instead (without the `_` prefix) and there's no use in assigning a new variable from one to the other, just use it directly.
```ts
if (Array.isArray(this.columnDefinitions)) {
  const columnRef = this.columnDefinitions.find((col: Column) => col.id === column.id);
 //...
```

ghiscoding: You also have a lot of unit tests failing with this change, can you please fix them

ghiscoding:
@mcallegario 
Do you need help with the PR? I can fix the unit tests later if you have problem with them

PR #739, opened 2022-08-10
fix(common): duplicate translation namespace prefix, fixes #738

### LINKED ISSUE #738: partial translation with custom i18n namespase
### Describe the bug

with custom namespace some items in menu not translated - duplicate namespace in key string

@slickgrid-universal\common\dist\esm\extensions\extensionUtility.js -> translateMenuItemsFromTitleKey(items)

this method add translationPrefix to item.titleKey before translate, but titleKey already contains translationPrefix:

@slickgrid-universal\common\dist\esm\extensions\slickGridMenu.js -> addGridMenuCustomCommands(originalCommandItems)

fix translateMenuItemsFromTitleKey to remove duplicate translationPrefix, or check, that the titleKey not already starts with translationPrefix 

### Reproduction

1. set gridOptions.translationNamespace = 'aurelia-slickgrid'
2. load i18n translations for grid to this namespace

### Environment Info

```shell
aurelia 1
@slickgrid-universal/common@1.3.7
typescript none
browsers any
os any
```

### Validations

someusersomeuser:


ghiscoding:
I took a quick look at this change and where this function is called, so it might fix most of them but that would probably break the Cell Menu because that seems to be the only one that doesn't have predefined commands and so we cannot add the prefix. The other problem that this will potentially break would be when the user adds a custom command for any of these plugins as for example Grid Menu, then the command will not get the prefix... so I think that the only solution is probably the inverse which would be to remove leave the prefix in this method but remove it everywhere else (Grid Menu, Context Menu, etc...). I know this would require more code change but is still probably better because we still want the prefix to be added even on custom menu commands

someusersomeuser:
I think this is the correct behaviour. if the user adds their own menu item, then the translation will be in the default space, if it is not, he can explicitly specify a different one.

ghiscoding:
I kinda disagree, if there's a translation prefix then it should all or nothing. I don't think requiring the user to explicit add the prefix when he already added it in the i18n is the correct way, again I think it should be all or nothing and I would prefer to go with the last post I wrote

someusersomeuser:
in this case user need copy and patch standart locale files to own project after each version upgrade.
otherwise, user add custom items in default locale file 'translation.json', or for example 'aurelia-slickgrid-custom.json'

ghiscoding:
Like I said I never used these prefix and/or i18n workspace but I'm getting confused

Let say that we have a workspace prefix of `aurelia-slickgrid` and the user adds Grid Menu (or any other menu) commands like this

```ts
this.gridOptions = {
    gridMenu: {
      // custom commands
      commandItems: [
        { command: 'help', titleKey: 'HELP', positionOrder: 70, action: (e, args) => console.log(args) },
      ],
    },
}
```
and slickgrid-universal internal commands like this with the prefix
```ts
// internal commands
gridMenuCommandItems.push(
            {
              iconCssClass: this._gridMenuOptions.iconClearAllFiltersCommand || 'fa fa-filter text-danger',
              titleKey: `${translationPrefix}${commandLabels?.clearAllFiltersCommandKey ?? 'CLEAR_ALL_FILTERS'}`,
              // will be transformed into:  titleKey: `aurelia-slickgridCLEAR_ALL_FILTERS`
              disabled: false,
              command: commandName,
              positionOrder: 50
            }
);
```
this mean that you'll have the slickgrid internal translations with the `aurelia-slickgrid` prefix and the user custom translations without the prefix. 
```ts
allCommandItems = [
  // internal commands
  { command: 'help', titleKey: 'aurelia-slickgridCLEAR_ALL_FILTERS', positionOrder: 70, action: (e, args) => console.log(args) },

  // custom commands
  { command: 'help', titleKey: 'HELP', positionOrder: 70, action: (e, args) => console.log(args) },
}
```
Is that really what you're expecting? Because that wasn't at all my expectation of this prefix and conflicts with me definition of the translation prefix. I mean that I know you're trying to fix the i18n workspace but my conception of the prefix is again defined as an all or nothing

someusersomeuser:
yes.

``` javascript
  aurelia.use.plugin(PLATFORM.moduleName('aurelia-i18n'), (instance) => {
    const aliases = ['t', 'i18n'];
    TCustomAttribute.configureAliases(aliases);
    instance.i18next.use(Backend);
    const locale = getLocaleFromLanguage(navigator.language);
    return instance.setup({
      backend: { loadPath: './locales/{{lng}}/{{ns}}.json' },
      attributes: aliases,
      lng: 'en',
      ns: ['translation', 'aurelia-slickgrid'],
      defaultNS: 'translation',
      fallbackLng: 'en',
      debug: false
    });
```

 'aurelia-slickgrid' - from npm with standart keys as is
 'translation' - this is all user keys for application - buttons, messages, etc, including custom grid menu items. 
default namespace, prefix dont need

ghiscoding:
yeah ok I checked today and considering that I named the grid option as `translationNamespace`, then I guess your point of view is acceptable. I assume that you managed to test this? If not you can modify the file from `node_modules/@slickgrid-universal/common` or run the `bundle` command and replace the same node_modules folder

ghiscoding:
Thanks for the PR, this was released under new version [v1.4.0](https://github.com/ghiscoding/slickgrid-universal/releases/tag/v1.4.0) and also a new version of Aurelia-Slickgrid, thanks again for the contribution 😉

PR #742, opened 2022-08-11
fix(collectionAsync): hidden column does not load edit field selection

mcallegario:
Bug: Hidden column does not load edit field with selection options

The "this.slickGrid.getColumns()" method only returns the visible fields of the grid, so if the field that was configured to return a list of elements (collectionAsync) is hidden in the grid, it does not receive the list of elements.

note: Unfortunately after version 1.3.x I am no longer able to perform the "pnpm run build".
image

### DIFF
--- a/packages/vanilla-bundle/src/components/__tests__/slick-vanilla-grid.spec.ts
+++ b/packages/vanilla-bundle/src/components/__tests__/slick-vanilla-grid.spec.ts
@@ -659,12 +659,12 @@ describe('Slick-Vanilla-Grid-Bundle Component instantiated via Constructor', () 
         const mockCollection = ['male', 'female'];
         const promise = new Promise(resolve => resolve(mockCollection));
         const mockColDefs = [{ id: 'gender', field: 'gender', editor: { model: Editors.text, collectionAsync: promise } }] as Column[];
-        const getColSpy = jest.spyOn(mockGrid, 'getColumns').mockReturnValue(mockColDefs);
+        //const getColSpy = jest.spyOn(mockGrid, 'getColumns').mockReturnValue(mockColDefs);
 
         component.columnDefinitions = mockColDefs;
 
         setTimeout(() => {
-          expect(getColSpy).toHaveBeenCalled();
+          //expect(getColSpy).toHaveBeenCalled();
           expect(component.columnDefinitions[0].editor).toBeTruthy();
           expect(component.columnDefinitions[0].editor.collection).toEqual(mockCollection);
           expect(component.columnDefinitions[0].internalColumnEditor.collection).toEqual(mockCollection);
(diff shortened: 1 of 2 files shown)


ghiscoding:
yeah don't use the `build` (which is meant for something else), use the `bundle` instead which will delete all previous `dist` folders

packages/vanilla-bundle/src/components/slick-vanilla-grid-bundle.ts
```diff
@@ -1434,7 +1434,7 @@ export class SlickVanillaGridBundle {
 
     if (this.slickGrid) {
       // find the new column reference pointer & re-assign the new editor to the internalColumnEditor
-      const columns = this.slickGrid.getColumns();
+      const columns = this.columnDefinitions;
```
ghiscoding, line 1437: please remove this line and apply it directly below
```diff
- const columns = this.columnDefinitions;
if (Array.isArray(columns)) {
-        const columnRef = columns.find((col: Column) => col.id === column.id);
+        const columnRef = this.columnDefinitions.find((col: Column) => col.id === column.id);
```

mcallegario: to remove the "const columns" I also have to change the "if" list. Is it good like this?

image

ghiscoding: that seems correct, are you going to be able to fix the unit tests or do you need help?

packages/vanilla-bundle/src/components/__tests__/slick-vanilla-grid.spec.ts
```diff
@@ -659,12 +659,12 @@ describe('Slick-Vanilla-Grid-Bundle Component instantiated via Constructor', ()
...
         const promise = new Promise(resolve => resolve(mockCollection));
         const mockColDefs = [{ id: 'gender', field: 'gender', editor: { model: Editors.text, collectionAsync: promise } }] as Column[];
-        const getColSpy = jest.spyOn(mockGrid, 'getColumns').mockReturnValue(mockColDefs);
+        //const getColSpy = jest.spyOn(mockGrid, 'getColumns').mockReturnValue(mockColDefs);
 
         component.columnDefinitions = mockColDefs;
 
         setTimeout(() => {
-          expect(getColSpy).toHaveBeenCalled();
+          //expect(getColSpy).toHaveBeenCalled();
```
ghiscoding, line 667: instead of commenting the unit test lines, can you remove all of the ones you commented out please

ghiscoding (changes requested):

packages/vanilla-bundle/src/components/__tests__/slick-vanilla-grid.spec.ts
mcallegario: Done, removed unnecessary test from getcolumn method

ghiscoding (approved):

packages/vanilla-bundle/src/components/__tests__/slick-vanilla-grid.spec.ts
ghiscoding: thanks :)

ghiscoding:
Thanks for the PR, this was released under new version [v1.4.0](https://github.com/ghiscoding/slickgrid-universal/releases/tag/v1.4.0), thanks again for the contribution 😉

PR #921, opened 2023-03-01
Update slickDraggableGrouping.ts

dmitov92:
The reorderedColumns is now not getting lost on new columns instance


dmitov92:
This won't work, because you are reassigning the variable at the bottom. Abandon this and remove the reassigning

ghiscoding:
Hello, I'm not sure what you were trying to fix, can you describe the problem a bit more? Perhaps an animated gif would be useful

PR #922, opened 2023-03-01
fix: draggable grouping shouldn't throw error when dynamically changing columns

dmitov92:
- get latest reorderedColumns on each end event handler


dmitov92:
@ghiscoding The columns are getting updated from outside of this function and it often results in "cannot use 'id' of undefined" error as the reorderedColumns instance was outdated. If we hide/show columns or we reassign the columns input with a new instance, we are getting this error. Preserving the columns instance doesn't solve all errors.

dmitov92:
> 

We are changing the columns via the angular-slickgrid component. You are saving the columns in a let variable, but then we are still able to call setColumns from other functions and when this happens, the let variable doesn't know about the new columns. We should always get all current columns in the beginning of the END event handler just like in this commit or else the let variable will become outdated. In our case, we are changing the columns after the initialization and this variable doesn't know about the change, but the angular-slickgrid component knows about it.

ghiscoding:
ok great thanks for the explanation, I was busy with other things but I should be able to test this in the coming days and publish a new release after that. I'm expecting that all done by the end of the week.

ghiscoding:
@dmitov92 this is now released under Slickgrid-Universal [v2.6.2](https://github.com/ghiscoding/slickgrid-universal/releases/tag/v2.6.2) and Angular-Slickgrid [v5.6.1](https://github.com/ghiscoding/Angular-Slickgrid/tree/v5.6.1), thanks for the contribution 😃

PR #971, opened 2023-05-12
fix(core): set `wheel` event listener to passive for better perf

ghiscoding:
- this removes 3 out of 20 warnings (17 remainings), all the other ones are coming from SlickGrid core lib and will be fixed there

image


ghiscoding:
actually found that not all events can use `passive` since some of the events do use `preventDefault` and that is not compatible with `passive` mode. So I reverted some of them in PR #770

PR #1117, opened 2023-09-23
fix(resizer): resize without container

zewa666:
as discussed here, this fixes the behavior if no autoresize.container option is defined

### DIFF
--- a/examples/vite-demo-vanilla-bundle/src/examples/example02.ts
+++ b/examples/vite-demo-vanilla-bundle/src/examples/example02.ts
@@ -182,7 +182,6 @@ export default class Example2 {
 
     this.gridOptions = {
       autoResize: {
-        container: '.demo-container',
         bottomPadding: 30,
         rightPadding: 10
       },
--- a/packages/common/src/services/__tests__/resizer.service.spec.ts
+++ b/packages/common/src/services/__tests__/resizer.service.spec.ts
@@ -298,6 +298,22 @@ describe('Resizer Service', () => {
       expect(serviceCalculateSpy).toReturnWith({ height: fixedHeight, width: fixedWidth });
     });
 
+    it('should calculate new dimensions even when no container element is defined', () => {
+      const newHeight = 440;
+      const fixedWidth = 800;
+      mockGridOptions.gridWidth = fixedWidth;
+      mockGridOptions.autoResize!.container = undefined;
+      service.init(gridStub, divContainer);
+      const serviceCalculateSpy = jest.spyOn(service, 'calculateGridNewDimensions');
+
+      Object.defineProperty(window, 'innerHeight', { writable: true, configurable: true, value: newHeight });
+      window.dispatchEvent(new Event('resize'));
+      service.calculateGridNewDimensions(mockGridOptions);
+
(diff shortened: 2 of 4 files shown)


zewa666:
holy moly this is an extensive suite of tests. what a joy seeing the CI crunching through them so slow just because of the sheer amount 🥰

ghiscoding:
yeah it took a really long time to get full 100% coverage 😆 

Can you provide more details about why create this PR? You want to make the auto-resize work without a container?

zewa666:
its not about the autoResize itself but without the setting currently the grid doesnt render at all as mentioned in the other discussion. So the eager exit with null instead of 0 caused that scenario. If you e.g. remove that prop on Example2 without the fix, half of your cypress tests fail

zewa666:
> yeah it took a really long time to get full 100% coverage 😆
> 
> Can you provide more details about why create this PR? You want to make the auto-resize work without a container?
> 
> EDIT
> 
> Oh I see, it's linked to your last 2 [comments](https://github.com/ghiscoding/Angular-Slickgrid/issues/1266#issuecomment-1732407804) in Angular-Slickgrid. I always provide a container, but what will happen when you don't provide a container, will it go to the closest parent container or something? The auto-resize technically need a container for it to resize to its content, so I don't fully get how this will work

no there is a code path where it will essentially exit with 0,0 instead null but nevertheless trigger the rendering

ghiscoding:
hmm yeah the flag is optional because it's part of `ResizerOption` instead of the `autoResize` option, perhaps it could be reviewed in the future and not extend `ResizerOption` that way, 

https://github.com/ghiscoding/slickgrid-universal/blob/c6a095702a672e14b442e71be492942c07d6f1e6/packages/common/src/interfaces/resizerOption.interface.ts#L11-L15

Maybe another way would be to keep the extend but add it to `AutoResizeOption` as non-optional in there, this way we could keep the extend

https://github.com/ghiscoding/slickgrid-universal/blob/c6a095702a672e14b442e71be492942c07d6f1e6/packages/common/src/interfaces/autoResizeOption.interface.ts#L1-L6

ghiscoding:
hmm looking again at your code change, I guess it's acceptable since in that case it will take the window container and resize against it, I could go ahead with a merge then. Any last comment?

zewa666:
I'm not sure I can follow you. The thing is that the modified function calculates based on the found container. if the selector doesnt find anything (bc container setting is null, or any selector pointing to nothing) the code continues with returning 0 (zero) for height and width. but if its undefined it exits with zero hence the calc (num + null) fails in the calling side

zewa666:
> hmm looking again at your code change, I guess it's acceptable since in that case it will take the window container and resize against it,

Exactly 👍 

 > I could go ahead with a merge then. Any last comment?

sure thing

ghiscoding:
Thanks for the contributions and welcome back in another framework :D

ghiscoding:
Thanks for the contributions and welcome back in another framework :D

ghiscoding:
I'll push a patch version sometime tonight, I actually pushed another fix yesterday that I needed to fix some Aurelia-Slickgrid while migrating to Au2, so an official version release would be good to have

PR #1153, opened 2023-10-26
feat(graphql): add optional cursor pagination to GraphQL backend service

Harsgalt86:
==== Purpose of this PR ===
A PR to implement cursor based pagination (by chunks).
Currently slickgrid cursor based pagination only supports `first` and so when building the query ends up retrieving A-B, A-C, A-D...A-N.
This PR adds implementation for `last`, `after` and `before` 
(This is in contrast to the documentation on [the wiki](https://github.com/ghiscoding/slickgrid-universal/wiki/GraphQL-Pagination) that acknowledges it not being fully supported, but has an example of using `after` which isn't respected)

We can then use these chunk results to create pages
   *    page1: {startCursor: A, endCursor: B }
   *    page2: {startCursor: B, endCursor: C }
   *    page3: {startCursor: C, endCursor: D } 

This is in contrast to relay-style pagination which is better suited for realtime and infinite scrolling
   *    page1: {startCursor: A, endCursor: B }
   *    page2: {startCursor: A, endCursor: C }
   *    page3: {startCursor: A, endCursor: D }

=== Behavioural Subtlties ===
Cursor pagination is conceptually similar to a linked list, while offset pagination is similar to a Vector.
As such "goToPage(x)" is not supported when using a backend service configured to use cursors. 
In this case, the page-number `input` is rendered as a `span` instead. (See `slick-pagination.component`)

=== Usage ===
Along with everything documented on
https://github.com/ghiscoding/slickgrid-universal/wiki/GraphQL
The `pagination-service` exposes a new function `updatePageInfo` which accepts an updated `PageInfo` object, as part of the graphql return result.
This currently needs to be manually by the application called as part of the `process(query)` callback. In the future it might be possible in (looking at Aurelia-Slickgrid library) that this call can be made internally in that library automatically.

The pagination-service automatically determines if it the associated backendService is using cursorBased pagination. 
  
=== Future work ===
It would be nice to possibly support relay-style pagination in slickgrid that given when scrolling and reaching the end of the table, it allows automatic fetching and appending of the next page of results rather than actual number paging. Essientially implementing a tabular basd news feed.


ghiscoding:
Thanks for the PR, I'll have a deeper look in the weekend and do a review at that time. However for a quick glance at it, I see that you added a lot of cursor based code in the Pagination Service, is it essential though? I mean other than GraphQL, I'm not sure that other backend services, like OData, will need this and that adds a lot of potential not essential code to these other services (again like OData). If you think this is really the only way to deal with this, I could understand but I'm just asking as a general questions because I'm trying to keep these backend pagination services as generic as possible. I see you added tests, which is great but it looks like 4 lines are missing coverage, that would be really great if you can cover them or if you really can't then worst case we could add some istanbul skip, I really want to keep my 100% coverage since it took me a long time to get to that point and I work to keep it that way 😉

Harsgalt86:
>  I see that you added a lot of cursor based code in the Pagination Service, is it essential though?
Yeah, I think it does need to be "cursor aware" primarily because cursor based pagination doesn't support the `goToPage(x)` paradigm. Because of this, the pagination component also needs to avoid rendering an input that allows the user to jump to page X, so at the very least that component needs to know whether the underlying service is configured to use cursor pagination (In this case exposed via getter on the pagination service directly to the backend service)

> I'm trying to keep these backend pagination services as generic as possible
Agreed, in fact, keeping it generic was my intent too. My rationale was that GraphQL is currently the only cursor based service I'm aware of, but it's forseeable another backend service in the future may implement the cursor based paradigm. 
Hence the `processOnPageChanged` signature provides optional cursorArguments if the underlying service supports it, ignoring it if doesn't.

My initial approach here was to create two interfaces `OffsetPaginatedBackendService` and `CursorBasedPaginatedBackendService` which OData would only implement the former, while graphQL would implement both, but I believe I was able to achieve the same more succinctly by providing `cursorArgs` to the `processOnPageChanged` function.

> I mean other than GraphQL, I'm not sure that other backend services, like OData, will need this and that adds a lot of potential not essential code
I think I've handled that, the OData backend service has no associated changes (it is only aware of the original processOnPaginationChanged structure)

>  I see you added tests, which is great but it looks like 4 lines are missing coverage, that would be really great if you can cover them
Done (Awesome having them, wish my project had such extensive tests :))

> Oh maybe another thing, I also have [Example 10](https://ghiscoding.github.io/slickgrid-universal/#/example10) in here for GraphQL
Yup I can take a look into doing that

> Also what happened when you start filtering? I mean typically when I filter, I go back to first page, I assume that would be the same use case with GraphQL cursor based, correct?
Yeah, good point, this is my expectation too. I might be misremembering but I recall when developing that I saw when the page size changes it goes back to the first page via use of `defaultPaginationOptions`. I created `defaultCursorPaginatiopnOptions` to do the same thing, but I need to double check the implementation - and provide a unit test for it too

packages/common/src/interfaces/pageInfo.interface.ts
```diff
@@ -0,0 +1,13 @@
+export interface PageInfo {
```
ghiscoding, line 1: I wonder if we should rename this to `CursorPageInfo` instead since this seems to be related to cursor only?

packages/common/src/services/pagination.service.ts
```diff
@@ -86,6 +89,27 @@ export class PaginationService {
...
+   *    page1: {startCursor: A, endCursor: B }
+   *    page2: {startCursor: A, endCursor: C }
+   *    page3: {startCursor: A, endCursor: D }
+   *
+   *  non-relay pagination - Getting page chunks
+   *    page1: {startCursor: A, endCursor: B }
+   *    page2: {startCursor: B, endCursor: C }
+   *    page3: {startCursor: C, endCursor: D }
+   */
+  get cursorBased(): boolean {
```
ghiscoding, line 109: can we rename this to `isCursorBased` to better represent a flag?

packages/common/src/services/pagination.service.ts
```diff
@@ -184,23 +208,44 @@ export class PaginationService {
...
+      } else {
+        return Promise.resolve(this.getFullPagination());
+      }
     }
     return Promise.resolve(false);
   }
 
   goToPageNumber(pageNumber: number, event?: any, triggerChangeEvent = true): Promise<boolean | ServicePagination> {
+    if (this.cursorBased) {
+      console.assert(true, 'Cursor based navigation cannot navigate to arbitrary page');
```
ghiscoding, line 245: oh that's interesting, I've never used this type of console, so it looks like a `console.error` when the assertion is failing. That could be interesting to use in the future. I guess it's also easier for unit tests since you don't need an `if` condition wrapper and so less lines of code as untested haha

packages/common/src/services/pagination.service.ts
```diff
@@ -437,6 +492,10 @@ export class PaginationService {
     }
   }
 
+  updatePageInfo(pageInfo: PageInfo) {
```
ghiscoding, line 495: it doesn't update much, it might be better as `setPageInfo` since it's a simple assignment

packages/graphql/src/interfaces/graphqlPaginatedResult.interface.ts
```diff
@@ -18,19 +18,7 @@ export interface GraphqlPaginatedResult {
...
-        /** Do we have a previous page from current cursor position? */
-        hasPreviousPage: boolean;
-
-        /** What is the last cursor? */
-        endCursor: string;
-
-        /** What is the first cursor? */
-        startCursor: string;
-      };
+      pageInfo?: PageInfo;
```
ghiscoding, line 21: going back to my earlier comment, this does look better and better representing with `CursorPageInfo`

packages/graphql/src/services/__tests__/graphql.service.spec.ts
```diff
@@ -625,6 +626,28 @@ describe('GraphqlService', () => {
       expect(querySpy).toHaveBeenCalled();
       expect(currentPagination).toEqual({ pageNumber: 3, pageSize: 20 });
     });
+
+    describe("CursorBased related scenarios", () => {
+      describe.each`
+        description           | cursorArgs                    | expectation
```
ghiscoding, line 632: looks like a nice way of describing the test 👍🏻 though I didn't run the test yet, I could imagine how it would look like (like a table I imagine) 😉

packages/graphql/src/services/graphql.service.ts
```diff
@@ -141,13 +145,14 @@ export class GraphqlService implements BackendService {
...
-      };
+      datasetFilters = {};
 
-      if (!this.options.isWithCursor) {
+      if (this.options.isWithCursor && this.options.paginationOptions) {
+        datasetFilters = { ...this.options.paginationOptions };
+      }
+      else {
         const paginationOptions = this.options?.paginationOptions;
+        datasetFilters.first = ((this.options.paginationOptions && this.options.paginationOptions.first) ? this.options.paginationOptions.first : ((this.pagination && this.pagination.pageSize) ? this.pagination.pageSize : null)) || this.defaultPaginationOptions.first;
```
ghiscoding, line 155: I think this could be simplified with optional chaining `?.`
```ts
this.options?.paginationOptions.first ? condition1 : condition2;
```

packages/graphql/src/services/graphql.service.ts
```diff
@@ -492,25 +502,35 @@ export class GraphqlService implements BackendService {
...
-        first: pageSize
-      };
+    let paginationOptions: GraphqlPaginationOption | GraphqlCursorPaginationOption = {};
+    if (this.options?.isWithCursor) {
+      // use cursor based pagination
+      // when using cursor pagination, expect to be given a PaginationCursorChangedArgs as arguments,
+      // but still handle the case where it's not (can happen when initial configuration not pre-configured (automatically corrects itself next updatePageInfo() call))
+      if (cursorArgs && cursorArgs instanceof Object) {
+        // remove pageSize and newPage from cursorArgs, otherwise they get put on the query input string
+        // eslint-disable-next-line
```
ghiscoding, line 523: so what's the reason you need to disable eslint from this line? If it's a simple eslint rule missing for deconstructing into variable, we could modify the `.eslintrc` config and add that rule, depending on what linting you're trying to bypass

ghiscoding (changes requested):

ghiscoding:
@Harsgalt86 I reviewed and left a few comments but they are very minor, really good job. I would be really interested to see this in action, so yeah modifying or adding a new example would be really nice to see it live. Were you actually able to test this with an actual live GraphQL server? Also in continuation with the last comment I wrote with filtering and going back to page 1, have you tested with sorting as well? How would that work, would it resort from current cursor position and we drop end cursor because it probably changed?

packages/common/src/interfaces/paginationCursorChangedArgs.interface.ts
```diff
@@ -0,0 +1,15 @@
+import { PaginationChangedArgs } from './paginationChangedArgs.interface';
```
ghiscoding, line 1: can you add `type` to this import, there are some benefits in adding them, you can see this SO for more info https://stackoverflow.com/a/64243357/1212166

packages/common/src/services/pagination.service.ts
```diff
@@ -13,6 +14,7 @@ import type {
 import type { BackendUtilityService } from './backendUtility.service';
 import type { SharedService } from './shared.service';
 import type { Observable, RxJsFacade } from './rxjsFacade';
+import { PageInfo } from '../interfaces/pageInfo.interface';
```
ghiscoding, line 17: same as earlier, please add `type` to the import

packages/graphql/src/services/graphql.service.ts
Harsgalt86: Made the reading even easier by using null coalescing `??`

packages/graphql/src/services/graphql.service.ts
Harsgalt86: Screenshot 2023-10-30 at 11 22 15 am

I've made the disable rule more specific.

Destructing to exempt `pageSize` and `newPage` was the cleanest way to write this, otherwise I'd be iterating through the object props.

packages/common/src/services/pagination.service.ts
Harsgalt86: I've deleted that line and made the import  from the index file

Harsgalt86:
@ghiscoding Thanks for the quick turnaround on the review. I've implement your suggestions except for the demo stuff which I'll look into doing this week.

> Were you actually able to test this with an actual live GraphQL server? Also in continuation with the last comment I wrote with filtering and going back to page 1, have you tested with sorting as well?

Yes, out server already supported cursor based pagination, but is in the process of supporting the sortBy/filterBy syntax so will get will work through that and get back to you.


> How would that work, would it resort from current cursor position and we drop end cursor because it probably changed?

On a filter/sort parameter change, we would need to go back to the first page `{first: 20}` because, current pageInfo will be stale.

eg: Say you on page 2 and make a change to the filters which makes the first page and its 20 items filtered out. The current pageInfo (eg: hasPreviousPage) is out of date and the pagination needs to be updated the first available 20 items becomes page 1.

>  It would be great to include your PR as a new feature for this next release.

Sounds good! I'll do what I can.

ghiscoding (approved):

ghiscoding:
@Harsgalt86 I took a look at the update Example 10 and tested locally and I noticed something that I'm unsure if it's ok or not. 

When using cursor, if we change page then only `after` gets changed which is great

image

but if we change filter (for example on gender), then `after` is reset to empty string but we now also have 2 extra props `before` and `last` which are both set to `undefined`. You can already guess my question... is that intended?

image

apart from that, the rest looks ok I think

Harsgalt86:
> @Harsgalt86 I took a look at the update Example 10 and tested locally and I noticed something that I'm unsure if it's ok or not.
> 
> When using cursor, if we change page then only `after` gets changed which is great
> 
> image
> 
> but if we change filter (for example on gender), then `after` is reset to empty string but we now also have 2 extra props `before` and `last` which are both set to `undefined`. You can already guess my question... is that intended?
> 
> image
> 
> apart from that, the rest looks ok I think

Yeah, not intended, just pushed a fix for it. 

I was playing around with the sorting too, it looks like it works without having to go back to the first page. I guess because the 1) totalCount doesn't change 2) the query is executed immediately and 3) pageInfo is updated immediately after. Seems to be consistent with how graphql offset works too.

ghiscoding:
great that was quick, so is that ready to be merged then? I might do a release tomorrow or in 2 days max

Harsgalt86:
> so is that ready to be merged then

Yeah, I think it can be. If I come across any bugs I'll just raise fixes for them as they come up.

ghiscoding:
awesome, let's go ahead and merge :)

ghiscoding:
@Harsgalt86 I just released a new version for Aurelia-Slickgrid [v6.4.0](https://github.com/ghiscoding/aurelia-slickgrid/releases/tag/v6.4.0) which includes this, I did however add `.setCursorBased()` method (as shown in PR #1171) in the Pagination Service because I wanted to change the pagination cursor without reloading the Aurelia-Slickgrid page like you did in here. 

I think the only thing missing would be to add more documentation, would it be possible for you to update the Wiki? I think the Aurelia-Slickgrid should be open to public for editing, if it's not then let me know. I'll update the other libs Wikis afterward. Perhaps this Wiki
https://github.com/ghiscoding/aurelia-slickgrid/wiki/GraphQL-Pagination

Thanks again for your contribution 🚀

Harsgalt86:
@ghiscoding No worries. Is there any consideration to keeping the paginationService and graphQL service in sync programmatically? Or is it just up to the user to keep both in sync? 
It probably doesn't matter too much since in the real world there's no need for it to be changed on the fly and it's already being initialized with the backendService value

 I think with your change `example10` could also be updated as switching between cursor/offset involves reinitialization of the slickgrid instance. I can update that with the wiki update.

ghiscoding:
> No worries. Is there any consideration to keeping the paginationService and graphQL service in sync programmatically? Or is it just up to the user to keep both in sync?

I'm not exactly what you mean... but maybe it's related to this fact, I think that because at some point it creates a deep copy of the grid options, it kind of makes a copy of the backend services like GraphQL at least I think that is what is happening and because of that I had to add the `setCursorBased` in the Pagination Service because it wasn't in sync with GraphQL Service, is that what you meant? Perhaps the problem is that it might not be a good idea to add all these services inside the grid options, I think I should have kept it for simple props and not add service instances like we do with the backend services, it might something to consider in a future release though I'm not sure where else to put them. 

For the Example 10 in Slickgrid-Universal, I'm trying to improve it so that it doesn't require disposing the entire component and then query some DOM elements, I'll be adding some PubSub in PaginationService so that when I call `setCursorBased`, it will recreate the entire pagination component because at this point if you test it in Aurelia-Slickgrid, you will see that it doesn't remove the page input number like what you did in universal. Anyway that is not something that rush, it would be very rare that users will toggle between cursor/offset in the same grid, it's mostly just good for the demo. I don't think it's required for the Wiki since like I said, it would be rare that users would switch on-the-fly between these two.

Harsgalt86:
>  it kind of makes a copy of the backend services like GraphQL at least I think that is what is happening and because of that I had to add the setCursorBased in the Pagination Service because it wasn't in sync with GraphQL Service

Yeah, by having it as below, changing the backendServiceApi options would automtically update the paginationService...
```
get cursorPageInfo() {
   return !!this._backendServiceApi?.options.isWithCursor;
}
```
 but it seems as though maybe this wasn't working if the options were being deep copied.
In which I was thinking of a way of having to both `paginationService.setCursorBased` and `backendService.isWithCursor` .

But yeah mostly a non-issue.

PR #1174, opened 2023-11-03
feat: Graphql verbatim search terms

Harsgalt86:
==== Purpose of this PR ===
A PR to use filter search terms in a query "verbatim" without modification

Currently the backendService `updateFilters()` performs some manipulation of the columnFilter.searchTerms property. The flag on the column `columnFilter.verbatimSearchTerms` bypasses this manipulation and stringifies the search terms "as is".

Some examples of problematic manipulation are searching for columns that can be nullable
```
{ columnId: 'x', operator: 'in', searchTerms: 'null}
// Current: 
query{users(first:10, offset:0) { totalCount,nodes{ id,company,gender,name } }}
// withVerbatim:
query{users(first:10, offset:0, filterBy:[{field:gender, operator:EQ, value:"null"}]) { totalCount,nodes{ id,company,gender,name } }} 
```

Please see unit tests for more output differences.

=== Usage ===
```
{
    columnId: "gender",
    operator: OperatorType.in,
    searchTerms: this.genders, // supports ['male'], ['female'], ['male'/'female'], []    
    verbatimSearchTerms: true,
} as any,
```

I did consider putting the flag on the backend service instead of the column, but it's possible that different backend end points may want to treat properties differently for backwards compatibility. By putting the flag on the column itself it allows for more flexibility.

=== TODO ===
Currently this flag is only respected on the graphqlService. This needs to be applied to other backend services.

### DIFF
--- a/packages/common/src/interfaces/currentFilter.interface.ts
+++ b/packages/common/src/interfaces/currentFilter.interface.ts
@@ -16,4 +16,7 @@ export interface CurrentFilter {
 
   /** Target element selector from which the filter was triggered from. */
   targetSelector?: string;
+
+  /** If true, stringifies the search terms as is without modification */
+  verbatim?: boolean;
 }
--- a/packages/common/src/interfaces/currentFilter.interface.ts
+++ b/packages/common/src/interfaces/currentFilter.interface.ts
@@ -18,5 +18,5 @@ export interface CurrentFilter {
   targetSelector?: string;
 
   /** If true, stringifies the search terms as is without modification */
-  verbatim?: boolean;
+  verbatimSearchTerms?: boolean;
 }
--- a/packages/graphql/src/services/__tests__/graphql.service.spec.ts
+++ b/packages/graphql/src/services/__tests__/graphql.service.spec.ts
@@ -1239,21 +1239,21 @@ describe('GraphqlService', () => {
 
     describe("Verbatim ColumnFilters", () => {
       describe.each`
-        description                                         | verbatim | operator    | searchTerms            | expectation
-        ${"Verbatim false, Filter for null"}                | ${false} | ${'EQ'}     | ${null}                | ${'query{users(first:10, offset:0) { totalCount,nodes{ id,company,gender,name } }}'}
-        ${"Verbatim true,  Filter for null"}                | ${true}  | ${'EQ'}     | ${null}                | ${'query{users(first:10, offset:0, filterBy:[{field:gender, operator:EQ, value:"null"}]) { totalCount,nodes{ id,company,gender,name } }}'}
-        ${"Verbatim false, Empty String"}                   | ${false} | ${'EQ'}     | ${''}                  | ${'query{users(first:10, offset:0) { totalCount,nodes{ id,company,gender,name } }}'}
-        ${"Verbatim true,  Empty String"}                   | ${true}  | ${'EQ'}     | ${''}                  | ${'query{users(first:10, offset:0, filterBy:[{field:gender, operator:EQ, value:"\\"\\""}]) { totalCount,nodes{ id,company,gender,name } }}'}
-        ${"Verbatim false, Empty List"}                     | ${false} | ${'IN'}     | ${[]}                  | ${'query{users(first:10, offset:0) { totalCount,nodes{ id,company,gender,name } }}'}
-        ${"Verbatim true,  Empty List"}                     | ${true}  | ${'IN'}     | ${[]}                  | ${'query{users(first:10, offset:0, filterBy:[{field:gender, operator:IN, value:"[]"}]) { totalCount,nodes{ id,company,gender,name } }}'}
-        ${"Verbatim false, Filter for null (List version)"} | ${false} | ${'IN'}     | ${[null]}              | ${'query{users(first:10, offset:0, filterBy:[{field:gender, operator:IN, value:""}]) { totalCount,nodes{ id,company,gender,name } }}'}
-        ${"Verbatim true,  Filter for null (List version)"} | ${true}  | ${'IN'}     | ${[null]}              | ${'query{users(first:10, offset:0, filterBy:[{field:gender, operator:IN, value:"[null]"}]) { totalCount,nodes{ id,company,gender,name } }}'}
-        ${"Verbatim false, Empty String (List Version)"}    | ${false} | ${'IN'}     | ${['']}                | ${'query{users(first:10, offset:0, filterBy:[{field:gender, operator:IN, value:""}]) { totalCount,nodes{ id,company,gender,name } }}'}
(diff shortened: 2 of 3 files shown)


ghiscoding:
> Currently the backendService `updateFilters()` performs some manipulation of the columnFilter.searchTerms property. The flag on the column `columnFilter.verbatimSearchTerms` bypasses this manipulation and stringifies the search terms "as is".

I'm not entirely sure what you mean but manipulation... is it because some null are changed to empty string or something similar? is that what this is about to leave some data untouched, like `null` and maybe `number` that are transformed to string, is that it?

Harsgalt86:
> I'm not entirely sure what you mean but manipulation... is it because some null are changed to empty string or something similar? is that what this is about to leave some data untouched, like `null`, `undefined` and maybe `number` that are sometime transformed to string, is that this PR is about?

Yeah, exactly. Specifically,`null, "", []` are all getting removed from the query. (First 6 unit tests in the table).

Further, `[null]` and `[""]` becomes `""`. This is problematic for our backend server in regards to 1) a set being transformed to a string 2) two distinct values in a set being transformed into the same value.

So yes, this PR allow some columns to leave the searchTerms untouched.

packages/common/src/interfaces/currentFilter.interface.ts
```diff
@@ -16,4 +16,7 @@ export interface CurrentFilter {
 
   /** Target element selector from which the filter was triggered from. */
   targetSelector?: string;
+
+  /** If true, stringifies the search terms as is without modification */
```
ghiscoding, line 20: I know it's just a comment and the grammar is ok but I would prefer to see `stringify` since that is what everyone uses without thinking about the grammar ;).

ghiscoding:
Note that there are some reasons why some of these are being transformed, it might not work with all type of Filters (some filters might require string only, for example an input value in theory only works with string). As long as you are aware of the reasons and that you're not introducing breaking changes, it should be fine and I have plenty of unit tests and E2E tests to cover them, so that should be fine, that's really where all these unit tests become handy :)

Harsgalt86:
Thanks again for reviewing this so quickly.

> some filters might require string only, for example an input value in theory only works with string). As long as you are aware of the reasons

Yeah, that's acceptable. In those cases, if using `verbatimSearchTerms` it should now be the responsibility of the developer to sanitise the `searchTerms` array/value.

> that you're not introducing breaking changes

For sure. Should be safe since the behaviour is opt-in.

> that's really where all these unit tests become handy 

Absolutely.

packages/graphql/src/interfaces/graphqlServiceOption.interface.ts
```diff
@@ -40,4 +40,10 @@ export interface GraphqlServiceOption extends BackendServiceOption {
    * ex.: { field: "name", operator: EQ, value: "John" }
    */
   keepArgumentFieldDoubleQuotes?: boolean;
+
+  /**
+   * When false, searchTerms may be manipulated to be functional with certain filters eg: string only filters.
+   * When true, JSON.stringify is used on the searchTerms and used in the query "as-is". It is then the responsibility of the developer to sanitise the `searchTerms` property if necessary.
+   */
+  verbatimSearchTerms?: boolean;
```
ghiscoding, line 48: would it make more sense to rename this one to something that better represent a flag as opposed to the top one in the other interface which is a prop and not a flag. Maybe `useVerbatimSearchTerms`, we might need to rename `isWithCursor` to maybe `useCursor` (we could deprecate the old name with `@deprecate` until the next major). 

Let me know what you think, I'm just looking at these options and trying to align them and use better name if possible, I don't think anyone uses the `isWithCursor` yet since we just released it, that is from your work of last week, so now would be a good time to deprecate :)

oh and I guess you could also remove the text "(NOT FULLY IMPLEMENTED)" that I see in the comment up there.

packages/common/src/interfaces/currentFilter.interface.ts
```diff
@@ -21,5 +21,5 @@ export interface CurrentFilter {
    * When false, searchTerms may be manipulated to be functional with certain filters eg: string only filters.
    * When true, JSON.stringify is used on the searchTerms and used in the query "as-is". It is then the responsibility of the developer to sanitise the `searchTerms` property if necessary.
    */
-  verbatimSearchTerms?: boolean;
+  useVerbatimSearchTerms?: boolean;
```
ghiscoding, line 24: oh I think I didn't explain well, I wanted to keep this one as `verbatimSearchTerms`

packages/graphql/src/interfaces/graphqlServiceOption.interface.ts
```diff
@@ -45,5 +45,5 @@ export interface GraphqlServiceOption extends BackendServiceOption {
    * When false, searchTerms may be manipulated to be functional with certain filters eg: string only filters.
    * When true, JSON.stringify is used on the searchTerms and used in the query "as-is". It is then the responsibility of the developer to sanitise the `searchTerms` property if necessary.
    */
-  verbatimSearchTerms?: boolean;
+  useVerbatimSearchTerms?: boolean;
```
ghiscoding, line 48: this one is ok as `useVerbatimSearchTerms` because it's a flag, so no change required here

ghiscoding (changes requested):
left some comments for code change, thanks

ghiscoding (approved):

ghiscoding:
looks good thanks, do you have anything planned? I'm currently working on the `multiple-select-vanilla` library, I should be done by the weekend, so I expect to be able to release a new version for both projects at that time

Harsgalt86:
No, nothing immediate.

I'm considering at some point looking into extending the pagination to allow for an infinite scrolling table so you could just scroll back up to see previous rows, rather than having to navigate back and forth through pages. Could be triggered by reaching the scroll bottom or a stand alone "fetch more" button.

ghiscoding:
If you want a hint on how to do that, I decided long time ago that in my libs we'll only use the SlickGrid DataView, so with that in mind you can use the methods from the DataView to push data with `addItem` (or maybe even easier with Slickgrid-Universal GridService). That's the easy part but I'm not sure what you'll do when you reach the end, I guess the easy way would be to stop but I see many grids just make it infinite by redisplaying the first rows... not sure how you'll do that.

Anyway out of context, but similar approach, with Backend Services I clear the DataView data and insert data of the next page. Also another similar approach, in the past, I've done a lazy loading implementation where I was loading the first 500 rows so that the user can start using the grid right away and then keep pushing sets of 5000 in a lazy loading way... in summary, these 2 approaches are similar to how infinite scroll would work but also a little different because I don't know how to deal with the "when you reach the end"

ghiscoding:
@Harsgalt86 this is now available in Slickgrid-Universal [v3.5.0](https://github.com/ghiscoding/slickgrid-universal/releases/tag/v3.5.0) and Aurelia-Slickgrid [v6.5.0](https://github.com/ghiscoding/aurelia-slickgrid/releases/tag/v6.5.0)

PR #1188, opened 2023-11-11
fix: improve build & types exports for all targets, Node, CJS/ESM

ghiscoding:
- previous implementation didn't pass all type exports as can be shown in [Are the types wrong?](https://arethetypeswrong.github.io/?p=%40slickgrid-universal%2Fcommon%403.5.0) website, already applied the changes in `multiple-select-vanilla` library with a much greater success
- better approach came from RxJS [package.json](https://github.com/ReactiveX/rxjs/blob/master/packages/rxjs/package.json) great implementation of it

image


ghiscoding:
Much better after releasing a new version with this change :)

image

PR #1217, opened 2023-11-24
feat: Column.excludeFieldFromQuery, exclude field but keep fields array

Harsgalt86:
**-- Situation --**
From the GraphQL, we may have a fragment that is as such
```
fragment user on User{
   firstName
   lastName
   fullName @client
}
```
We want the `firstName` and `lastName` from the server, but `fullName` will be populated locally through policy
https://www.apollographql.com/docs/react/local-state/managing-state-with-field-policies/#querying

In the table, we want the column to display the fullname
```
{
    id: "full-name",
    field: "fullName",
    fields: ["firstName", "lastName"],
}
```

**-- Problem --**
The generated query by slickgrid attempts to request all 3 properties `firstName`, `lastName` and `fullName`.
We want to only request `firstName` and `lastName`

**-- Solution --**
There already exists a column propety `excludeFromQuery` that excludes the entire columnDef from the buildQuery. This PR introduces another property `excludeFieldFromQuery` which excludes `column.field` but still adds `column.fields` to the query 

The result is the following column definition
```
{
    id: "full-name",
    field: "fullName",
    fields: ["firstName", "lastName"],
    excludeFieldFromQuery: true,
}
```

Side note: The `queryField` property basically solves the same underlying problem but specifically  for filtering and sorting.


ghiscoding:
hmm I'm wondering if it wouldn't be better to do the inverse instead? For example keep using `excludeFromQuery` but have another inversed flag for the fields `includeFieldsInQuery`. I had this idea just because it might be confusing to have `excludeFromQuery` and `excludeFieldFromQuery`, they seem very similar... your new flag is not that bad though, I'm just thinking what would be less confusing if possible, though I'm not crazy about the name I gave `includeFieldsInQuery`, it's just a thought. It's bedtime for me now, so think about it and I'll reply tomorrow

Harsgalt86:
> they seem very similar...

From how I've been using TypeScript, interfaces and documentation, I think it's handy they similar as they appear close together when looking for similar behaviours. eg all the other `exclude*` options on the column are close together (`excludeFromExport`, `excludeFromGridMenu`, `excludeFromHeaderMenu` etc). Having said that `queryField` is functionally similar and not anywhere near them...

However, I think the strongest argument is that `excludeFromQuery` and `excludeFieldFromQuery` are logically consistent. For example:
```
{
    excludeFromQuery: true, 
    excludeFieldFromQuery: true,
}
```
=> The entire column, and by extension, the field, are excluded from the query. The behaviour is consistent with the settings

Conversely
```
{
    excludeFromQuery: true, 
    includeFieldsInQuery: true,
}
```
is not logically consistent as the entire column is excluded and the `fields` are NOT included. The behaviour is not consistent with specifying `includeFieldsInQuery: true`

Harsgalt86:
Another property name could be `queryOnlyUsesFields`?

ghiscoding:
> is not logically consistent as the entire column is excluded and the `fields` are NOT included. The behaviour is not consistent with specifying `includeFieldsInQuery: true`
> 
> Would you ever put the two options together? Probably not... but at least with the former the behaviour isn't ambiguous? I'll leave it to you to decide :)

Yes that was my intention, I thought of that approach because that is how a TypeScript `tsconfig.json` works, you can exclude but also include and I'm pretty sure include has priority over the exclude but I could be wrong too, anyway that is how I got the idea

> Another property name could be `queryOnlyUsesFields`?

hmm not crazy about the name either haha... but in that case then we might as well keep your first approach but I'm wondering if we should have 3 options then? `excludeFromQuery`, `excludeFieldFromQuery` and `excludeFieldsFromQuery`, however in that case you might think, why declare the `fields` in the first place? ... anyway, we might as well go with your current approach and be done with it :)

Let me know if you have any other suggestions or if I should go ahead with merging?

packages/common/src/interfaces/column.interface.ts
```diff
@@ -108,6 +108,9 @@ export interface Column<T = any> {
   /** Default to false, which leads to exclude the column title from the Grid Menu. */
   excludeFromGridMenu?: boolean;
 
+  /** Defaults to false, which leads to exclude the field property, but including the fields property, from the query (typically a backend service query) */
```
ghiscoding, line 111: can you use backticks on the column options, when we use that it shows them as code in the editor as well
```ts
/** ... which leads to exclude the `field` property, but including the `fields` property
```
Could you also update the next option I created as well. I started using this approach not long ago, I didn't know VSCode also shows code this way :)  see example below, thought it's hard to see, the field has a color background

image

packages/graphql/src/services/graphql.service.ts
```diff
@@ -98,7 +98,9 @@ export class GraphqlService implements BackendService {
     const columnIds: string[] = [];
     if (columnDefinitions && Array.isArray(columnDefinitions)) {
       for (const column of columnDefinitions) {
-        columnIds.push(column.field);
+        if (!column.excludeFieldFromQuery) {
+          columnIds.push(column.field);
```
ghiscoding, line 102: could you apply the same logic in the OData Service as well, they should have the same feature as much as possible

ghiscoding (changes requested):

ghiscoding:
@Harsgalt86 I want to release a new version this weekend (Saturday or Sunday), I would rather wait for this PR to be updated but I don't want to wait too long 😉

Harsgalt86:
Sorry, only just saw this. I did look at updating the odata service before
opening the PR but didn't see any code directly adding 'fields' in the
buildQuery function nor were there any unit tests directly testing the same
way the graphql.spec was. I'll take another look with fresh eyes, hopefully
tomorrow morning if I get some spare time, otherwise I will on Monday. It
would be good to get this merged in for the next release, but don't want to
hold you up either :)

ghiscoding:
> I did look at updating the odata service before opening the PR but didn't see any code directly adding 'fields' in the buildQuery function nor were there any unit tests directly testing the same way the graphql.spec was.

oh right I just remembered now that it's only for GraphQL, we didn't need it for OData because most of the time in OData and REST API you pull all the fields at once, that's kinda why we migrated to GraphQL. 

So it's all good, if you could just do the other little changes I asked, that would be great. I

Harsgalt86:
All done.

ghiscoding (approved):

PR #1224, opened 2023-11-27
feat(GraphQL): Provide ability to specify operationName

Harsgalt86:
-- Background --
By default when using Graphql (Apollo), all network requests hit the same endpoint, and as such they all have the same name in the devtools. e.g: `graphql`.  This makes performance monitoring and diagnosing problematic queries difficult. To get around this, we append the operationName of a graphql query to the url. e.g `graphql?GetUsers`, `graphql?getUserData` so we can uniquely identify each query.
https://stackoverflow.com/questions/64642948/is-there-a-way-to-name-graphql-requests-in-the-devtool-network-tab

-- Problem --
The dynamically built query in slickgrid does not allow specifying an operationName and so all queries generated by slickgrid have an operationName of `undefined`
eg: `query { datasetName (....) } `
https://stackoverflow.com/questions/57049366/get-query-mutation-operation-name

As such, all the slickgrid queries look like `graphql?undefined` when looking in devtools.

-- Solution -- 
Provide an optional GraphQlOption `operationName` that if provided will be used during `GraphQlService.buildQuery`
eg: `query operationName { datasetName (...) }`

### DIFF
--- a/packages/graphql/src/interfaces/graphqlServiceOption.interface.ts
+++ b/packages/graphql/src/interfaces/graphqlServiceOption.interface.ts
@@ -16,7 +16,7 @@ export interface GraphqlServiceOption extends BackendServiceOption {
   /** What is the dataset, this is required for the GraphQL query to be built */
   datasetName: string;
 
-  /** Use for defining the operation name when building the GraphQL query */
+  /** Used for defining the operation name when building the GraphQL query */
   operationName?: string;
 
   /**
--- a/packages/graphql/src/interfaces/graphqlServiceOption.interface.ts
+++ b/packages/graphql/src/interfaces/graphqlServiceOption.interface.ts
@@ -16,6 +16,9 @@ export interface GraphqlServiceOption extends BackendServiceOption {
   /** What is the dataset, this is required for the GraphQL query to be built */
   datasetName: string;
 
+  /** Use for defining the operation name when building the GraphQL query */
+  operationName?: string;
+
   /**
    * Extra query arguments that be passed in addition to the default query arguments
    * For example in GraphQL, if we want to pass "userId" and we want the query to look like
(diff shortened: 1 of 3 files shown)


ghiscoding:
that looks fine but I would like to know if you have many more changes like this? Because I don't want to publish new version every week, especially since it's only GraphQL related 😝

Harsgalt86:
>  I would like to know if you have many more changes like this


I don't have anything planned, but we are in testing phase + new devolopment phase so something might pop up.

> I don't want to publish new version every week


Yeah, no worries, this change isn't urgent.  If anything does pop up I'll avoid staggering the PRs and open them together (individually) so they can be released at once. 

Please don't feel the need to release on my account :) 
  
> I published the other PR over the weekend already


Thank you very much for being so responsive.

ghiscoding:
Yeah no worries, my main focus at the moment is the next major version which I hope to release before Christmas. See Slickgrid-Universal [Roadmap 4.0](https://github.com/ghiscoding/slickgrid-universal/discussions/1108). While working on this, I found that there are things that I could do in current version but for the most part it's only the next branch.

So I will probably wait until next weekend to release your new code. Another thing about code change in Slickgrid-Universal is that at the moment, anytime I publish a new version of Slickgrid-Universal, I have to release all downstream libs (Angular-Slickgrid, Aurelia-Slickgrid, ...) because of a TypeScript issue that I have with the lib, I think it's related to the fact that I `extends GridOption` for custom grid options in Aurelia-Slickgrid and others and that causes TypeScript to not be happy when doing that and users get build errors, so at the moment I have to release all libs on the same day, I have an idea on how to fix that in the next major release but in the meantime I have to publish many projects on the same day, hence why I was asking if you had more things to bring 😉 

At least now, since last weekend, I ship new releases via GitHub Action in CI workflow to provide npm provenance. So it is becoming quicker to release, so there's that 😄

Harsgalt86:
> I don't have anything planned


Spoke to soon, I'm looking into disabling the "select all" checkbox when there are no rows or all the rows are disabled. Hoping to get that in before your release this weekend.

PR #1234, opened 2023-11-29
bug/feat: Disable select all checkbox when there is nothing to select

Harsgalt86:
== Issue Reproduction ==
Setup: 
Page 1 has 10 items, 4 selectable, 6 non selectable
Page 2 has 10 items, none selectable

Repro:
1) Navigate to page 2
2) Check the select all checkbox. Nothing will be selected, but the select all will be ticked. This is OK since all selectable items are "selected" (even if that's 0)
3) Navigate to page 1. The select all checkbox will be ticked but nothing will selected, even though there are 4 items that could be selected.

Expected: Select all checkbox should be consistent with what's displayed on the table
Actual: Select all checkbox was selected, but nothing on the table was selected even though they could be

Fix:
First attempt to fix was to simply clear the select all checkbox by calling `gridState.setSelected([])`, but because the selection set wasn't actually changing, the `slickgCheckboxSelectColumn.handleSelectedRows()` wouldn't fire.

Second attempt is this PR which is when there is nothing to select, clear and disable the selectAll checkbox to prevent the user from putting it into an inconsistent state. Also had to respect to `applySelectOnAllPages` as we could not make assumptions on data that wasn't on the grid.

### DIFF
--- a/packages/common/src/extensions/slickCheckboxSelectColumn.ts
+++ b/packages/common/src/extensions/slickCheckboxSelectColumn.ts
@@ -81,10 +81,14 @@ export class SlickCheckboxSelectColumn<T = any> {
       .subscribe(grid.onClick, this.handleClick.bind(this) as EventListener)
       .subscribe(grid.onKeyDown, this.handleKeyDown.bind(this) as EventListener);
 
-    if (this._isUsingDataView && this._dataView && this._addonOptions.applySelectOnAllPages) {
-      this._eventHandler
-        .subscribe(this._dataView.onSelectedRowIdsChanged, this.handleDataViewSelectedIdsChanged.bind(this) as EventListener)
-        .subscribe(this._dataView.onPagingInfoChanged, this.handleDataViewSelectedIdsChanged.bind(this) as EventListener);
+    if (this._isUsingDataView && this._dataView) {
+      this._eventHandler.subscribe(this._dataView.onRowsChanged, this.handleRowsChanged.bind(this) as EventListener);
+
+      if (this._addonOptions.applySelectOnAllPages) {
+        this._eventHandler
+          .subscribe(this._dataView.onSelectedRowIdsChanged, this.handleDataViewSelectedIdsChanged.bind(this) as EventListener)
+          .subscribe(this._dataView.onPagingInfoChanged, this.handleDataViewSelectedIdsChanged.bind(this) as EventListener);
+      }
     }
 
     if (!this._addonOptions.hideInFilterHeaderRow) {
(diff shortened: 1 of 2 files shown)


Harsgalt86:
Page 1 - initial load. 2 Rows can be selected 


Page 2 - after clicking the check all checkbox when they are are no rows selectable
Screenshot 2023-11-30 at 9 49 55 am

Page 1 - The checkbox is checked, but the selectable items are not checked
Screenshot 2023-11-30 at 9 50 15 am

packages/common/src/extensions/slickCheckboxSelectColumn.ts
```diff
@@ -461,6 +465,18 @@ export class SlickCheckboxSelectColumn<T = any> {
...
   }
 
+  protected handleRowsChanged() {
+    let disableSelectAll = false;
+
+    if (!this._addonOptions.applySelectOnAllPages && typeof this._selectableOverride === 'function') {
+      const filteredItems = this._dataView.getFilteredItems();
+      // disable select all checkbox if 1) nothing to select or 2) no item can be selected
+      // requires applySelectOnAllPages to be false, as we don't want to make assumptions on data on other pages
+      disableSelectAll = filteredItems.length === 0 || filteredItems.every(dataItem => !this.checkSelectableOverride(0, dataItem, this._grid));
```
ghiscoding, line 475: I'm not too crazy about this approach, if I understand correctly it loop through the entire dataset every single time a row changes, which is in some way really bad performance wise. Imagine a grid with 100K rows this would keep the CPU a little too busy... There must be a way to do it in 1 pass instead of multiple passes every time a row changes. 

Another thing to note also would be that if we have 1 valid row, then there's no reason to investigate further

ghiscoding (commented):
see comment

packages/common/src/extensions/slickCheckboxSelectColumn.ts
ghiscoding: What about using a State object to keep the count of enabled/disabled like `state = { enabled: 333, disabled: 1 }` and if you want to restart from scratch then also add a reset function like `forceStateCheck()`. If that doesn't then another solution could be to use a `Set` or a `Map` to keep just the `id`/`state` pair, at least that is a lot more performant than looping through the entire dataset, updating the Map or Set is very quick, it also gets rid of duplicate since by default it doesn't allow duplicates.

ghiscoding:
I'll close this PR since it's out of date and I don't think the implementation is the correct way to go here. We might need to revisit this in the future with a better approach, thanks.

PR #1279, opened 2023-12-21
feat: datasetIdPropertyName respected in newRowCreator

zewa666:
Hey ho, when playing around with the Excel Copy Buffer I noticed that in my sample when copying and pasting multiple rows which overflow, new row creation failed with

`[SlickGrid DataView] Each data element must implement a unique 'id' property`

My example was using the OData Backend hence with a custom id column, set via `datasetIdPropertyName`.

It took me quite a while to figure out that this can be solved by overriding the following property in the options:

```
excelCopyBufferOptions: {
     newRowCreator: (count) => { ... }
}
```

so I thought adding this feature to respect the `datasetIdPropertyName` should make it work out of the box.

See the added unit test for better clarification :)

### DIFF
--- a/packages/common/src/extensions/__tests__/slickCellExcelCopyManager.spec.ts
+++ b/packages/common/src/extensions/__tests__/slickCellExcelCopyManager.spec.ts
@@ -305,6 +305,21 @@ describe('CellExcelCopyManager', () => {
       expect(addItemSpy).toHaveBeenCalledWith(expect.objectContaining({ id: 'newRow_1' }));
     });
 
+    it('should respect datasetIdPropertyName when calling "newRowCreator" callback', () => {
+      const expectedColumnName = 'mySuperSpecialId';
+      gridOptionsMock.datasetIdPropertyName = expectedColumnName;
+      plugin.init(gridStub);
+      const mockGetData = { addItem: jest.fn() };
+      const getDataSpy = jest.spyOn(gridStub, 'getData').mockReturnValue(mockGetData as any);
+      const addItemSpy = jest.spyOn(mockGetData, 'addItem');
+
+      plugin.addonOptions!.newRowCreator!(2);
+
+      expect(getDataSpy).toHaveBeenCalled();
+      expect(addItemSpy).toHaveBeenCalledWith(expect.objectContaining({ [expectedColumnName]: 'newRow_0' }));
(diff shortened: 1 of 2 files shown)


zewa666:
I'll be on the lookout for additional issues/convenience features that might arise, so from my perspective dont bother with releasing this asap

ghiscoding:
ah good catch, I almost never use this feature because it conflicts with the Checkbox Row Selection since we cannot have 2 selections model loaded at the same time (excel copy requires cell selection model), however some users found ways to switch between models when mouse clicking and releasing. 

I agree that I'll wait a bit more before releasing anything, I also just got Aurelia-Slickgrid working Aurelia 2 as well officially yesterday, so that's a nice Xmas gift in advance 🎁 I also had trouble finding the Angular-Slickgrid issue reported by some users after publishing to AWS and others, anyway it looks to be all fine now, so I think we're good for some time :)

zewa666:
the excel like pasting is quite useful for those spreadsheety applications. often customers literally just ask for that in order to move from their precious excel spreadsheets 😅

ghiscoding:
I know the feeling... most users are just expecting a datagrid to behave the same as Excel or if it doesn't then they tag it as being a bad lib lol

Anyway, I'm happy with the journey to modernize the project, I'm pretty much done with what I wanted to achieve and even features that I only dreamed of are now part of the lib (removing jQuery was the biggest dream). The only thing that a user recently asked for was Dark Mode, but when I tell him to please contribute, I heard...nothing back as you can expect lol

zewa666:
you should be proud of your work if dark-mode is "all thats missing" 😅

to me the biggest selling point, besides the tons of features, is the test coverage. it gives that comfy feeling of knowing things wont fall apart when touching unknown code.

ghiscoding:
indeed the test coverage is really what made it possible to release new major versions with confidence. I know since a while ago that a lib needs to have proper test coverage to be well recognized. I also made the 6pac/slickgrid more popular since I added few hundreds Cypress tests, there's a lot more users since I dropped jQuery/jQueryUI and converted it to TypeScript, that migration was really the missing piece that I needed for me to migrate the core files into here and there's a lot more users now because of this change. Some users waited a long time to have full ES6/ESM imports, it's now all possible... woohoo

Happy early Xmas and always enjoyed all contributions from your side, thanks for being a long time user and echoing this to your coworkers, nice to have all these ⭐ too :D

PR #1298, opened 2023-12-30
feat: add onBeforePasteCell event

zewa666:
This adds the ability to define on a per cell basis whether a paste should be denied or allowed.

It can be configured with the new excelCopyBufferOption:

```typescript
excelCopyBufferOptions: {
  onBeforePasteCell: (e, args) => {
    // only disable for first cell of a row
    return args.cell > 0;
  }
},
```

This is going to be a pre-requisite for the Row Based Edit Plugin

### DIFF
--- a/packages/common/src/extensions/slickCellExternalCopyManager.ts
+++ b/packages/common/src/extensions/slickCellExternalCopyManager.ts
@@ -1,7 +1,7 @@
 import { createDomElement, stripTags } from '@slickgrid-universal/utils';
 
-import type { Column, ExcelCopyBufferOption, ExternalCopyClipCommand } from '../interfaces/index';
-import { SlickEvent, SlickEventData, SlickEventHandler, type SlickGrid, SlickRange } from '../core/index';
+import type { Column, ExcelCopyBufferOption, ExternalCopyClipCommand, OnEventArgs } from '../interfaces/index';
+import { SlickEvent, SlickEventData, SlickEventHandler, type SlickGrid, SlickRange, SlickDataView } from '../core/index';
 
 // using external SlickGrid JS libraries
 const CLEAR_COPY_SELECTION_DELAY = 2000;
@@ -20,6 +20,7 @@ export class SlickCellExternalCopyManager {
   onCopyCells = new SlickEvent<{ ranges: SlickRange[]; }>();
   onCopyCancelled = new SlickEvent<{ ranges: SlickRange[]; }>();
   onPasteCells = new SlickEvent<{ ranges: SlickRange[]; }>();
+  onBeforePasteCell = new SlickEvent<{ cell: number; row: number; item: any; columnDef: Column; value: any; }>();
 
   protected _addonOptions!: ExcelCopyBufferOption;
   protected _bodyElement = document.body;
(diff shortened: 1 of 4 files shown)


zewa666:
I wasn't sure where to document this since there doesn't seem to be any kind of docs around for the CellExcelCopyManager. If you could point me to the proper place I'll update the docs as well.

Furthermore, I'm again not sure in which example to add this so that I can write a couple of Cypress tests.

packages/common/src/extensions/slickCellExternalCopyManager.ts
```diff
@@ -345,6 +371,12 @@ export class SlickCellExternalCopyManager {
     }
   }
 
+  protected trigger<ArgType = any>(evt: SlickEvent, args?: ArgType, e?: Event | SlickEventData) {
```
ghiscoding, line 374: so this looks like a full copy of the `trigger` method from SlickGrid, so perhaps we could just make the SlickGrid method as a public method and call it from here (and probably better to rename it as something like `triggerEvent`, or any other name, to be a little more specific)

packages/common/src/extensions/slickCellExternalCopyManager.ts
```diff
@@ -68,6 +69,26 @@ export class SlickCellExternalCopyManager {
         this._grid.focus();
       }
     });
+
+    if (typeof this._addonOptions?.onBeforePasteCell === 'function') {
+      const dataView = grid?.getData<SlickDataView>();
```
ghiscoding, line 74: we could probably check the `grid` object in the `if` condition above since I don't think it will run without it and it would be nearly impossible to be undefined anyway

ghiscoding (changes requested):

ghiscoding:
> I wasn't sure where to document this since there doesn't seem to be any kind of docs around for the CellExcelCopyManager. If you could point me to the proper place I'll update the docs as well.
> 
> Furthermore, I'm again not sure in which example to add this so that I can write a couple of Cypress tests.

I do have a `excel-copy-buffer.md` in the Angular-Slickgrid project, I might have forgot to copy it here, you could copy it in the same grid functionalities folder and remove/update what's necessary

packages/common/src/extensions/slickCellExternalCopyManager.ts
zewa666: I was hesitant to ask for that but yep this would be my prefered solution as well

zewa666:
alright your mentioned feedback should be addressed wit hthe last commit.

The last part would be to define which example to take. I thought I'd go with example19 and deny pasting on a couple of cells and a complete row to show off how it works.

zewa666:
image

ghiscoding (approved):

ghiscoding:
looking good, did you wanted me to merge it now or did you wanted to add the Example 19 changes in this PR?

zewa666:
just wanted to make sure no 19 is the right one. yep I'll address the unit tests along with the cypress tests. but thats gonna happen next year. wish you a great new year 🥳

zewa666:
hmmm cant figure out how to properly simulate copy & paste witb cypress. I've prepped the sample, although with a little linter issue. maybe you coupd take a look at that one?

ghiscoding:
There's a small build error preventing the Cypress to run, just changing `e` to `_e` should resolve the problem. 

image

As for the Copy+Paste, I found this article [Handling Copy and Paste in Cypress](https://egghead.io/blog/handling-copy-and-paste-in-cypress) on eggheade.io, he's using `.invoke` which is required for pasting into an input value `cy.get("[type='email']").invoke('val', textToPaste)`. I'm not exactly sure what you're trying to do, worst we can simply test it by hand and skip Cypress tests for now

zewa666:
yep I've tried that but its not exactly doing whats expected since it works around the original browser paste behavior. I tried copying 3 consecutive cells and pasting them so that 2 cells are written but the 3rd overlaps with a blocked region. no matter what I tried I couldnt get it to perform the original browser behavior. so yeah I agree on skipping these. I'll later push the linter fix.

ghiscoding:
note that I often have to use the `force: true` in Cypress to overcome some problems with it. Anyway skipping it is ok too, let me know when you want the PR to be merged, looking good so far

zewa666:
@ghiscoding sry I've been sick the last two days and slowly recovering. I've fixed the linter issue and yeah I think we can go ahead and merge this, so that I can continue on my other branch

zewa666:
hmm there is still an issue with coverage as it seems. need too check that. havent been able to notice which part is yet to be covered

zewa666:
ok so it complains because I touched slickGrid.ts but it hasn't enough coverage I guess right? not sure that's something related to this PR, so from my side all fine and rdy for merge

ghiscoding:
> ok so it complains because I touched slickGrid.ts but it hasn't enough coverage I guess right? not sure that's something related to this PR, so from my side all fine and rdy for merge

no problem there, I've completed full coverage of SlickDataView yesterday but SlickGrid coverage will take a while longer, so it's fine don't worry about it, Codecov is a bit weird when coverage is missing, it will display that coverage is missing for x lines if you did touch these lines even if coverage didn't change.

Alright I'll go ahead and merge then, hopefully you recover soon, it's not a good year start but probably expected with all Christmas reunions 😅

PR #1311, opened 2024-01-05
feat(core): expose all SlickEvent via internal PubSub Service

ghiscoding:
- it was already available for `SlickGrid` and `SlickDataView` `SlickEvent`s but it wasn't yet available for other plugin events
- add a `getPubSubService()` method in SlickGrid to be able to easily get PubSub instance from within any plugins or for the user themselves
- you can also cancel the event from bubbling **but this is only available** when it's a **`SlickEvent`** (because regular PubSub events do not include any native event), see example below
- Note: the internal PubSub can now deal with both regular PubSub events and SlickEvents but their structure are different
1. regular PubSub event have a structure of:: `((args: any) => ...`
    - e.g. `onBeforeGridCreate` 
3. SlickEvent have a different structure of:: `((e: { eventData: SlickEventData; args: any; }) => ...`
   - e.g. `onSelectedRangesChanged` 

#### Example
So for example, the user can subscribe the event in multiple ways

1. old approach (still works), in this example we subscribe to the `CellSelectionModel`
```ts
// bind any of the grid events, e.g. onSelectedRangesChanged to show selection range on screen
const cellSelectionModel = this.sgb.slickGrid!.getSelectionModel();
this._eventHandler.subscribe(cellSelectionModel!.onSelectedRangesChanged, (_e, args) => {
  const targetRange = document.querySelector('#selectionRange') as HTMLSpanElement;
  targetRange.textContent = '';
  for (const slickRange of args) {
    targetRange.textContent += JSON.stringify(slickRange);
  }
});
```

2. new approach (get PubSub instance via `this.sgb.instances?.eventPubSubService`
```ts
this.sgb.instances?.eventPubSubService?.subscribe('onSelectedRangesChanged', ((e: { eventData: SlickEventData; args: any; })) => console.log(e));
```

5. or second new approach (get PubSub instance from SlickGrid via `this.sgb.slickGrid?.getPubSubService()`)
```ts
this.sgb.slickGrid?.getPubSubService()?.subscribe('onSelectedRangesChanged', ((e: { eventData: SlickEventData; args: any; })) => console.log(e));
```

### Cancel Event Bubbling
You can also cancel the event bubbling for some of the SlickEvent that supports it, for example the `onBeforeEditCell` 

```ts
this.sgb.instances?.eventPubSubService?.subscribe('onSelectedRangesChanged', ((e: { eventData: SlickEventData; args: any; })) => {
  const { column, item, grid } = args;
  if (item.percentComplete < 10) {
    return false;
  }
  return true;
});
```

### DIFF
--- a/examples/vite-demo-vanilla-bundle/src/examples/example19.ts
+++ b/examples/vite-demo-vanilla-bundle/src/examples/example19.ts
@@ -28,16 +28,27 @@ export default class Example19 {
     this.sgb = new Slicker.GridBundle(document.querySelector(`.grid19`) as HTMLDivElement, this.columnDefinitions, { ...ExampleGridOptions, ...this.gridOptions }, this.dataset);
     document.body.classList.add('salesforce-theme');
 
-    // bind any of the grid events
+    // bind any of the grid events, e.g. onSelectedRangesChanged to show selection range on screen
     const cellSelectionModel = this.sgb.slickGrid!.getSelectionModel();
     this._eventHandler.subscribe(cellSelectionModel!.onSelectedRangesChanged, (_e, args) => {
       const targetRange = document.querySelector('#selectionRange') as HTMLSpanElement;
       targetRange.textContent = '';
-
       for (const slickRange of args) {
         targetRange.textContent += JSON.stringify(slickRange);
       }
     });
+
+    // or subscribe to any events via internal PubSub (from `this.sgb.instances?.eventPubSubService` or `this.sgb.slickGrid?.getPubSubService()`)
+    // Note: SlickEvent use the structure:: { eventData: SlickEventData; args: any; }
+    //       while other regular PubSub events use the structure:: args: any;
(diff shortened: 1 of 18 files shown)


ghiscoding:
@zewa666 this might interest you... I added this for SlickGrid & SlickDataView when I created the new major but I forgot to add these to all plugins. 

The reason I added this was mainly for SlickGrid/SlickDataView because prior to this, I was looping through **all** of their events and was subscribing to all of them, if an SlickEvent was triggering, I was then rethrowing the event to my internal PubSub so that it could be used in frameworks like Angular-Slickgrid... but as you can imagine that had a lot of overhead, so this removes all of that overhead and also exposes it to the user :) 

below is the overhead code I had prior to this new approach (lots of overhead)

https://github.com/ghiscoding/slickgrid-universal/blob/263bf1388c99daf31e1c7d62111ef0215c3effe9/packages/vanilla-bundle/src/components/slick-vanilla-grid-bundle.ts#L778-L797

zewa666:
this looks like a great api simplification. not directly affecting me for now but great to know

PR #1323, opened 2024-01-11
feat(plugin): new Row Based Editor

### LINKED ISSUE #1299: Row Based Edit Plugin

The motivation of this plugin is to create a way for Users to be able to:

* [x] toggle one or multiple rows into editmode
  * [x] when in editmode, a custom css class is applied to the row, along with css vars for styleability (use of `slickGrid?.setCellCssStyles` ?!)
* [x] this is done via a new column
  * [x] which can be positioned by the consumer (https://github.com/zewa666/slickgrid-universal/commit/d4372aae30e774bff4d6ad577779ae93d79fdba2#r135914186)
  * [x] contains an edit, delete, cancel and update button
  * [x] modifications are tracked per row and can be undone using the cancel button
  * [x] the buttons can be styled by the user
  * [x] optionally a prompt message can be defined for cancel and update
* [x] The plugin also makes sure that if CellExternalCopyManager is turned on, pastes are only enabled within editmode cells, respecting any consumer overrides
* [x] A new example is created, along with Cypress tests
* [ ] Documentation is added as a new page to Grid Functionalities with a reference in Editors.md

### Suggested solution

Implement the requirements from above within the common/extensions package using a set of well defined gridOptions.

See https://github.com/ghiscoding/slickgrid-universal/discussions/1290 as related discussion to this issue

### Alternative

_No response_

### Additional context

_No response_

### Validations

- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

zewa666:
this adds the new row based edit plugin

closes #1299


ghiscoding:
for visibility could you add the animated gif you demoed in the issue in here for anyone who opens the PR to see more details. Thanks

zewa666:
yep I'll create one when writing the docs and attach it as comment. but first those nasty final two lines need coverage 😅

packages/common/src/extensions/slickRowBasedEdit.ts
```diff
@@ -0,0 +1,596 @@
...
+    const target = event.target as HTMLElement;
+    const idProperty = this._grid.getOptions().datasetIdPropertyName ?? 'id';
+    const targetRow = this._editedRows.get(dataContext[idProperty]);
+    if (
+      (target.classList.contains(BTN_ACTION_DELETE) ||
+        target.parentElement?.classList.contains(BTN_ACTION_DELETE)) &&
+      this._gridService
+    ) {
+      if (this._addonOptions?.actionButtons?.deleteButtonPrompt) {
+        if (
```
ghiscoding, line 371: is it more readable to merge these 2 `if` into a single one? ... I guess it's ok as is since I see the next few conditions are much longer, so we might as well stay consistent

packages/common/src/extensions/slickRowBasedEdit.ts
```diff
@@ -0,0 +1,596 @@
...
+    actionFragment
+      .appendChild(
+        createDomElement('span', {
+          className:
+            `${
+              options.rowBasedEditOptions?.actionButtons?.editButtonClassName ||
+              'button-style padding-1px mr-2'
+            } action-btns ` + BTN_ACTION_EDIT,
+          title:
+            options.rowBasedEditOptions?.actionButtons?.editButtonTitle ||
```
ghiscoding, line 456: I also usually have a set of Locale translation keys for I18N, we could probably add them in the future though `editButtonTitleKey`, `deleteButtonTitleKey` and `updateButtonTitleKey`... (e.g.: `updateButtonTitleKey: 'UPDATE'`)

It's a lot of work to make everyone happy 😅 

but that might be ok to skip that since it's only for buttons?

packages/common/src/services/extension.service.ts
```diff
@@ -162,6 +166,21 @@ export class ExtensionService {
         this.translateItems(this.sharedService.allColumns, 'nameKey', 'name');
       }
 
+      // Row Based Edit Plugin
+      if (this.gridOptions.enableRowBasedEdit) {
+        this._rowBasedEdit = this._rowBasedEdit || new SlickRowBasedEdit(this.pubSubService, this.gridOptions.rowBasedEditOptions);
+        const gridService = this.lazyGridService?.();
+        if (!gridService) {
+          throw new Error('[Slickgrid-Universal] requires a GridService to be configured and available');
```
ghiscoding, line 174: I think we should add RowBaseEdit somewhere in the error message to advise the throw caller to the user

packages/common/src/services/extension.service.ts
```diff
@@ -62,6 +65,7 @@ export class ExtensionService {
     protected readonly sortService: SortService,
     protected readonly treeDataService: TreeDataService,
     protected readonly translaterService?: TranslaterService,
+    protected readonly lazyGridService?: () => GridService
```
ghiscoding, line 68: nice name since it really is lazy loaded, I'm still impressed by this approach 😉

ghiscoding (changes requested):

ghiscoding:
> yep I'll create one when writing the docs and attach it as comment. but first those nasty final two lines need coverage 😅

feel to use `istanbul ignore` if they're too hard to cover, I started using that for some core files which I'm not sure how to reach certain lines 😅

packages/common/src/extensions/__tests__/slickRowBasedEdit.spec.ts
```diff
@@ -0,0 +1,824 @@
...
+  it('should throw an error when "enableRowEdit" is set without "editable"', () => {
+    gridStub.getOptions.mockReturnValue({ enableCellNavigation: true });
+
+    expect(() => plugin.init(gridStub, gridService)).toThrow(
+      /(editable = true)/
+    );
+  });
+
+  it('should warn the user when autoEdit is not set or false and turn it on', () => {
+    const consoleSpy = jest.spyOn(console, 'warn');
```
ghiscoding, line 184: I just saw in the GitHub Action that this console is showing up, to avoid showing them in Jest, you can (should) use the following
```ts
const consoleSpy = jest.spyOn(global.console, 'warn').mockReturnValue();
```
image

examples/vite-demo-vanilla-bundle/src/examples/example22.ts
```diff
@@ -0,0 +1,223 @@
...
+        finish: new Date(randomYear + 1, randomMonth + 1, randomDay),
+        effortDriven: i % 5 === 0,
+      };
+    }
+
+    return mockDataset;
+  }
+
+  toggleSingleMultiRowEdit() {
+    this.sgb.gridOptions = this.sgb.gridOptions = {
```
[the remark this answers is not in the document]
zewa666: the double assignment is just a mistake. dunno how I managed to pull that off 😅

I dunno I just saw the other examples using sgb.gridOptions hence continuing in the same vein. I'll check up on that

examples/vite-demo-vanilla-bundle/src/examples/example22.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    this.gridOptions = this.sgb.gridOptions;
+  }
+}
+
+function fakeFetch(_input: string | URL | Request, _init?: RequestInit | undefined): Promise<Response> {
+  return new Promise((resolve) => {
+    setTimeout(() => {
+      resolve(new Response(JSON.stringify({ status: 200, message: 'success' })));
+    }, 500);
```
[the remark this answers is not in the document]
zewa666: its just one time to simulate the save delay, but nevertheless good idea to reduce it in case of testing

package.json
```diff
@@ -95,5 +95,8 @@
   "funding": {
     "type": "ko_fi",
     "url": "https://ko-fi.com/ghiscoding"
+  },
+  "prettier": {
+    "singleQuote": true
```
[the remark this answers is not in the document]
zewa666: I will as last cleanup but I'd ask that once the PR is ready you give it a try by yourself and tweak the value to your likings

packages/common/src/extensions/__tests__/slickRowBasedEdit.spec.ts
```diff
@@ -0,0 +1,824 @@
...
+        { item: fakeItem } as OnBeforeEditCellEventArgs
+      )
+    ).toBe(true);
+  });
+
+  it('should throw an error when "enableRowEdit" is set without "enableCellNavigation"', () => {
+    gridStub.getOptions.mockReturnValue({});
+
+    expect(() => plugin.init(gridStub, gridService)).toThrow(
+      /(enableCellNavigation = true)/
```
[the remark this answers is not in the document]
zewa666: say thanks to Copilot I didnt notice either 😅

packages/common/src/extensions/__tests__/slickRowBasedEdit.spec.ts
```diff
@@ -0,0 +1,824 @@
...
+
+      const cols = [...mockColumns];
+      plugin.create(cols, options);
+
+      const actionColumn = plugin.getColumnDefinition();
+      expect(cols.at(-1)?.name).toEqual(actionColumn.name);
+    });
+
+    [-1, 0, 2].forEach((position) => {
+      it(
```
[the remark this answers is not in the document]
zewa666: no actually I didnt. looks good. most of the times I'm fine with the simple loop but that gherkin style table they support looks quite convenient for a few use cases

packages/common/src/extensions/slickRowBasedEdit.ts
```diff
@@ -0,0 +1,596 @@
...
+
+      this._grid.setOptions({
+        excelCopyBufferOptions: {
+          ...options.excelCopyBufferOptions,
+          onBeforePasteCell: (e: SlickEventData<any>, args: OnEventArgs) => {
+            let userResult = true;
+            if (existingBeforePasteCellHandler) {
+              userResult = existingBeforePasteCellHandler(e, args);
+
+              if (userResult === false) {
```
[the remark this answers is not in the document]
zewa666: yep I wanted to explicitely make clear that this is the original user defined rules result as the handler name might not suggest that straight away. I prefer an additional variable over comments in these cases. but I can remove it if it bothers you

packages/common/src/extensions/slickRowBasedEdit.ts
```diff
@@ -0,0 +1,596 @@
...
+      this._grid.onSetOptions,
+      this.optionsUpdatedHandler.bind(this)
+    );
+    this._eventHandler.subscribe(
+      dataView.onRowsOrCountChanged,
+      this.handleAllRowRerender.bind(this)
+    );
+
+    this._grid.invalidate();
+    this._grid.render();
```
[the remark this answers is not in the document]
zewa666: yep just copied that behavior from other places. I thought you meant the double call explicitely

packages/common/src/extensions/slickRowBasedEdit.ts
```diff
@@ -0,0 +1,596 @@
...
+    );
+
+    return {
+      id: columnId,
+      name: this._addonOptions?.actionsColumnLabel,
+      field: 'action',
+      minWidth: 70,
+      width: 75,
+      maxWidth: 75,
+      excludeFromExport: true,
```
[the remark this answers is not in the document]
zewa666: good catch. actually I think instead of defining width and others we should accept `actionColumnConfig: Partial<Column<any>> and spread that over the previously applied. this way users can override and extens whatever they want

packages/common/src/extensions/slickRowBasedEdit.ts
```diff
@@ -0,0 +1,596 @@
...
+      if (prevSerializedValue !== serializedValue || serializedValue === '') {
+        const finalColumn = Array.isArray(editCommand.prevSerializedValue)
+          ? editorColumns?.[index]
+          : column;
+
+        if (!finalColumn) {
+          return;
+        }
+
+        this._grid.invalidate();
```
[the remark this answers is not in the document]
zewa666: oh thats great to know. looked for sonething like that already a couple of times

packages/common/src/extensions/slickRowBasedEdit.ts
```diff
@@ -0,0 +1,596 @@
...
+    }
+
+    if (!options?.editable) {
+      throw new Error(
+        `[Slickgrid-Universal] Row Based Edit Plugin requires the gridOption editable (editable = true)`
+      );
+    }
+  }
+
+  protected undoRowEdit(item: any) {
```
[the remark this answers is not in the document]
zewa666: actually I think this might make a great extension by itself. UndoRedoPlugin sort of which than could expose hooks and events for other plugins to consume/drive features.

that way there would be one feature-rich well tested implementation which users could also take for their own apps.

packages/common/src/extensions/slickRowBasedEdit.ts
zewa666: usually I'm the one being nitpicky about i18n ;) good catch

packages/common/src/services/extension.service.ts
zewa666: yeah lazy resolvers are a great thing over at C# standard DI and also great in Aurelia. Its a cool way to push initialization of transient tokens all the way to the point when they are actually needed.

In Angular its a bit cumbersome with Lazy Module loading but serves the same purpose.

the async nature on the other hand makes it a nice escape hatch for circular refs ;)

zewa666:
the best thing about prettier is its so damn conservative and restrictive. the worst thing with that is that feelings get hurt. I personally grew so tired of formatting discussions that it became like a godsend tool to me.

but I'm happy to change to whatever style you prefer. I just werent able to determine clear paths for certain constructs.

so I'd like to turn this around, once the PR is done feel free to reformat the whole file and I'll turn off prettier for this file like all the others

packages/common/src/extensions/slickRowBasedEdit.ts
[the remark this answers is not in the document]
ghiscoding: yeah that's another good idea, future, future...

ghiscoding:
@zewa666 yeah I think that I won't use Prettier in this project, it's getting quite large and I don't feel like reformatting the whole thing and revisit all test coverage. However, I do use Prettier in newer project, it's easier to maintain when the project initially started with it and when it isn't too large like the project. I'm not expecting to change the format for new files but yeah if you modify some existing file then it's probably better to disable Prettier or at the minimum enable it for "changed lines" instead of the whole file.

Just out of curiosity for my schedule and I don't want to be pushy but are you expecting to finish the PR by tomorrow? I'd like to go ahead with a new release tomorrow and I'm also guessing that a new PR that I just made might conflicts with yours so I'll wait a bit... but anyway even if it's not by tomorrow, I can still push later next week, no biggy 

Thanks a ton for this new contribution, this is for sure the biggest PR I've received since the creation of the project 🚀 ... I feel a little less lonely to maintain the project now (finally someone who takes time to read the code 😝) hehe

zewa666:
no hard feelings on that. yeah just used it for the new stuff, i dont use it in the other PRs or touching existing files.

No I dont think I'll make it until tomorrow. i'd like to properly finish the docs and your feedback and get in the couple of last details like proper coloring. If you have any conflicting PR just get them in so I can merge and resolve conflicts. I'll be done next week but again no urgency in getting this released.

this plugin was meant for getting into the codebase a little bit deeper and understanding how things are connected. I feel actually super confident after what I'm seeing. the architecture is really flexible and allows for tons of extension points. plus you've been extremely helpful with tips and guidance. thanks again for that.

packages/common/src/extensions/slickRowBasedEdit.ts
```diff
@@ -0,0 +1,596 @@
...
+            options.rowBasedEditOptions?.actionButtons
+              ?.iconCancelButtonClassName || 'mdi mdi-cancel color-danger',
+        })
+      );
+
+    return actionFragment;
+  }
+
+  protected onBeforeEditCellHandler = (
+    _e: Event,
```
ghiscoding, line 536:
```suggestion
    _e: SlickEventData,
```
because of recent PR #1327

ghiscoding:
Alright I'll go ahead and merge PR #1327 and hopefully won't cause too much trouble on your side, the PR is about using the correct `SlickEventData` when subscribing with `SlickEventHandler`, it was previously `.subscribe((e: any, args: ArgType) => {})` but now it's `(e: SlickEventData, ...)` which is the correct type, that required a lot of refactoring though which is why ~I'm expecting conflicts~ (nope it didn't add conflicts, great, but just 1 small change to do `_e: SlickEventData` on your side)


> this plugin was meant for getting into the codebase a little bit deeper and understanding how things are connected. I feel actually super confident after what I'm seeing. the architecture is really flexible and allows for tons of extension points. plus you've been extremely helpful with tips and guidance. thanks again for that.

I like to see that kind of comments :)  I do love the project, the person who created SlickGrid eventually stopped working on it and got hired by Google, his way of structuring the project was quite ahead of its time (it was pretty much the precursor of ES Module but using jQuery, which was pretty awesome at the time), it's super customizable and I'm trying to keep it that way as much as possible, I also see a lot more stars on the original `6pac/slickgrid` since we migrated to native & TypeScript. It's more work for me maintain (but it's good for visibility, more users, for example the latest PR that I've mentioned was an issue opened on that project that I've pushed in here too)

Next week is fine, I'll write some more unit tests (98% now) lol... enjoy your weekend ⛄

examples/vite-demo-vanilla-bundle/src/examples/example22.ts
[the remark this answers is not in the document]
zewa666: the approach with this.sgb.gridOptions and assigning it to this.gridOptions is the same as in example 11, 12, 14 and 19. I've just removed the wrong double assignment

zewa666:
I've went through your feedback and two things are open.

1. I didn't get what you meant was necessary to do for the i18n keys. Perhaps you could point me to an example?
2. The invlidateRow wasn't working as it introduced side-effects (empty row) so I kept the invalidate for now. Perhaps you want to look into that one

ghiscoding:
> I've went through your feedback and two things are open.
> 
> 1. I didn't get what you meant was necessary to do for the i18n keys. Perhaps you could point me to an example?
> 2. The invlidateRow wasn't working as it introduced side-effects (empty row) so I kept the invalidate for now. Perhaps you want to look into that one

Well I wasn't sure if these are used for button tooltips only? If that's the case then I guess we could skip that task entirely, I usually add locale translation keys as extra options for users who use I18N and typically it's set so that if the system finds that the property with a `Key` suffix is defined then it uses it, or if not then use default option. For example, if column picker option `columnTitleKey` is provided and `enableTranslate` is enabled then we use the key for translation, but if not we revert to using `columnTitle`

So I usually provide 2 options, the locale is with the `Key` suffix

https://github.com/ghiscoding/slickgrid-universal/blob/a9cb8ee3f1a5da4249851e5b701b027b3f72ad26/packages/common/src/interfaces/gridMenuOption.interface.ts#L10-L14

then in the code you can use `extensionUtility.translateItems` by providing an object array with the prop name (like `title`) to look for and the util will search for associated translation key by adding `Key` suffix to prop name (like `titleKey`) and translate it 

https://github.com/ghiscoding/slickgrid-universal/blob/a9cb8ee3f1a5da4249851e5b701b027b3f72ad26/packages/common/src/extensions/slickColumnPicker.ts#L174-L178

zewa666:
I like the idea of the overloading translation key. I'll definitely add it. yeah its just a tooltip but nevertheless it would this way follow the same pattern other plugins use.

zewa666:
aside from the two above I'd also like to ask you whether you could give the scss files a look. I'm not sure I caught all the places or if something else doesnt conform to the base idea.

packages/common/src/styles/slick-bootstrap.scss
```diff
@@ -107,6 +107,27 @@
...
       }
+
+      &.slick-rbe-editmode.active .slick-cell,
+      &.slick-rbe-editmode .slick-cell {
+        background-color: var(--slick-row-based-edit-editmode-bgcolor, $slick-row-based-edit-editmode-bgcolor);
+
+        &:hover {
+          background-color: var(--slick-row-based-edit-editmode-hover-bgcolor, $slick-row-based-edit-editmode-hover-bgcolor);
+
+          & .active {
```
ghiscoding, line 118: I don't think the space is necessary, right? `&.active` if it is necessary then remove the `&` I think

packages/common/src/styles/slick-grid.scss
```diff
@@ -84,6 +87,26 @@
...
        background-color: $slick-cell-odd-active-background-color;
      }
+
+     &.slick-rbe-editmode .slick-cell {
+      background-color: var(--slick-row-based-edit-editmode-bgcolor, $slick-row-based-edit-editmode-bgcolor);
+
+      &:hover {
+        background-color: var(--slick-row-based-edit-editmode-hover-bgcolor, $slick-row-based-edit-editmode-hover-bgcolor);
+
+        & .active {
```
ghiscoding, line 97: again is the space necessary?

zewa666:
ok I'm still not sure I really understand how to approach the i18n part.
Here's what I've tried so far:
[i18n.patch](https://github.com/ghiscoding/slickgrid-universal/files/13929602/i18n.patch)

the mentioned function `translateItems` isn't really suitable (array based) plus it doesn't really do anything besides calling translate.

ghiscoding:
ah sorry that wasn't a good explanation, without looking at the patch you linked (I'm supposed to be off on Sundays ;)). So anyway, I'll assume you might be looking for code when the locale changes, below is the code that calls that. 

https://github.com/ghiscoding/slickgrid-universal/blob/1d759f772fa3f1334f66b1cb76c37a1134f5c44d/packages/vanilla-bundle/src/components/slick-vanilla-grid-bundle.ts#L733-L752

So you probably need to add a method that will be called to from the callback above. 

If that is still not helping, just let me know and I'll take a proper look at the patch

zewa666:
ok think I got it working. There was still a nasty bug with Cypress but I got a workaround to force drop the old cached items.
Next up ... writing docs

examples/vite-demo-vanilla-bundle/public/i18n/fr.json
```diff
@@ -100,5 +100,7 @@
   "TITLE": "Titre",
   "TITLE.NAME": "Nom du Titre",
   "TRUE": "Vrai",
-  "X_DAY_PLURAL": "{{x}} journée{{plural}}"
+  "X_DAY_PLURAL": "{{x}} journée{{plural}}",
+  "RBE_BTN_UPDATE": "Mettre à jour la ligne actuelle",
+  "RBE_BTN_CANCEL": "Annuler la ligne actuelle"
```
ghiscoding, line 105: this is just a French thing but I think it would be better with `ligne courante` because `ligne actuelle` actually means `actual line` which is not exactly the best to use. I know it's Google Translate result that you used but I think it's wrong hehe. BTW, there's 2 of them to change 


image

packages/common/src/styles/slick-bootstrap.scss
```diff
@@ -107,6 +107,27 @@
...
+
+        &:hover {
+          background-color: var(--slick-row-based-edit-editmode-hover-bgcolor, $slick-row-based-edit-editmode-hover-bgcolor);
+
+          .active {
+            background-color: var(--slick-row-based-edit-editmode-active-hover-bgcolor, $slick-row-based-edit-editmode-active-hover-bgcolor) !important;
+          }
+        }
+
+        & .active {
```
ghiscoding, line 123: I think you missed that one when I mentioned it couple days ago

packages/common/src/styles/slick-grid.scss
```diff
@@ -84,6 +87,26 @@
...
+
+      &:hover {
+        background-color: var(--slick-row-based-edit-editmode-hover-bgcolor, $slick-row-based-edit-editmode-hover-bgcolor);
+
+        .active {
+          background-color: var(--slick-row-based-edit-editmode-active-hover-bgcolor, $slick-row-based-edit-editmode-active-hover-bgcolor) !important;
+        }
+      }
+
+      & .active {
```
ghiscoding, line 102: same as previous, `&` can be removed

packages/common/src/extensions/slickRowBasedEdit.ts
```diff
@@ -229,6 +239,45 @@ export class SlickRowBasedEdit {
...
   }
 
+  translate() {
+    this.btnUpdateTitle =
+      (this.gridOptions.rowBasedEditOptions?.actionButtons?.updateButtonTitleKey &&
+        this.extensionUtility.translaterService?.translate?.(
+          this.gridOptions.rowBasedEditOptions?.actionButtons?.updateButtonTitleKey
+        )) ||
+      this.gridOptions.rowBasedEditOptions?.actionButtons?.updateButtonTitle ||
+      'Update the Row';
```
ghiscoding, line 249: I see a lot of the same code repeated, perhaps we could migrate this to an external protected function? Something along these lines

```ts
protected getTitleOrDefault(key: string, defaultTitle: string) {
  const actionBtnOptions = this.gridOptions.rowBasedEditOptions?.actionButtons; 
  const hasTranslateService = this.extensionUtility.translaterService?.translate;

  // another approach might be to move the line above to GETTERs in the class?
  
  return hasTranslateService ? actionBtnOptions[key] : defaultTitle;
}

this.btnUpdateTitle = this.getTitleOrDefault('updateButtonTitleKey', 'Update the Row');
```

I should probably do that in some other extensions like the GridMenu that I think I did something like you did and I think that this could make the code more readable.

ghiscoding:
this is out of context but, it's a nice story... so I keep adding unit tests for the core files and it's not exactly a fun exercise but I keep find and fixing small bugs and remove a bunch of unused or unreachable code. So it does seem to be worth the work and a worth exercise (now at 99.4% coverage, the end of the tunnel is in sight woohoo)

zewa666:
yeah its great when you see that type of payoff. I wasnt able to invest time the last few days sadly but hope to manage it in the next few ones.

zewa666:
hey with the latest commit I've added again a bit of TS type inference helpers.

image

Way back than I've added already some to create the Column defintion type. So my question is whether it's ok to leave them all sprinkled around or you'd want to co-locate them all in one ts-utils.ts file sort of

ghiscoding:
> hey with the latest commit I've added again a bit of TS type inference helpers.
> 
> image
> 
> Way back than I've added already some to create the Column defintion type. So my question is whether it's ok to leave them all sprinkled around or you'd want to co-locate them all in one ts-utils.ts file sort of

ah yeah I saw the inference type you added, at first glance I thought it was overkill since it's only used once hahah but nonetheless it's quite powerful and you can't go wrong with that approach. Feel free to put it where you think is best, when I started the project, I used to create a file per interface but I see my folder is huge now and I started grouping them by feature lately depending on what it is. So you could add it to `interfaces/rowBasedEditOption.interface.ts` if it's only used for that or create something like `inference.helper` or whatever and if that is what you go with then maybe move the ones you've done in the past from `column.interface`. I don't have any preferences for that, go with what you think is best

https://github.com/ghiscoding/slickgrid-universal/blob/03cad4a34bf13a8e1342306f9210525f5025321f/packages/common/src/interfaces/column.interface.ts#L18-L31

zewa666:
yeah I'm a bit nervous when magic strings are floating around as my future self isn't that mindful of knowing what the intention was. So it serves as a real well gatekeeper to focus the type and be more expressive (all of that just to save one line of comment 🤣 )

Nevertheless the docs are now also added. A quick review on those would be very appreciated.

docs/TOC.md
```diff
@@ -54,6 +54,7 @@
 * [Pinning (frozen) of Columns/Rows](grid-functionalities/frozen-columns-rows.md)
 * [Row Selection](grid-functionalities/Row-Selection.md)
 * [Tree Data Grid](grid-functionalities/Tree-Data-Grid.md)
+* [Row Based Editing Plugin](grid-functionalities/Row-based-edit.md)
```
ghiscoding, line 57: doesn't need to be changed but just as a side note, I want to use only lowercase for new docs. I haven't changed them all because I'm on Windows and Git doesn't detect case change by default (there's some extra git command to run which I forgot), so just a reminder for future docs...

docs/grid-functionalities/Row-based-edit.md
```diff
@@ -0,0 +1,70 @@
...
+- [Change Dynamically Single/Multiple Selections](#changing-dynamically-from-single-to-multiple-selections-and-vice-versa)
+- [Mixing Single & Multiple Row Selections](#mixing-single--multiple-row-selections)
+- [Disable Custom Rows Selections via `selectableOverride`](#disable-custom-rows-selections-via-selectableoverride)
+- [Disable External Button when having Empty Selection](#disable-external-button-when-having-empty-selection)
+- [Change Row Selections](#change-row-selections)
+- Troubleshooting
+  - [Adding a Column dynamically is removing the Row Selection, why is that?](#adding-a-column-dynamically-is-removing-the-row-selection-why-is-that)
+
+### Description
+The Row based editing plugin makes it possible to keep the grid readonly except for rows which the user explicitely toggles into editmode.
```
ghiscoding, line 13: is "editmode" a property? If so, wrap it in backtick or if not then maybe separate it in 2 words?

docs/grid-functionalities/Row-based-edit.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Support for the Excel Copy Buffer Plugin
+If the [Excel Copy Buffer Plugin](grid-functionalities/excel-copy-buffer.md) is configured, the Row based edting pluging will override it's behavior by denying pastes on all cells not within a editmode row. Nevertheless, any existing `BeforePasteCellHandler` will be respected.
+
+## How the plugin works
+The idea of the plugin is to focus the users editing experience on specific individual rows and and save them individually. This is achieved by letting the user toggle one or more rows into editmode. Now changes can be made to those rows and will be highlighted and tracked. The user may cancel the editmode at any time and revert all cells changes. If the save button is pressed on the other hand an `onBeforeRowUpdated` hook, which you define via plugin options, is called and expects a `Promise<boolean>`. In that method you'd typically write the changes to your backend and return either true or false based on the operations outcome. If a negative boolean is returned the editmode is kept, otherwise the row applies the changes and toggles back into readonly mode.
```
ghiscoding, line 33: "...and toggles back into readonly mode."  so just to confirm, when this edit mode is used, the grid itself cannot be edited the typical way (click on any cell), we really need to use the edit mode, is that it? We might need to add a sentence explaining what I just asked, it wasn't clear to me

docs/column-functionalities/Editors.md
```diff
@@ -369,4 +369,8 @@ For SalesForce it's nearly the same, the only difference is that we add our even
 <div class="grid-container slds-p-horizontal">
     <div class="grid1" onslickergridcreated={handleOnSlickerGridCreated}>
     </div>
-</div>
\ No newline at end of file
+</div>
+```
+
+## Turning individual rows into editmode
+Using the [Row Based Editing Plugin](../grid-functionalities/Row-based-edit.md) you can let the user toggle either one or multiple rows into editmode, keep track of cell changes and either discard or save them on an individual basis using a custom onBeforeSave hook.
```
ghiscoding, line 376: probably better to escape `onBeforeSave` hook with backtick

docs/grid-functionalities/Row-based-edit.md
zewa666: exactly. i've added that in the description at the begin. do you want to repeat the info here to be on the safe side?

ghiscoding: yeah might as well to be clearer since I didn't notice the top description

zewa666:
alright I think I got everything in that I've planned for. How do you want to proceed from here?

ghiscoding (approved):

ghiscoding:
looking good.... let's hit the Merge button 🚀 

I think the next steps will be to replicate the new Example into the external repos (Angular-Slickgrid, ...), did you wanted to me to do that or did you wanted to do it too (for Angular I mean)?

zewa666:
I can do so for Angular whenever you release slickgrid universal and bump it as dependency in Angular-slickgrid.

PR #1346, opened 2024-01-18
chore: add missing tests for RowBaseEditor & tweak UI

ghiscoding:
- change fetch response alert with notification in UI
- add missing unit tests for translation

### DIFF
--- a/examples/vite-demo-vanilla-bundle/src/examples/example22.html
+++ b/examples/vite-demo-vanilla-bundle/src/examples/example22.html
@@ -69,4 +69,17 @@
   </div>
 </section>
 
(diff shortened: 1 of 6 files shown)


ghiscoding:
@zewa666 I looked at your new Example 22 and tweaked the UI a bit and added missing translation unit tests. However while doing that I saw that the "Duration > 40" doesn't always work, it doesn't seem to always detect the error if we use the mouse for the next cell, it seems to work better with Enter key. Sometime the error doesn't show and sometime even fixing the error it doesn't want to go away (see animated gif)

msedge_VSMYI2x0Hi

PR #1353, opened 2024-01-19
feat(editor): auto commit before save; add `onBeforeEditMode` callback

zewa666:
This addresses the latest review feedback by adding the onBeforeEditMode callback and commiting active changes before running save.

related issue #https://github.com/ghiscoding/slickgrid-universal/issues/1299


ghiscoding:
awesome, so I assume the Duration problem is all good since you use the editor commit?

zewa666:
yep. additionally I've made sure to only call it if the respective save button is the same as the active editor (multirow edit).

hmm seems example16 is failing in e2e tests. will have to check whehter its due to my changes

ghiscoding:
> hmm seems example16 is failing in e2e tests. will have to check whehter its due to my changes

no forget about it, it's a flaky Cypress tests, it fails every 10 runs or so. The auto-scroll is not entirely consistent, I typically just rerun the test and it passes every time. I wish that Cypress had an option to rerun the entire spec file but they don't, they only offer `it` block rerun which is not enough in this spec file

ghiscoding:
Totally out of topic but didn't want to open a new discussion or anything... I just added a new Column option `reorderable` to lock certain columns (like checkbox selector for row selection). I added it first in SlickGrid and will add it to Slickgrid-Universal tomorrow before releasing, see this SlickGrid [PR](https://github.com/6pac/SlickGrid/pull/983), anything to comment feel free to leave a comment in the PR (will merge tomorrow, time for bed now it's 2AM)

PR #1357, opened 2024-01-20
feat: add column `reorderable` option to optionally lock a column

ghiscoding:
- we can take advantage of SortableJS `filter` option which can be used to filter column header with a certain CSS class. When using `reorderable: false` on a column, SlickGrid will add configurable CSS class to the column header which will be filtered out by SortableJS hence locking the column from being reorderable
- new options available:
  - Column: `reorderable` (defaults to true)
  - Grid Option: `unorderableColumnCssClass` (defaults to "unorderable")
  - CSS/SASS variable for Alpine Theme: `$slick-grid-header-unorderable-bg-color` (replace `$` with `--` for CSS var)
- also referencing an old PR #64 that brought the original idea

#### ⚠️ Warnings
I did add and set `reorderable: false` to the following 4 plugins, they can all be overriden when instantiating the plugin via their respective options
- Checkbox Selector Column (Row Selection)
- Row Detail
- Row Move Manager
- Row Base Editor

brave_RO4KUzXlJ7


ghiscoding:
@zewa666 I added it to Row Base Editor as well and changed default to `reorderable: false`, all 4x plugins that I've disabled it all have them as a configurable option to each plugin

zewa666:
thats great. how did you came to think of this feature?

ghiscoding:
> thats great. how did you came to think of this feature?

haha I was just looking at some of opened issues in SlickGrid and see if I can work on some of them and then I remember that old issue to lock columns and took a shot at it. It's easier now that we use [SortableJS](https://sortablejs.github.io/Sortable/), they have a good API with an easy `filter` option. However that locking doesn't have the expected effect when the column is in the middle since other columns can be moved, hence why it works better when column is first/last.

PR #1359, opened 2024-01-22
perf(plugins): decrease number of calls to translate all extensions only once

ghiscoding:
found a couple of places in the code where it was translating the same things more than once
- `translateAllExtensions()` is calling translate on all extensions including a call to `translateColumnHeaders()` but that one was also calling GridMenu translates which was already translated, instead we can expect that `translateColumnHeaders()` should also indirectly translate both ColumnPicker/GridMenu only once inside `renderColumnHeaders()` when reassigning translated `columns`
- also `translateColumnHeaders()` will indirectly call a `grid.invalidate()` so I think this call should be made last in `translateAllExtensions()` so that everything else had a chance to be translated prior to reaching the invalidate
- `onLanguageChange` is calling `translateAllExtensions()` but was also calling `translateColumnHeaderTitleKeys()` and `translateColumnGroupKeys()`, however these last 2 were already covered by `translateColumnHeaders()` so they can be removed
- also a rare but potential infinite loop could happen if `translateColumnHeaders()` is provided with a lang, which then triggers a `onLanguageChange` which then calls back `translateColumnHeaders()` which could become infinite, so we should make sure that we only trigger a lang change event only when we see the lang being different
- another small perf improvement, there's no need to check if the translateXYZ methods exists on each extensions because we already know it exists (that code was put in place when we were using the external `6pac/SlickGrid` dependency, but now everything is internal and we know all these methods already exists since it must follow the interface), e.g.
```diff
translateGridMenu() {
-    this._gridMenuControl?.translateGridMenu?.();
+    this._gridMenuControl?.translateGridMenu();
}
```

### DIFF
--- a/examples/vite-demo-vanilla-bundle/src/translate.service.ts
+++ b/examples/vite-demo-vanilla-bundle/src/translate.service.ts
@@ -13,7 +13,7 @@ interface TranslateOptions {
 export class TranslateService implements TranslaterService {
   eventName = 'onLanguageChange' as TranslateServiceEventName;
   protected _currentLanguage = 'en';
-  protected _locales: { [language: string]: Locales } = {};
+  protected _locales: { [language: string]: Locales; } = {};
   protected _pubSubServices: PubSubService[] = [];
   protected _options;
   protected templateMatcher = /{{\s?([^{}\s]*)\s?}}/g;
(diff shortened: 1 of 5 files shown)


ghiscoding:
@zewa666 here's is a first PR to fix multiple translate calls, I wasn't sure if I should name this PR as `fix` or `perf` because it really does both. You can review if you want, I'll probably merge soon anyway

zewa666:
its good to know the tracing of calls to invalidate but I wonder whether this method perhaps should be debounced so that multiple calls are batched into a single one

ghiscoding:
> its good to know the tracing of calls to invalidate but I wonder whether this method perhaps should be debounced so that multiple calls are batched into a single one

I don't really want to add debounce since that would require 1 more cycle before the invalidate kicks in, and this PR is about translation calls, not really invalidate which I'm looking at to do in another PR. There's no call to invalidate at all in this current PR, I did mention about it in the description but it's mainly about the order of execution

zewa666:
yeah I got that and the additional cycle is yet another argument.  the core fix in itself makes absolut sense as there's no reason for that additional emit if lang stays the same

ghiscoding:
I think the PR is ok, so let's merge and a follow up PR is coming for `invalidate()`

PR #1360, opened 2024-01-23
perf(plugins): Row Base Editing tooltip title translate only once

ghiscoding:
- when adding console logs to see why some of the buttons weren't translated, I discovered that it was calling the translate function on every single dataset row because of the Formatter, however once we have the translation we shouldn't have to retranslate the same tooltip texts over and over

### DIFF
--- a/packages/common/src/extensions/slickRowBasedEdit.ts
+++ b/packages/common/src/extensions/slickRowBasedEdit.ts
@@ -32,6 +32,13 @@ export interface EditedRowDetails {
   cssStyleKeys: string[];
 }
 
+interface ButtonTranslation {
+  btnUpdateTitle: string;
+  btnEditTitle: string;
+  btnDeleteTitle: string;
+  btnCancelTitle: string;
+}
+
 /**
  * Row based edit plugin to add edit/delete buttons to each row and only allow editing rows currently in editmode
  */
@@ -52,10 +59,8 @@ export class SlickRowBasedEdit {
   protected _editedRows: Map<string, EditedRowDetails> = new Map();
 
   private _existingEditCommandHandler: ((item: any, column: Column<any>, command: EditCommand) => void) | undefined;
-  private btnUpdateTitle: string = '';
-  private btnEditTitle: string = '';
-  private btnDeleteTitle: string = '';
-  private btnCancelTitle: string = '';
+  protected _currentLang = 'en';
+  private _translations: { [locale: string]: ButtonTranslation; } = {};
 
   /** Constructor of the SlickGrid 3rd party plugin, it can optionally receive options */
   constructor(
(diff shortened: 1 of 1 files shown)


ghiscoding:
@zewa666 I definitely need a review on this one, I'm open to feedback but the main idea is to not translate the same texts over and over (which was what the previous code was doing through the Formatter). Another approach might be to use a Map but using an Object is simple enough and for 4 texts, I don't think using either or will make any difference in perf

zewa666:
the dictionary approach is fine enough. as you said only 4 keys. what I'm a bit concerned though is why the plugin has to keep track of the currentlng. this doesnt seem to be a concern of it and might be better suited in a general place like some translation service or the likes. otherwise all plugins start to do the same thing.

the init was an original code from the boilerplate plugin I've copied this from. since the translations are added and they are now called upfront there shouldnt be a need for the call in init. yep cypress tests should verify that

ghiscoding:
> what I'm a bit concerned though is why the plugin has to keep track of the currentlng

it's just for creating the language key to keep the translated value and English by default
```ts
{
  en: { btnEdit: 'Edit', ... },
  fr: { btnEdit: 'Editer', ... },
}
```
I also thought of it if it was necessary but I don't see other ways to keep track of the texts

zewa666:
but isnt there already some service where you could call getCurrentLanguage?

ghiscoding:
I'm not sure what you mean, I don't store the current language anywhere. If you have other ways of doing please let me know, since I'm not sure how else at this point and it's getting late so time for bed..

packages/common/src/extensions/slickRowBasedEdit.ts
```diff
@@ -52,10 +59,8 @@ export class SlickRowBasedEdit {
   protected _editedRows: Map<string, EditedRowDetails> = new Map();
 
   private _existingEditCommandHandler: ((item: any, column: Column<any>, command: EditCommand) => void) | undefined;
-  private btnUpdateTitle: string = '';
-  private btnEditTitle: string = '';
-  private btnDeleteTitle: string = '';
-  private btnCancelTitle: string = '';
+  protected _currentLang = 'en';
```
zewa666, line 62: I meant this one.

packages/common/src/extensions/slickRowBasedEdit.ts
```diff
@@ -241,16 +246,26 @@ export class SlickRowBasedEdit {
...
-
-    const viewport = this._grid.getViewport();
+  /**
+   * Translate with current locale when `enableTranslate` is set or use title texts provided by user.
+   * We want to translate only once to avoid having to translate on each dataset item, we'll just reuse the translations if it was already translated earlier.
+   * User could optionally force a retranslate even if it was already translated
+   * @param {Boolean} [forceRetranslate] - even if it was already translate, force a retranslate
+   */
+  translate(forceRetranslate = false) {
+    this._currentLang = this.extensionUtility.translaterService?.getCurrentLanguage() ?? 'en';
```
zewa666, line 256: if its just for readability we could turn it into a getter but I'd ultimately always fetch it from extensionUtility

zewa666 (commented):
left inline comments

packages/common/src/extensions/slickRowBasedEdit.ts
ghiscoding: I know you meant that one but the reason to keep it under a variable is to avoid having to calling `getCurrentLanguage()` in the Formatter for every row of the dataset which was mainly the concern of this PR

packages/common/src/extensions/slickRowBasedEdit.ts
ghiscoding: the reason I've done this way is first because that is exactly how you called it in the `getTitleOrDefault()`, if I refactor this one then it would be better to refactor both

zewa666:
ok I was under the assumption that the utils service already short circuits the current lang by storing it by himself and offering it via the get function whereas a subscription would update it.

ghiscoding:
> ok I was under the assumption that the utils service already short circuits the current lang by storing it by himself and offering it via the get function whereas a subscription would update it.

it doesn't since it's not really its job but on the other hand it's the Translater Service job to keep the current lang. The ExtensionUtility Service is just a few utils functions, I don't remember why I've converted it to a TS Class instead of simply exporting functions since it doesn't have any local variables... Taking a quick look at it, I think it's a Class because it requires access to other Service Instances from DI

https://github.com/ghiscoding/slickgrid-universal/blob/cab68989ebd53178dfcee5ed293379dc8932a72f/packages/common/src/extensions/extensionUtility.ts#L8-L13

ghiscoding:
I'll go ahead with current change and push a release shortly, feel free to submit a PR in the future if you think it could be improved.

PR #1362, opened 2024-01-23
chore: use JSON Cypress test report files to publish report summary

ghiscoding:


ghiscoding:
doesn't really work and is rather complex for the potential small gain so let's close this

PR #1382, opened 2024-02-10
fix: mouse cell selection with active editor

### LINKED ISSUE #1380: Mouse multi cell selection does not work with active editor
### Describe the bug

As we recently found out in Example19, while an Editor is open we cant select multiple cells with the mouse. it works with the keyboard shift + arrow navigation though.

I guess its the editor stealing the focus but I'll investigate and provide a PR

### Reproduction

* example19
* activate editor in a cell
* try selecting it plus a few neighbors using the mouse

Vanilla / Plain JS

### Environment Info

```shell
```

### Validations

zewa666:
this fixes errors preventing to do a multi cell selection while an editor is opened.

Closes #1380


packages/common/src/extensions/__tests__/slickCellRangeSelector.spec.ts
```diff
@@ -701,4 +705,52 @@ describe('CellRangeSelector Plugin', () => {
       done();
     }, 7);
   });
+
+  it.only('should maintain propagation only if editor is on current cell', () => {
```
ghiscoding, line 709: oops looks like you forgot to remove `.only` after you're done 😉

zewa666: yeah thats super unfortunate ... I've added eslint-plugin-jest to detect those. Than started to fix the main ones by adjusting the the eslintignore file. As soon as I touched __tests__ though, ~250 errors came up. So those should be handled in a separate shot not to mess up this PR

ghiscoding: hmm yeah that's just linting cosmetic though, it should indeed be better to add ESLint stuff as a separate PR

zewa666:
ok I think I got it set now. so what I've essentially changed is to apply previous behaviors only if the selected/iterated cell isn't the active editor cell. That means you still can't do a multi-select starting from the editor cell but it works for all others.

ghiscoding:
> ok I think I got it set now. so what I've essentially changed is to apply previous behaviors only if the selected/iterated cell isn't the active editor cell. That means you still can't do a multi-select starting from the editor cell but it works for all others.

that seems fine by me, I'm assuming most of the time it will be from another cell anyway. Thanks for the fix :)

PR #1383, opened 2024-02-10
feat(ExcelExport): migrate to Excel-Export-Vanilla (ESM)

ghiscoding:
migrate to [Excel-Builder-Vanilla](https://github.com/ghiscoding/excel-builder-vanilla) which is a fork of `excel-builder.js` rewritten and modernized. It was rewritten as TypeScript, drops `Q` and `Lodash` dependencies and also swapped `JSZip` with `fflate` since it offers an ESM build with Tree Shaking support and better perf by using Web Workers. 

So in summary, we get a much smaller build, better perf and 1 less CJS dependency to take care about, all win-win 🚀 

#### Size Diff
_file on disk on Windows OS, this is only for comparison and is expected to be much smaller with gzip_
| [SF Bundle Zip File](https://github.com/ghiscoding/slickgrid-universal/tree/master/packages/vanilla-force-bundle/dist-grid-bundle-zip) | 590Kb (604,738) | 531Kb (544,733) | 59Kb | 9.92% smaller |
| [GitHub Vite Demo Website](https://ghiscoding.github.io/slickgrid-universal/) | 4.81Mb (5,048,893) | 4.63Mb (4,857,095) | 0.18Mb (or 192Kb) | 3.8% smaller |


ghiscoding:
cc @zewa666 :)

PR #1385, opened 2024-02-12
chore: introduce husky commit-msg hook and pr title linter

zewa666:


ghiscoding:
will it check for current PR? I mean even your current PR doesn't have the scope so it should fail in the future, is it? 😝

zewa666:
@ghiscoding hehe I'm just playing with it right now to get everything setup properly

ghiscoding:
ok let me know when it's ready to be merged, looks good from my point of view... also note that I pushed my new release already during lunch time and will do the other repos later tonight after work :)

ghiscoding:
> I'm not sure if you're following standard conventional-changelog so I kept the defaults as are. If you'd like to tweak things lemme know. Aside from that I'm done with this one and its ready for merge

Yeah that is indeed my preference, I use Lerna-Lite with the `--conventional-changelog` enabled, so it does follow the conventional changelog. I've been using conventional changelog for years but never took the time to setup Husky, this is great, we'll have to add them to other repos as well (Angular-Slickgrid, Aurelia-Slickgrid, ...)

https://github.com/ghiscoding/slickgrid-universal/blob/492dd06578a78ecbf9492e55a4dcf2bc97fdcf32/lerna.json#L14-L21

zewa666:
I was exactly thinking the same ... it's merely build tooling but ultimately wasn't sure. Good to have the re-confirmation.

yeah Husky is a great little tool full of helpful bits and pieces. Used it for tons of things like license checks, coverage mins etc.

ghiscoding:
Awesome, it's always to have contributions by power users like you :)  Thanks again!

ghiscoding:
@zewa666 I had to adjust the commitlint rules because it failed on me because the body content went over 100 chars but I don't really care about the body, I do care about the title not being more than 72 chars though. So anyway, I adjusted as below, but I also would like to know if this can be moved to maybe an `.mjs` file? I'm trying to get away from CJS as much as possible and that should include config files too. I've never extended a config other than below so I'm not even sure if it's possible?

https://github.com/ghiscoding/slickgrid-universal/blob/4ed241b7f5c00791e85986bfcef238a507f8c280/commitlint.config.js#L1-L7

Also another side note, I upgraded to the new `@typescript-eslint` v7 that got released today and the PR linter failed because it expect correct peer dependencies installed and that is no longer true with v7 (probably because not all dependencies are at v7 yet), so I switched to `npm install --force` to get rid of the error. However I also see that you fixed `@commitlint/config-conventional` at v17 even though I have v18 in the root `package.json`

ghiscoding:
ahh so after searching on commitlint GitHub, I found that this works 
> I changed `module.exports =` to `export default` and it worked for me.

however, `pull-request-name-linter-action` doesn't support it yet :(

[How to work with commitlint.config.ts (TypeScript config file)](https://github.com/JulienKode/pull-request-name-linter-action/issues/235)

ghiscoding:
@zewa666  ahh gosh I got a much bigger problem, it looks like Husky screwed up my release. I'm completely missing the typical git push. I think Husky blocked it for some reason but didn't exit. The release was created just before you added the PR linter. 

I typically have a commit of this shape `chore(release): publish new version 4.3.1` (here's the previous [commit](https://github.com/ghiscoding/slickgrid-universal/commit/aeade0564c11fbdca4997ee0470c7432376911a5) one from 2 weeks ago), it didn't create one for today's release for v4.4.0 and I don't see much in the CI but I do see Husky being triggered. I think it should be completely disabled when pushing a new release. I thought only my changelog weren't updated so I changed them by hand but then I realized the entire release commit is missing... This is bad... 

image
[chore(release): publish new version 4.3.1](https://github.com/ghiscoding/slickgrid-universal/commit/aeade0564c11fbdca4997ee0470c7432376911a5)`

PR #1389, opened 2024-02-12
chore: switch to ESM commitlint.config.ts

ghiscoding:


ghiscoding:
will have to close since `pull-request-name-linter-action/issues` doesn't support TS config files yet :(

 [How to work with commitlint.config.ts (TypeScript config file)](https://github.com/JulienKode/pull-request-name-linter-action/issues/235)

PR #1390, opened 2024-02-13
fix(ci): disable Husky when pushing new release

ghiscoding:
- as per issue identified in this [comment](https://github.com/ghiscoding/slickgrid-universal/pull/1385#issuecomment-1940388130)
- Husky that was added recently and caused an issue with the last release, the typical `chore(release)` commit was completely missing in today's release, possibly because of a racing condition or a prompt that cannot be answered in a CI environment
- as per Husky [docs](https://typicode.github.io/husky/how-to.html#ci-server-and-docker) we can disable Husky via 
```sh
env:
  HUSKY: 0
```


zewa666:
good morning. super sorry for that didnt think that you're performing git pushes on ci during your release. glad you figured it out though

ghiscoding:
yeah just piece it back together, it's kinda late here though hehe

PR #1427, opened 2024-03-08
fix: Join type

### LINKED ISSUE #1422: TypeScript type error cannot assign `unknown` to string...
### Describe the bug

I think the upgrade to the new TypeScript version created this new issue but only after I deleted the pnpm lock file and reinstalled everything

https://github.com/ghiscoding/slickgrid-universal/blob/5fab1792cfdf24172bd46556b6fe5513c93d19d1/packages/common/src/interfaces/column.interface.ts#L24-L28

image

### Reproduction

pnpm install and then run a TypeScript build

Vanilla / Plain JS

### Environment Info

```shell
n/a
```

### Validations

zewa666:
fixes #1422

### DIFF
--- a/packages/common/src/interfaces/__tests__/column.interface.spec.ts
+++ b/packages/common/src/interfaces/__tests__/column.interface.spec.ts
@@ -1,0 +1,65 @@
+import type { Join, PathsToStringProps } from '../column.interface';
+
+type Expect<T extends true> = T;
+type Equal<X, Y> = (<T>() => T extends X ? 1 : 2) extends <
+  T
+>() => T extends Y ? 1 : 2
+  ? true
+  : false;
+
+const typeCheckValue = <T>(arg: T) => arg;
+
+describe('the Join type', () => {
+  it('should concatenate strings with a separator', () => {
+    const sut = 'a.b.c';
+    type TypeToTest = Join<['a', 'b', 'c'], '.'>;
+    const value = typeCheckValue<TypeToTest>(sut);
+
+    type Check = Equal<typeof value, typeof sut>;
+    type Result = Expect<Check>;
+
+    const x: Result = true;
+    expect(x).toBe(true);
+  });
+
+  it('should concatenate mixed types with a separator', () => {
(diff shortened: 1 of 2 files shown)


zewa666:
I think this should do it. I couldn't really reproduce your original issue but, removing the // @ts-ignore reveleaded it as well for me. So now that one is also gone.

I've also made sure to add some tests if we need to test that further. So far my custom type assertions were enough but we might should go with something like tsd if we want to do more proper checks.

zewa666:
ahh ok now I got ya. It doesn't fail in the IDE but when running `pnpm run bundle`. ok need to check more on that

ghiscoding:
> ahh ok now I got ya. It doesn't fail in the IDE but when running `pnpm run bundle`. ok need to check more on that

hmm actually it failed when I ran the full Build and after going into the file, I did see the error in the IDE/VSCode as shown in the issue print screen. Also FYI, the reason this came up was that I deleted the pnpm lock file and deleted node_modules, because I wanted get rid of the `ip` package security issue (when a lock file exists, it sometime doesn't pick up version patch fixes).

zewa666:
ok `rm -rf node_modules` might was the missing thing for me. Anyways, good the bundler caught it as well

ghiscoding:
> ok `rm -rf node_modules` might was the missing thing for me. Anyways, good the bundler caught it as well

I'm a bit surprised, I thought that the lock file was enough to do the update, however I'm not even sure since when this error actually existed. I thought it was the latest TypeScript 5.4 that was released a couple days but as it turns out, the previous version also failed on this type... anyway, as long as we fixed it for the future :)

BTW, I saw Cypress failing again so I restarted it

zewa666:
well if you can't repro it locally and the bundler cant either plus the original intent still is in place ... I'd say we're good until the next time ;)

ghiscoding:
Thanks a lot for this PR, you even added more unit tests which is always very welcome... again thx for all your contributions :)

BTW, have you had a chance to try the Dark Mode yet? 😉

PR #1448, opened 2024-03-29
fix(tooltip): allow multiple tooltips per grid cell

ghiscoding:
- prior to this PR, we could only have 1 tooltip per cell, but in some occasions, we might have a cell with multiple icons & tooltips, if that happens then we should have separate tooltips (with their dedicated & different tooltipt text) positioned on that inner icon

#### TODOs
- [x] possibly reposition the arrow at the mouse location?
- [x] possibly center tooltip with center of target element?
- [x] test with multiple tooltips for header row/filter row elements as well
  - this actually requires new SlickEvents because the existing events for header/headerRow are using `mouseenter` but we need `mouseover` to differentiate multiple tooltip within the same header (`mouseenter` only gets triggered once while `mouseover` gets triggered multiple times while hovering any of element within container). So instead of possible chaning the previous behavior, I decided to duplicate `onHeaderMouseEnter` to `onHeaderMouseOver` and the same goes for `onHeaderRowMouseEnter` and also `...MouseOut` for both header types too (so adding 4 new events to cover all use cases). The funny thing is that the event for the cell `onMouseEnter` is actually using `mouseover` which is not actually matching with its name.
- [x] add new option to align tooltip to target element (default to false unless we have multiple tooltips, in that case it's always true)
- [x] unit test coverage
- [x] E2E test coverage

> for now the POC is mostly working, note that I purposely set a negative top margin in order to also test tooltip that can sometime show at the bottom (in 95% of the case it's showing on top except when there's not enough space on top)

### Example 22 (align right)
brave_9GJWaisxNl

### Example 11 (align left)

brave_3DHk9zGNbF

### Example 12 (multiple tooltips in header row)

brave_lvAr9wqtJK

## existing bugs
- [x] found existing bug when opening slider editor (not a new bug, this one existed before opening this PR)
brave_AvWeTzevAt
- [x] another bug when tooltip text is too wide, we should use text overflow ellipsis
image

### DIFF
--- a/docs/events/Available-Events.md
+++ b/docs/events/Available-Events.md
@@ -147,9 +147,13 @@ handleOnHeaderMenuCommand(e) {
   - `onHeaderContextMenu`
   - `onHeaderMouseEnter`
   - `onHeaderMouseLeave`
+  - `onHeaderMouseOver`
+  - `onHeaderMouseOut`
   - `onHeaderRowCellRendered`
   - `onHeaderRowMouseEnter`
   - `onHeaderRowMouseLeave`
+  - `onHeaderRowMouseOver`
+  - `onHeaderRowMouseOut`
   - `onKeyDown`
   - `onMouseEnter`
   - `onMouseLeave`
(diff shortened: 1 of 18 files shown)


ghiscoding:
@zewa666 I got a POC mostly working, however I did find some separate/existing bugs with the tooltip plugin and I have identified a few things to potentially do in the TODOs list above. Feel free to provide feedback, you could also give it a try by using this awesome new individual PR Stackblitz link

PR #1456, opened 2024-04-05
feat(common)!: migrate from `moment` to `moment-tiny`

ghiscoding:
- so I created yet another repo, it is very similar to `moment-mini` except that it includes both CJS and ESM builds (which actually both come from the Moment project itself), however `moment-mini` package only references the CJS build even they build both CJS & ESM nonetheless (more info can be found on the [moment-tiny](https://github.com/ghiscoding/moment-tiny) repo)

#### Note
The newest version of MomentJS introduced a breaking change via this [PR](https://github.com/moment/moment/pull/5592) even if they released it under patch, it really is a breaking change (more strict on 1-2 digits). I think what I could do to ease the pain would be to add a new `strictDateParse` (defaulting to `true`) so that user could avoid the breaking change. The reason not to use non-strict is mostly it's way too loose by default which is why I prefer strict by default but providing a date of `01/01/23` with `D/M/YY` should still be accepted and it was prior to that referenced commit. The strict mode is mostly used in the Date Sorting and Filtering

#### TODOs
- [ ] for the breaking change, another possibility is to provide not just 1 format but an array of formats that Moment will loop through internally for example `['M/D/YY', 'MM/DD'YY']` (however note that this could potentially impact perf unless 1st format is the correct one)
- [ ] similar to previous task, another way would be to add an option to disable strict mode (when filtering/sorting)


packages/common/src/filter-conditions/__tests__/dateEuroShortFilterCondition.spec.ts
```diff
@@ -75,35 +75,35 @@ describe('dateEuroShortFilterCondition method', () => {
...
+    const searchTerms = ['1/12/93..31/12/93'];
+    const options = { dataKey: '', operator: 'RangeInclusive', cellValue: '1/12/93', fieldType: FieldType.dateEuroShort, searchTerms } as FilterConditionOption;
     const output = executeFilterConditionTest(options, getFilterParsedDates(searchTerms, FieldType.dateEuroShort));
     expect(output).toBe(true);
   });
 
   it('should return False when input value equals the search terms min inclusive value and operator is set to "RangeExclusive"', () => {
-    const searchTerms = ['01/12/93..31/12/93'];
-    const options = { dataKey: '', operator: 'RangeExclusive', cellValue: '01/12/93', fieldType: FieldType.dateEuroShort, searchTerms } as FilterConditionOption;
+    const searchTerms = ['1/12/93..31/12/93'];
```
zewa666, line 99: is there a different behavior now with the date parsing?

zewa666:
hah thats a nice coincidence. we just installed moment yesterday for the date parsing of our dynamic column formatstrings and adding cjs to the new vite build felt so wrong 😅

I'm gonna give it a try as well

packages/common/src/filter-conditions/__tests__/dateEuroShortFilterCondition.spec.ts
ghiscoding: for unknown reason the newer version of moment doesn't accept zero padding anymore and in some ways it makes sense though, If we look at their [Format](https://momentjs.com/docs/#/displaying/format/) docs, it does show only 1 digit when using `D` or `M`. I could use `DD` or `MM` but I'm not sure if that would work and what I'm almost certain is that would probably fail in strict mode when zero padding is missing... I can give it another try tonight though

image

ok so I went through all their recent PRs and found this [commit](https://github.com/moment/moment/commit/6367947c3301d9bcb888c70c17bcc3a1486d6d45) introduced in v2.30.0 so it seems that it doesn't accept zero padding in strict mode only. I think I use strict mode when sorting and filtering

### Sorting 

https://github.com/ghiscoding/slickgrid-universal/blob/5ab671bee805f7b7d9b139e9bf7a14b31b5aea56/packages/common/src/sortComparers/dateUtilities.ts#L8-L16

### Filtering 
https://github.com/ghiscoding/slickgrid-universal/blob/5ab671bee805f7b7d9b139e9bf7a14b31b5aea56/packages/common/src/filter-conditions/dateFilterCondition.ts#L11-L20

ghiscoding:
> hah thats a nice coincidence. we just installed moment yesterday for the date parsing of our dynamic column formatstrings and adding cjs to the new vite build felt so wrong 😅
> 
> I'm gonna give it a try as well

Yeah so as I mentioned in the other discussion, as per this Moment [commit](https://github.com/moment/moment/commit/87994b745c20febf378ccd8f2dc190cd8d225020), they transpile to ESM but if you look at their `package.json`, it's not even linked anywhere (except JSPM which I don't use). I would guess that it's because they don't want to break their user base. So I simply cloned their repo, copied the ESM build (from `min/moment.min.js` location) and copied the root one which is CJS or UMD maybe and created a repo with that (I think moment-mini were doing the same anyway). I also provide a better `exports` in the package as well, so in theory it should work better with Vite (I haven't fully tested it though but I no longer see the CJS warning in Angular at least which was my goal)

packages/common/src/filter-conditions/__tests__/dateEuroShortFilterCondition.spec.ts
ghiscoding: I gave it a try and it is as I thought, when having a date like `01/01/92` and `M/D/YY` will not pass in strict mode. I think I should keep using strict mode in my lib nonetheless... So it really looks like MomentJS introduced a breaking change without even knowing and for that reason, I guess I have no choice but to postpone it for the next major release in 2 months and mention this new strictness in the migration

ghiscoding: @zewa666 alright, so what I found is that to go around the more strictness parsing problem with the new MomentJS version, what we can do is to provide multiple formats via an array. So what I've done is to add 2 formats for all the date formats that are considered as short. For example with `Field.dateUsShort`, I'm now passing `['MM/DD/YY', 'M/D/YY']` so that the user can still pass dates with/without zero padding. However there are 3 things to note
1. I can only use 1 output format and so I will always prefer the format with zero padding `MM/DD/YY`
2. parsing with multiple formats is slower than passing a single format and the array order matters, so zero padding will be quicker
3. I use strict mode when filtering & sorting, another idea would be to loosen the parsing and not use strict mode but non-strict mode is often too loose and I prefer strict mode

More info can be found in this Moment doc:
https://momentjscom.readthedocs.io/en/stable/moment/01-parsing/04-string-formats/

zewa666: awesome feature. just need something like that for the paste handling

ghiscoding: wow you're not sleeping, I thought you would reply in a day or two 😋 

> awesome feature. just need something like that for the paste handling

do you mean for another feature like the copy handler or did you mean something else? Should I go ahead with the merge then? what is your thought on the subject?

zewa666: kids just woke me up ;)

yeah we've just experienced during the POC that handling date pastes might be tricky as people usually have excel files and stuff whatever they found right into the grid. So having the combo of a couple of supported formats + a locale is a great option.

I think this is definitely good for merge.

going to have another week to focus on the grid in 2w so looking forward to incorporate the new releases and fixes 🎉

ghiscoding: yeah I wish that I could have switched to date-fns but it always seems that MomentJS is still the best and no other lib have all the bells and whistles of Moment. 

So I might have a beta by next week but it's still earlier than Angular 18 (which is scheduled for around May 20), however I will still make it the new minimum version requirement. I'm quite happy with the switch to pure CSS SVG, the Dark Mode is looking so so much better now :)

zewa666: dont look at me, still using moment and numeral in most projects bc it just works and the size overhead is ignorable in enterprise internal apps only. 😅

ghiscoding: I totally agree with all of that

ghiscoding:
@zewa666 have you had a chance to try `moment-tiny` with Vite? It looks like the min JS that I thought was an ESM build is not after all and I'm still getting the CJS in Angular and need to add it to allowed CJS. So I'm not sure if it's worth the move in that case.

zewa666:
just created a fresh new empty vite project, installed moment-tiny and it loaded it from the esm folder.

image

ghiscoding:
@zewa666  yeah ok but that's probably just because I did add an `exports` entry for ESM, however what I'm saying is that the build (file) in itself doesn't seem to be ESM at all because Angular still complains that it's CJS and for Jest I still have to use this ESM mock below for my unit tests to work (notice the use of `__esModule: true` which is to simulate an ESM package in a non-ESM package) 

https://github.com/ghiscoding/slickgrid-universal/blob/36494840122dfeb981b61e331f8a0daa2a4b7751/test/jest-global-mocks.ts#L41-L45

zewa666:
oh ok I see. I'll give it a try with angular and see how it goes

PR #1461, opened 2024-04-09
fix: wrong operator comparison

zewa666:
this fixes the wrong operator comparison, causing FieldTypes.date in OData to be always checked with the `eq` operator, regardless of the users definition


ghiscoding:
wow what a mistake, good catch. However I do see the exact same issue on the GraphQL Service, would you mind fixing that one too?

zewa666:
ups seems its also happening for graphql. Can't validated that with a real sample

ghiscoding:
Thanks again 😉

PR #1462, opened 2024-04-09
feat: notify onValidationError on paste if validation failed

zewa666:
this is both a new feature and a bug fix.
Now when pasting if the target cell has an editor with a validator, and that one returns invalid, the onValidationError event will be triggered.

But far more sever is that the paste on target cells with editors was currently broken.
See when going to https://ghiscoding.github.io/slickgrid-universal/#/example22, setting any row into edit mode, and pasting a number from Excel lets say into the column Completion. (check the devtools).

This regression somewhat was related to this PR https://github.com/ghiscoding/slickgrid-universal/pull/1429

### DIFF
--- a/examples/vite-demo-vanilla-bundle/src/examples/example22.ts
+++ b/examples/vite-demo-vanilla-bundle/src/examples/example22.ts
@@ -8,6 +8,8 @@ import {
 import { SlickCustomTooltip } from '@slickgrid-universal/custom-tooltip-plugin';
 import { Slicker, type SlickVanillaGridBundle } from '@slickgrid-universal/vanilla-bundle';
 import { ExampleGridOptions } from './example-grid-options';
+import { BindingEventService } from '@slickgrid-universal/binding';
+
 
 import './example22.scss';
 import type { TranslateService } from '../translate.service';
@@ -25,11 +27,13 @@ export default class Example22 {
   fetchResult = '';
   statusClass = 'is-success';
   statusStyle = 'display: none';
+  private _bindingEventService: BindingEventService;
 
   constructor() {
     this.translateService = (<any>window).TranslateService;
     this.selectedLanguage = this.translateService.getCurrentLanguage();
     this.selectedLanguageFile = `${this.selectedLanguage}.json`;
+    this._bindingEventService = new BindingEventService();
   }
 
   attached() {
@@ -44,10 +48,15 @@ export default class Example22 {
       { ...ExampleGridOptions, ...this.gridOptions },
       this.dataset
     );
+
+    this._bindingEventService.bind(document.querySelector(`.grid1`)!, 'onvalidationerror', (event) =>
+      alert((event as CustomEvent)?.detail.args.validationResults.msg)
+    );
   }
 
   dispose() {
     this.sgb?.dispose();
+    this._bindingEventService.unbindAll();
   }
 
   /* Define grid Options and Columns */
@@ -80,7 +89,7 @@ export default class Example22 {
         minWidth: 100,
         filterable: true,
         type: FieldType.number,
-        editor: { model: Editors.text },
+        editor: { model: Editors.text, validator: (val) => (val > 100 ? { msg: 'Max 100% allowed', valid: false} : { msg: '', valid: true}) },
       },
       {
         id: 'start',
--- a/examples/vite-demo-vanilla-bundle/src/examples/example22.ts
+++ b/examples/vite-demo-vanilla-bundle/src/examples/example22.ts
@@ -7,10 +7,9 @@ import {
 } from '@slickgrid-universal/common';
 import { SlickCustomTooltip } from '@slickgrid-universal/custom-tooltip-plugin';
 import { Slicker, type SlickVanillaGridBundle } from '@slickgrid-universal/vanilla-bundle';
-import { ExampleGridOptions } from './example-grid-options';
 import { BindingEventService } from '@slickgrid-universal/binding';
 
-
+import { ExampleGridOptions } from './example-grid-options';
 import './example22.scss';
 import type { TranslateService } from '../translate.service';
 
--- a/package.json
+++ b/package.json
@@ -97,5 +97,8 @@
   "funding": {
     "type": "ko_fi",
     "url": "https://ko-fi.com/ghiscoding"
+  },
+  "dependencies": {
+    "cypress-real-events": "^1.12.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "@types/node": "^20.12.7",
     "cross-env": "^7.0.3",
     "cypress": "^13.7.3",
+    "cypress-real-events": "^1.12.0",
     "dotenv": "^16.4.5",
     "eslint": "^9.0.0",
     "eslint-plugin-cypress": "^2.15.1",
@@ -97,8 +98,5 @@
   "funding": {
     "type": "ko_fi",
     "url": "https://ko-fi.com/ghiscoding"
-  },
-  "dependencies": {
-    "cypress-real-events": "^1.12.0"
   }
 }
--- a/packages/common/src/extensions/slickCellExternalCopyManager.ts
+++ b/packages/common/src/extensions/slickCellExternalCopyManager.ts
@@ -1,6 +1,6 @@
 import { createDomElement, getHtmlStringOutput, stripTags } from '@slickgrid-universal/utils';
 
-import type { Column, ExcelCopyBufferOption, ExternalCopyClipCommand, OnEventArgs } from '../interfaces/index';
+import type { Column, Editor, ExcelCopyBufferOption, ExternalCopyClipCommand, OnEventArgs } from '../interfaces/index';
 import { SlickEvent, SlickEventData, SlickEventHandler, type SlickGrid, SlickRange, type SlickDataView, Utils as SlickUtils } from '../core/index';
 
 // using external SlickGrid JS libraries
@@ -127,15 +127,15 @@ export class SlickCellExternalCopyManager {
 
     // if a custom getter is not defined, we call serializeValue of the editor to serialize
     if (columnDef) {
-      if (columnDef.editor) {
+      if (columnDef.editor?.model) {
         const tmpP = document.createElement('p');
-        const editor = new (columnDef as any).editor({
+        const editor = new (columnDef as any).editor.model({
           container: tmpP,  // a dummy container
           column: columnDef,
           event,
           position: { top: 0, left: 0 },  // a dummy position required by some editors
           grid: this._grid,
-        });
+        }) as Editor;
         editor.loadValue(item);
         retVal = editor.serializeValue();
         editor.destroy();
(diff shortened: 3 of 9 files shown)


packages/common/src/extensions/slickCellExternalCopyManager.ts
```diff
@@ -127,15 +127,15 @@ export class SlickCellExternalCopyManager {
 
     // if a custom getter is not defined, we call serializeValue of the editor to serialize
     if (columnDef) {
-      if (columnDef.editor) {
+      if (columnDef.editor?.model) {
```
ghiscoding, line 130: actually I think it would be best to call the one that the SlickGrid is aware of which is `editorClass` since SlickGrid doesn't have any knowledge whatsoever of `editor.model` (you won't find `editor.model` anywhere in SlickGrid core file, it is filled later by the wrapper which can be VanillaGridBundle, AngularSlickgrid, AureliaSlickgrid, ...). Note that there a few of them to replace

zewa666: thats interesting bc the examples all assign the model directly in the column definitions

hmm that also means though, while internalEditor is only deprecated but not gone we should take that first before editorClass?

ghiscoding:
> This regression somewhat was related to this PR #1429

yeah so we don't have much Cypress tests for copy+paste I guess and that is for sure not something I would think to test myself which is why it wasn't caught. There was this recent SlickGrid [PR](https://github.com/6pac/SlickGrid/pull/1011), which we should probably replicate, that added some Cypress tests for copy+paste in SlickGrid lib. We should probably do the same in here

zewa666:
cool realPress was the command I was missing. yep will add some Cypress tests now that I know how to replicate it 💯

not today as its getting quite late over here but will get things sorted this week

packages/common/src/extensions/slickCellExternalCopyManager.ts
ghiscoding: yeah I understand that it might be confusing, it might be just in my head but because I want to keep my `slickgrid.core.ts` file as close as possible to the original SlickGrid project. Some properties are not meant to be used within SlickGrid core file, for example any props related to export services and any plugin props and so on.. when I wasn't sure if I was going to still use `6pac/slickgrid` in here, I added Generics on SlickGrid that could be extended and so when I was using it in the wrapper, I could simply provide my custom `GridOption` and `Column` interfaces from the wrapper and SlickGrid was happy with it but it's just an extra layer that I would rather not really use too much (in fact I still have a lot of places where I have `grid: SlickGrid` instead of the full typed ref of `grid: SlickGrid<GridOption, Column>` because I find it too long and I don't have to do that anymore since I merged SlickGrid in here and typically the props I want to use are already in the GridOption/Column base interfaces)... so long story short, it's kind of "only me" knowing this little thing but I'm not sure I really want to go with the full Generics+extend interface way either because I think it would overcomplicate things.

https://github.com/ghiscoding/slickgrid-universal/blob/abe772b9ad9480688ef000dddfa86d2cdc6b7e52/packages/common/src/core/slickGrid.ts#L123-L124

ghiscoding: > hmm that also means though, while internalEditor is only deprecated but not gone we should take that first before editorClass?

the `internalEditor` will be removed completely in the next major and its goal was simply to hide the complexity of Editor creation to the user. I didn't want to introduce a breaking change so it's still in place and still work but will be removed completely in next major. SlickGrid always had a `editor` grid option (it never had the `filter` option though) and it was used like this `editor: Slick.longText`, but since I use a few external libraries (ms-select, autocomplete, flatpickr, ...), I needed a way to potentially provide extra options and I strongly felt that the developer would still prefer to use `editor` without necessarily caring about how it was used internally which is why at the end I came up with that `internalEditor` thing, it's something that was meant to be used only within the wrapper and hidden to the user...but in some cases the user want to have access to the editor instance and modify it (e.g. add some items to the ms-select `collection`) and in that case the user needed to reference the instance (not the class itself that SlickGrid editor had) and that was confusing that the editor was creating it with `editor` but later had to reference it with `internalEditor`... that's long story of how it all started, hope it helps understanding why I'm trying to make it simpler ;)

zewa666: ok got it. yeah I make use of the internalEditor a couple of times in my ForeignKey Filter/Editor so I understand what you're trying to simplify here.

zewa666:
alright I got my change now verified with Cypress as well. I'll add others as I touch the code base again. Pretty sure there will be more to come as I continue with my projects POC to Prod conversion

examples/vite-demo-vanilla-bundle/src/examples/example22.ts
```diff
@@ -8,6 +8,8 @@ import {
 import { SlickCustomTooltip } from '@slickgrid-universal/custom-tooltip-plugin';
 import { Slicker, type SlickVanillaGridBundle } from '@slickgrid-universal/vanilla-bundle';
 import { ExampleGridOptions } from './example-grid-options';
+import { BindingEventService } from '@slickgrid-universal/binding';
```
ghiscoding, line 11: I like separating external vs internal imports, so would you mind moving this import further to the top? Thanks :)

package.json
```diff
@@ -97,5 +97,8 @@
   "funding": {
     "type": "ko_fi",
     "url": "https://ko-fi.com/ghiscoding"
+  },
+  "dependencies": {
+    "cypress-real-events": "^1.12.0"
```
ghiscoding, line 102: I'm assuming this should go in Dev Dependencies instead

examples/vite-demo-vanilla-bundle/src/examples/example22.ts
zewa666: how about a linter rule for that? ;)

ghiscoding: oh does that actually exists? Dang ESLint is so powerful :P

ghiscoding:
I'm not sure what's happening with Example 12 but it keeps failing with this PR, does it have any reference to `internalEditor` or something?

examples/vite-demo-vanilla-bundle/src/examples/example22.ts
zewa666: I'll create an additional PR after this one as it currently would involve 319 problems ;)

zewa666:
meh ... example12 was the first target I wanted to look into using for E2E tests until I realized there was no copy&paste configured. I was fixing the validator condition as it didn't sound right according to the description and because I had errors when trying an inline edit. So I guess the tests build upon that bad assumption. I'll revert that again

examples/vite-demo-vanilla-bundle/src/examples/example22.ts
zewa666: its not only separation but also unifies blank lines between those and ordered imports. So with the extra PR you get a chance to tweak those configs as you prefer.

I'll set it up tomorrow as I need to get some rest today

ghiscoding:
> meh ... example12 was the first target I wanted to look into using for E2E tests until I realized there was no copy&paste configured. I was fixing the validator condition as it didn't sound right according to the description and because I had errors when trying an inline edit. So I guess the tests build upon that bad assumption. I'll revert that again.
> 
> Edit: yep thats it. But do you agree that it should be an OR instead of AND check there?

I'm not exactly sure what you mean, would you mind explaining more? I remember adding a validation of the "Task [number]" with a RegEx, is that at all related?

zewa666:
@ghiscoding all good, I'll create a new PR to clearly depict what I meant. As mentioned meanwhile I rolled back the change I've done here to example12 which makes the tests now pass.

So this PR should be done from my side

packages/common/src/extensions/slickCellExternalCopyManager.ts
```diff
@@ -155,15 +155,29 @@ export class SlickCellExternalCopyManager {
...
+      if (columnDef.editorClass) {
         const tmpDiv = document.createElement('div');
-        const editor = new (columnDef as any).editor({
+        const editor = new (columnDef as any).editorClass({
           container: tmpDiv, // a dummy container
           column: columnDef,
           position: { top: 0, left: 0 },  // a dummy position required by some editors
           grid: this._grid
-        });
+        }) as Editor;
```
ghiscoding, line 165: ahh I missed that earlier, I think we can remove the `any` on the column and make the `editorClass` parsed as `EditorConstructor` that was added not long ago. I'm guessing we might have to explicitly parse it though

```ts
const editor = new (columnDef.editorClass as EditorConstructor)({ })`
```
unless this other suggestion works, which would be better
```ts
const editor = new columnDef.editorClass({ }) as EditorConstructor`
```
Note that there's 2 of these to change

zewa666: not sure if thats gonna work as editorClass can be both the ctor and an editor according to the type

ghiscoding: yeah ok I'll take a look at that later, I wanted to check on Stackblitz but it looks like it's blocked by the VPN on work laptop, so anyway I can review that on a separate PR if it could be improved (or not)

ghiscoding:
@zewa666 your PR did not include the changes that I mentioned from SlickGrid [PR](https://github.com/6pac/SlickGrid/pull/1011/files#diff-96de36def70b26cc23034d06314bdc79220732b669b94fc5eba5e316a49de5d8), I'm assuming that we should probably add these changes right? That is to reapply the previous focus

zewa666:
nope I thought you mentioned to only replicate the copy&paste tests

ghiscoding:
ahh my bad, I meant that code change as well, but still I'm wondering if it's useful or not? You are definitely using it more than I do so... should we replicate it? What do you think?

zewa666:
need to check out that one today, especially with multi cell pastes

ghiscoding:
> need to check out that one today, especially with multi cell pastes

Have you had a chance to check this one? There's no rush per say, I would just like to know because I'm planning to push a new version this weekend. It could be postpone if you don't have time to check, no worries. Cheers

zewa666:
nope not yet but will do so tomorrow.

I've also seen the new date lib, which could be a real improvement. especially bc flatpickr lacks some really standard stuff like a clear button in the textfield (we had to hack one by ourselves).

with regards to the next branch, I think thats a good idea to have one early preview available (could also be npm published via @beta flags).

the couple smaller PRs now where the ones my teammates and me found during our first sprint. next one starts in two weeks but I feel like we've already covered a huge area. So yeah I'll check out and if it makes sense try to squeeze in the other PRs feature so that its rdy for your next release.

zewa666:
Hmmm ... I'm not 100% sure this is exactly a safe thing to do as focus stealing in general is regarded a bad practice. Imagine your validation lets another text box pop up where you need to write in a confirmation text. Now this PR would steal the focus and destroy your workflow. While the example is made up, generally we dont know what the user might do.

Now in this exact scenario I think I would be pro this feature but I'd introduce it during the breaking change release as focus redirects can lead to undesired side-effects. So should we open an issue with a link to said PR for the sake of not forgetting it once we're around the breaking change timing?

ghiscoding:
if you think it's more of a problem than a solution then we can just forget about the change. However if you think it's still worth it, then we can add it to the list of future changes or open an issue, up to you

PR #1474, opened 2024-04-19
feat(styling)!: convert SVG icons to pure CSS

ghiscoding:
- change how to colorize SVG icons to pure CSS approach by following AntFu [icons in pure CSS](https://antfu.me/posts/icons-in-pure-css) (creator of UnoCSS), this will help colorize icons for a much better Light/Dark Mode support

#### TODOs
- [x] all unit tests should pass
- [x] all E2E should pass
- [x] remove any reference to Font Awesome "fa" icons
- [x] update the [Styling SVG Icons](https://ghiscoding.gitbook.io/slickgrid-universal/styling/styling) doc about new ways to colorize SVG
   - [x] also all docs referencing `.mdi` requires an update
- [x] pagination SVG icons & change disable button icon to gray
- [x] row selection checkboxes
- [x] ColumnPicker/GridMenu lists
- [x] Draggable Grouping is not correct in Dark Mode
- [x] Grouping/DraggableGrouping/TreeData (collapse/expand) icons
- [x] Sort icons
- [x] RowDetail icons
- [x] Autocomplete loading icon
   - this is an exception, we cannot make it pure CSS because we use `::after` pseudo to display loading spinner within the input and it's only possible via `content` and we'll keep a light gray because we can only set the color once
- [x] remove all unused ms-select SASS variables
- [x] ~remove all references to `.mdi` classes and replace them by `.sgi`~
  - [x] decided to rollback and keep `.mdi` to make the version migration easier and smoother
- [x] remove all `recolor` SASS code and also color using CSS filter
- [x] we currently have `$color-` and `$text-color-` but with new pure CSS, we can probably merge into one.
  - [x] keep only `text-color-xx` and delete previous SASS `color-xx` since all SVG can now use same color as any text
  - [x] possibly make some dark mode alternative as well (e.g. custom `text-color-primary` and `text-color-secondary` in `style.scss`)
- [x] remove SASS variable `$slick-icon-font-family` since we no longer use any Fonts internally
- [x] remove old jQueryUI `.ui-state` and others

### Build Size Comparison
I wasn't expecting much of a difference but I'm pleasantly surprised to see this minimal but nice drop in build size, that is with an improved use of pure CSS and a much better UI/UX :)

| [SF Bundle Zip File](https://github.com/ghiscoding/slickgrid-universal/tree/master/packages/vanilla-force-bundle/dist-grid-bundle-zip) | 582Kb (596,961) | 575Kb (589,262) | 7Kb | 1.29% smaller |


ghiscoding:
@zewa666 out of curiosity, which slickgrid theme and icons are you using (the default Slickgrid-Bootstrap maybe)? Because I'm doing a major revamp, as seen in TODO list above, and I'm renaming all `.mdi` CSS classes to `.sgi` because this fit better with the project itself (SlickGrid Icons) and mostly because the way to create the SVG is entirely different (now I'm using AntFu's approach for pure CSS as opposed to prior code using `::before` pseudo with `content`). The native approach allows the user to colorize any SVG as opposed to before it was a fixed color... anyway long story short, I'd like to know what users use and if the rename is a huge thing (I guess it's ok to just do search and replace in the project) and also I might get rid of some SASS files that might be barely used (I created `bare` and `lite` versions for Material & Salesforce Themes but I'd be surprised if anyone uses them, the description is shown in comments below). I'd be happy to receive feedback before I merge this large PR.

https://github.com/ghiscoding/slickgrid-universal/blob/094d7602d7170b2f395985ce5635041bb2b803d2/packages/common/package.json#L50-L54

https://github.com/ghiscoding/slickgrid-universal/blob/094d7602d7170b2f395985ce5635041bb2b803d2/packages/common/src/styles/slickgrid-theme-material.bare.scss#L7-L9

https://github.com/ghiscoding/slickgrid-universal/blob/094d7602d7170b2f395985ce5635041bb2b803d2/packages/common/src/styles/slickgrid-theme-material.lite.scss#L7-L11

zewa666:
I'm using both the full material and also bootstrap theme. if not necessary I'd personally advocate for keeping them around as is. I'd really like to also push the grid for one of our SFDC use cases, so having that specific theme available would be a plus

ghiscoding:
> I'm using both the full material and also bootstrap theme. if its not too much of a hassle I'd personally advocate for keeping them around as is. I'd really like to also push the grid for one of our SFDC use cases, so having that specific theme available would be a plus

hmm yeah I was undecided on the `.sgi` change, I prefer `.sgi` to represent SlickGrid Icons (which I have actually done in SlickGrid core lib) but I did think about potential users might be holding their migration because of the size of search & replace. The new icons are not aligned the same way either, in previous version I had to add some extra align bottom because I was using `::before { content: 'svg...' }` but now that it's pure CSS, it follows the same alignment and colors as any other text. I was also wondering if anyone using the real Material Design Icon font, it might conflict in that case. 

I have already changed everything to `.sgi` in this PR, but the main reason was mostly to make sure I wasn't referencing the old SVGs and that I tested everything (UI), but now that I'm done then I guess I could go back to `.mdi`. I still have a preference for `.sgi` but I do understand users might be holding their migration because of the massive search & replace.

ghiscoding:
I decided to rollback the `.mdi` to `.sgi` change to make the migration smoother and easier for the developers and avoid having users postponing their upgrades because of it. @zewa666 I'm still curious if you could reply to my previous paragraph though, thanks

zewa666:
I'll have to check on Tuesday when back in office about the lite/bare use

zewa666:
we're using the full material theme:

`@import "@slickgrid-universal/common/dist/styles/sass/slickgrid-theme-material.scss"`

ghiscoding:
good to know, I wasn't even sure if anyone were using the Material Theme. cool

PR #1493, opened 2024-04-29
feat(styling)!: delete "bare" Themes but keep "lite" & add to Bootstrap

ghiscoding:
Here's quick details about file size comparison with/without "lite" for each theme. I think Material is slightly larger because it also includes the the Roboto font family

image

Also tested in Angular-Slickgrid, all is good after migrating all `.fa` to `.mdi` icons. I also tested the "lite" edition and it works correctly too (sort, grouping, ... icons are showing but not the menu ones)


ghiscoding:
After testing the change in Angular-Slickgrid, I pushed a couple of small commits and I don't see any problems with this new implementation, so I'm gonna go ahead and merge.

PR #1507, opened 2024-05-05
feat!: migrate from Moment to Tempo

ghiscoding:
- after more research, I found [Tempo](https://tempo.formkit.com) and I can finally migrate away from MomentJS. Tempo is simple, is ESM, is small and has all necessary functions to replace Moment
- by removing Moment (finally), we also get a much smaller build (not surprisingly since it's ESM)
- Few notes about Tempo
  1. it uses the exact same [format tokens](https://tempo.formkit.com/#format-tokens) as Moment (awesome)
  2. it doesn't support multiple formats like Moment did for parsing with/without zero padding but that doesn't matter since the non-padding token works for both usage (e.g. 'M/D/YY' will work for both '4/7/24' and '04/07/24'). With that in mind, I will use 'M/D/YY' to parse but I will use 'MM/DD/YY' to format date output

#### TODOs
- [x] investigate why locally `a` token is showing locally as `a.m.` but in CI it's showing as `am` (I would prefer same result on both side as `am`). I might need to open an issue on Tempo. I found out that it's caused by `Intl.DateTimeFormat` displaying `a.m.` when using "en-CA" but `am` when using "en-US" and so I'll use "en-US" everywhere to avoid such inconsistenticies
- [x] all unit tests should pass (except `a.m.` format, they all do)
- [x] all Cypress E2E tests should pass
- [x] requires Migration Guide & other docs update

### Build Size Comparison
| [SF Bundle Zip File](https://github.com/ghiscoding/slickgrid-universal/tree/master/packages/vanilla-force-bundle/dist-grid-bundle-zip) | 549Kb (562,835)  | 535Kb (548,088) | -14Kb |  2.62% smaller |
| [GitHub Vite Demo Website](https://ghiscoding.github.io/slickgrid-universal/) | 5.16Mb (5,412,541)  | 5.11Mb (5,364,919) | -0.05Mb  | 0.88% smaller |


ghiscoding:
cc @zewa666

PR #1528, opened 2024-05-14
feat(filters): add a `filterPredicate` option for user customization

ghiscoding:
- adding a `filterPredicate` would help user who want to have a custom filter predicate callback that cannot be handled by built-in filters
- also added a patch filter version, that was only for testing purpose (will probably remove before merging) and to provide a way to answer this SO question: https://stackoverflow.com/questions/78471412/angular-slickgrid-filter

### TODOS
- [x] unit tests
- [x] E2E tests
- [x] add documentation


ghiscoding:
@zewa666 I wouldn't mind a PR review, it brings a `filterPredicate` to allow user to provide his own filter predicate which answers this SO [Angular-Slickgrid Filter](https://stackoverflow.com/questions/78471412/angular-slickgrid-filter). You can ignore the `patchDataViewFilter()`, it was only meant to provide a way to answer the person's request and I will remove it before merging

zewa666:
so the predicate would essentially by-pass the built-in filter service filtering function? what happens to the operators though?

I'm actually very interested in the exact same case searching using an asterisk character but I thought of that more along the lines of being a new operator and also working for the OData backend with either eq operators including the wildcard or a custom odata function.

ghiscoding:
> so the predicate would essentially by-pass the built-in filter service filtering function? what happens to the operators though?

hmm yeah I forgot about the operator and I guess it would be skipped too, but if I take a look at the `searchFilterArgs`, the `operator` is included, so in theory the user would also customize that if he wants.

```ts
filter: {
  filterPredicate: (dataContext, searchFilterArgs) => {
     console.log(searchFilterArgs);
  }
}
```
is giving this console, it is currently set as `Contains` because that is the default operator, but if the `operator` prop is changed it should show up as well. The operator is analyzed before the call to the filter predicate, the reason it's called before is because we want to analyze it only once and not doing this work for every data row which would kill perf

image

if I type "*Ta" then the operator becomes "EndsWith" which is the typical operator detection. If the user wants to use something else as operator he could just parse the search value for whatever token he wants to use. 

image

I also added a `OperatorType.Custom` in previous PR #1526 which would probably be more useful for customizing backend services, by itself the `OperatorType.Custom`  does nothing, I just added to make it a valid operator in case the user wants to use his own customized backend

zewa666:
oh ok thats great, didnt notice that one. guess thats also on the next branch right now? while at it can we also define what the visible text will be in the operator dropdown? e.g for the like I'd like to have "a*b" as that is neither starts nor ends with but also not contains.

back to the original question though, ideally the dev wouldnt pick a filter type that allows to select an operator if he provides a custom filterPredicate in order to reduce UI complexity for the users.

filterPredicate sounds good as well as predicate alone might not be intuitive enough as you said

ghiscoding:
> hmm also note your implementation is not exactly correct as while %Ta% might be a contains what about %Ta%1? So the wildcard is treated like a regex dot-character zero or multiple times.

bahh I don't care that much, I did improve it but forgot to push the commit (which I just did). I mean, I spent like 2 hours trying to get it working with regex and just gave up and using `.split()` is a quick and dirty way to do it (I would prefer regex ideally). After updating the code, I added an animated gif in the [SO](https://stackoverflow.com/a/78479236/1212166) which shows what the user was looking for (more or less).

ghiscoding:
hahah taking another look at my own docs, I forgot that I actually already done the change for a completely custom compound list.. dohhhh 😆 

See [Compound Operator List (custom list)](https://ghiscoding.gitbook.io/slickgrid-universal/column-functionalities/filters/compound-filters#compound-operator-alternate-texts) which is just above the section "Compound Operator Alternate Texts".... isn't that great, when you just realized there's nothing more to do haha

The only thing left I guess would be to provide custom operators, I still don't know how we could do that though. Because, currently I use `OperatorType` (which is an enum) and/or `OperatorString` (which is a `type`) and after googling a little, I couldn't really find a way to extend both in TS
- [How to extend a Type in TypeScript](https://bobbyhadz.com/blog/typescript-extend-type)
- [Extending Enum in typescript](https://stackoverflow.com/questions/52857698/extending-enum-in-typescript)

I don't see a way to address this without a breaking change, which I'm not looking to do at all. Not just because I just released a breaking change version but mostly because of usability, it's easy as it is today to write either `type: OperatorType.greaterThan` or `type: '>='` (both are currently accepted). The only other way would be to have something like `type: OperatorType | OperatorString | any` (or `unknown` maybe) but that would kill the intellisense which is not ideal either, so... no idea on how to fix this

zewa666:
not actively using Aurelia any longer sadly due to timeconstraints and lack of projects.

thx for all the info, I guess with the one new Operator.Custom I should be all set for my needs

ngcvm:
Hi @ghiscoding, I'm in the old version of `slickgrid-universal` and cannot update the version easily. How can I achieve the same functionality of `filterPredicate` in the old versions?

ghiscoding:
@ngcvm  The Stack Overflow answer does show you how to do it with an old version and yes it's achievable. I spent a lot of effort to not just add this new option but also to answer that SO, so you should give it a try... and sorry but I just don't have time to support old versions, I'm not being paid for any of this either (though I did get a large Ko-Fi donation from that user), so... good luck

ngcvm:
> @ngcvm  The Stack Overflow answer does show you how to do it with an old version and yes it's achievable. I spent a lot of effort to not just add this new option but also to answer that SO, so you should give it a try... and sorry but I just don't have time to support old versions, I'm not being paid for any of this either (though I did get a large Ko-Fi donation from that user), so... good luck

@ghiscoding Thank you for your answer and amazing works.

PR #1529, opened 2024-05-14
chore(demo): improve SQL LIKE filter predicate example

ghiscoding:
- improves SQL LIKE filter predicate demo by using Regex, I'm not exactly sure why `.match()` doesn't return all matches while `.matchAll()` does, but anyway that is what works so let's go with that even it's more bulky (it's just for demo purposes anyway).

### DIFF
--- a/examples/vite-demo-vanilla-bundle/src/examples/example14.ts
+++ b/examples/vite-demo-vanilla-bundle/src/examples/example14.ts
@@ -158,19 +158,23 @@ export default class Example14 {
             const searchVals = (searchFilterArgs.parsedSearchTerms || []) as SearchTerm[];
             if (searchVals?.length) {
               const columnId = searchFilterArgs.columnId;
-              const searchVal = searchVals[0] as string;
-              const likeMatches = searchVal.split('%');
-              if (likeMatches.length > 3) {
-                // for matches like "%Ta%10%" will return text that starts with "Ta" and ends with "10" (e.g. "Task 10", "Task 110", "Task 210")
-                const [_, start, end] = likeMatches;
+              const results = (searchVals[0] as string).matchAll(/%(.*)%(.*)%|%(.*)%(.*)|(.*)/gi);
+              const arrayOfMatches = Array.from(results);
+              const matches = arrayOfMatches.length ? arrayOfMatches[0] : [];
+              const [_, start, end, firstContain, containLeftover, others] = matches;
+
+              if (start && end) {
+                // example: "%Ti%001%"
                 return dataContext[columnId].startsWith(start) && dataContext[columnId].endsWith(end);
-              } else if (likeMatches.length > 2) {
-                // for matches like "%Ta%10" will return text that starts with "Ta" and contains "10" (e.g. "Task 10", "Task 100", "Task 101")
-                const [_, start, contain] = likeMatches;
-                return dataContext[columnId].startsWith(start) && dataContext[columnId].includes(contain);
+              } else if (firstContain && containLeftover) {
+                // example: "%Ti%001"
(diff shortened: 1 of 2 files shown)


ghiscoding:
@zewa666 so I got it working with regex by using `.matchAll()` instead of `.match()`, not exactly sure why but hey it works so... I also added an extra condition, so I currently cover these 4 LIKE use cases
- `%Ta%001%`: starts with "Ta" and ends with "001"
- `%Ta%001`: starts with "Ta" and contains "001" (anywhere)
- `%Ta%`: contains "Ta" (anywhere)
- `%Ta` or `Ta` or anything else really

I'm not entirely sure if it's the correct LIKE use case but I'll go with that

zewa666:
make sure to escape potential special regex characters present in the searchString like dots or braces.

here is the same URL I gave in the last PR which has some good hints https://codereview.stackexchange.com/questions/36861/convert-sql-like-to-regex/36864#36864

for my case handling those options is merely passing them to the LIKE operator in SQL after sanitization. So it kinda should just work.

ghiscoding:
hmm I'm not exactly sure what you mean by sending this SO, are you saying that I should remove my regex and use this dynamic function to create a regex dynamically? It might be helpful if you provide the whole function if you don't mind then... and why would the search value need to be escaped? I mean if I use `(.*)`, that would return anything between, unless you mean with this new dynamic regex function, then I guess yes

ghiscoding:
Alright, so after more Regex tweaks, I think I'm much more aligned with the SQL LIKE now (at least more than what the original SlickGrid question required) and this is what I came up with by testing it in regex101 and then JSfiddle, now I just have to update the demo code and adjust tests. Anyway, I'm adding this here for reference and in case I ever need to come back to it in the future (even if it's for demo purposes 😆)

```ts
const collection = ['Title 100.001', 'Title 100.002', 'Title 11.001', 'Title 1001.002'];
const searchVal1 = '%001';    // ends with
const searchVal2 = '%Ti%001'; // contains A + ends with B
const searchVal3 = '%Ti%';    // contains A
const searchVal4 = 'Ti%001';  // combo starts with A + ends with B
const searchVal5 = 'Ti%';     // starts with A
const searchVal6 = '%%';      // other
const searchVal7 = '100';     // other

let searchVal = searchVal1;
const result = collection.filter(s => {
  const results = searchVal.matchAll(/^%([^%\r\n]+)[^%\r\n]*$|%(.+)%(.*)|(.+)%(.+)|([^%\r\n]+)%$/gi);
  const arrayOfMatches = Array.from(results);
  let matches = arrayOfMatches.length ? arrayOfMatches[0] : [];
  const [_, endW, contain, containEndW, comboSW, comboEW, startW] = matches;
  //console.log(matches)
  if (endW) {
  	console.log('1')
    // example: "%001" ends with A
    return s.endsWith(endW);
  } else if (contain && containEndW) {
  	console.log('2')
    // example: "%Ti%001", contains A + ends with B
    return s.includes(contain) && s.endsWith(containEndW);
  } else if (contain && !containEndW) {
  	console.log('3')
    // example: "%Ti%", contains A anywhere
    return s.includes(contain);
  } else if (comboSW && comboEW) {
  	console.log('4')
    // example: "Ti%001", combo starts with A + ends with B
    return s.startsWith(comboSW) && s.endsWith(contain);
  } else if (startW) {
  	console.log('5')
    // example: "Ti%", starts with A
    return s.startsWith(startW);
  }
  // anything else
  console.log('6')
  return s.includes(searchVal);
});
console.log(result)
```

I think the combo starts with + ends with is what we should eventually add as a new operator which is what you asked @zewa666

zewa666:
exactly, * is the search token and you cant explicitly search for it as value. thats fine. good work

PR #1530, opened 2024-05-16
feat(filters): add StartsWith/EndsWith (`a*z`) filter combo

ghiscoding:
### TODOs
- [x] add unit tests
- [x] add E2E tests
- [x] requires more manual tests in UI
- [x] not sure what happens when using something like `Ta*10*`, that probably won't work, should we support that?
   - returns same result as `Ta*10` but that's ok, I won't support more complex scenarios like this one. 
   - we'll support only 1 `*` symbol

brave_40cP5F4RL9


ghiscoding:
@zewa666 first PR to add a new `OperatorType.StartsWithEndsWith` (alias to `a*z` string), I've done some quick tests and it seems to work as intended for a local JSON dataset. I assume that you could look at the OData/GraphQL in a separate PR 😉 It's looking good so far

ghiscoding:
I'm not sure what would happen if we type `Ta*10*`, that probably won't work, however should we really support that?

zewa666:
to be honest I'm not sure. on one hand its overkill for a simple client side filter. on the other hand since SQL supports this natively with LIKE it could cause a
mismatch between client/server filtering. but that said its also a very narrow usecase for a db driven backend. who says a graphql service needs to grab data from a DB.

in all honesty, I think you went waaaay over the stretch goal to answer the original SO question, so I'd keep it at the state you have and call it a day. It can always be improved later on.

ghiscoding:
Yeah I also think it's probably overkill to go that deep into a simple input text filter and yeah I know I went over the top with the SO question but I was kind of curious at the same time and thought we should support at least `a*z`, so that was a good exercise (even though it was much longer than I expected lol).

> BTW, do you have a rough estimate on when you're planing to release the new major both for universal and angular, absolutely no pressure here just curious? I wonder whether I should wait for that or do an interim update for my app. heck you're so damn fast with new releases I'm already 3-4 behind 🤣

I'm waiting for Angular 18 which is scheduled for next week, I hope that they won't have any delay though (which happened in the past). Since I already release 5.0 here and the new stuff will be under 5.1, I'm thinking of pushing 2 successive release of Angular-Slickgrid to be on the same minor (unless I figure out how to force `release-it` directly to v8.1, I now use GitHub Action to release so that it comes with `--provenance` enabled but the versions are auto-detected via conventional changelog, so anyway I could look into a prompt release if possible)

image

ghiscoding:
also the reason I'm waiting for Angular 18 is not just because it's nearby but mainly because I don't expect to ship that many (if any) major versions in the future (since I consider to be pretty much feature ready and done after 7 years), so I think it's better to stick with latest Angular if possible (that is why I decided to wait few weeks for Angular latest version)

ghiscoding:
indeed it's a long journey and I learned quite a lot, also happy to hear that these backend services were not just useful for also bringing auto-magic in some ways hehe. 🐰

PR #1536, opened 2024-05-21
feat(OData): add `filterQueryOverride` to OData Service

zewa666:
Hey @jr01 and @ghiscoding,
 
In order to continue the discussion from https://github.com/ghiscoding/slickgrid-universal/pull/1532 I'd like to start with this PR as a discussion basis.

I'm not sure I fully understood the original requirements but here's what I came up with so far:

* Accept an optional `filterPredicate` as ODataOption
* If one is defined it gets precedence over other operator type checks
* If it returns undefined, the others are re-visited
* By providing all necessary information the predicate can now target based on
  * Individual columns
  * Different Operators (both user selected and inferred)

So my goal was to stay as generic as possible in order to support as much use cases as there might be.

Let me know what you guys think of this. Happy change according to our discussions outcome


examples/vite-demo-vanilla-bundle/src/examples/example09.ts
```diff
@@ -104,6 +110,11 @@ export default class Example09 {
         hideInFilterHeaderRow: false,
         hideInColumnTitleRow: true
       },
+      // TODO: doesn't work in combo with compoundOperatorList
+      compoundOperatorAltTexts: {
+        // where '=' is any of the `OperatorString` type shown above
+        text: { 'Custom': { operatorAlt: '⊂', descAlt: 'SQL Like' } },
```
ghiscoding, line 116: I would rather use `operatorAlt: '%'`, can you elaborate on the TODO above? Shouldn't the `compoundOperatorList` be enough without the need to also use `compoundOperatorAltTexts`? I'm saying this because I don't fully recall, been too long, if both are needed, all I remember is that `compoundOperatorAltTexts` is strictly for text alternate

zewa666: with compoundOperatorList you cant define your own altText but need to pick from a union string/enum. if you do provide a custom operator list though, the path that applies compoundOperatorAltTexts is skipped

packages/odata/src/services/grid-odata.service.ts
```diff
@@ -426,8 +426,20 @@ export class GridOdataService implements BackendService {
             fieldName = titleCase(getHtmlStringOutput(fieldName || ''));
           }
 
-          // StartsWith + EndsWith combo
-          if (operator === OperatorType.startsWithEndsWith && Array.isArray(searchTerms) && searchTerms.length === 2) {
+          let predicateResult: string | undefined = undefined;
+          if (typeof this._odataService?.options?.filterPredicate === 'function') {
```
ghiscoding, line 430: I took a quick look at `this._odataService` and since it's instantiated in the constructor, I believe we can safely assume that it's always filled and we should remove optional chaining from it, not just yours but everywhere else in this file

```diff
- this._odataService?.options
+ this._odataService.options
```

zewa666: PRs with you feel like talking to a Linter ... no excuses even for DraftPRs 🤣

packages/odata/src/services/grid-odata.service.ts
```diff
@@ -426,8 +426,20 @@ export class GridOdataService implements BackendService {
...
+              fieldName: getHtmlStringOutput(fieldName),
+              columnDef,
+              operator,
+              columnFilterOperator: columnFilter.operator,
+              searchValue
+            });
+          }
+
+          if (predicateResult !== undefined) {
+            searchBy = predicateResult;
```
ghiscoding, line 441: I'm not sure if it's necessary or not, but when I introduced `filterPredicate` for local JSON dataset, I was returning the filter predicate result all the time, however that had an indirect regression that I forgot to test at first and fixed it later. The regression was that when predicate returns `true`, it assumed it was true for all other column filters (which is obviously wrong), so I fixed the regression by only returning when the predicate result is `false`.. but I'm not sure if it's relevant in here though, I'm just adding this a step to verify

https://github.com/ghiscoding/slickgrid-universal/blob/a2799f2234eece619ec949165b76c23285c91698/packages/common/src/services/filter.service.ts#L367-L373

packages/odata/src/interfaces/odataOption.interface.ts
```diff
@@ -37,6 +37,9 @@ export interface OdataOption extends BackendServiceOption {
   /** Filter queue */
   filterQueue?: any[];
 
+  /** An optional predicate function to overide the built-in filter construction  */
+  filterPredicate?: (args: OdataFilterPredicateArgs) => string | undefined;
```
ghiscoding, line 41: typically a filter predicate returns a `boolean` value (to filter or not certain values), I wonder if we should use a different callback name instead? Another reason to rename it would be because the one I implemented earlier to local JSON dataset does return a `boolean`, so if this one is returning something else then it's not an exact match in implementation to the user, which might bring confusions. I'm not sure what other name to use though (other similar subject are potentially `reduce` so `filterReducer` or anything else that is not confusing the user to think it's not returning a `boolean`

cc @jr01 on this one too

examples/vite-demo-vanilla-bundle/src/examples/example09.ts
ghiscoding: hmm ok will have to revisit, we might need a more generic approach to make them both happy

packages/odata/src/interfaces/odataOption.interface.ts
zewa666: absolutely agree, kept filterPredicate for the sake of discussion but something like `filterOverride` or `filterBypass`

packages/odata/src/interfaces/odataOption.interface.ts
```diff
@@ -55,3 +58,16 @@ export interface OdataOption extends BackendServiceOption {
   /** A callback which will extract and return the dataset from the data queried. Defaults to 'd.results' for v2, 'results' for v3 and 'value' for v4. */
   datasetExtractor?: (response: any) => number;
 }
+
+export interface OdataFilterPredicateArgs {
```
zewa666, line 62: should we include any other arguments here? like the gridInstance?

ghiscoding: the grid instance seems like a good idea, I'm wondering about the `searchValue`, shouldn't this be `searchTerms`? For example a ms-select can have multiple values selected and that's why I've set all filter to use `searchTerms` (plural). But I'm not sure if it was really meant to be typed value in this case here? Another prop might be the `columnFilter` though we can get it from `columnDef.filter`

packages/odata/src/interfaces/odataOption.interface.ts
ghiscoding: I like `filterOverride` or `filterQueryOverride` if the result is towards the query string composition

packages/odata/src/interfaces/odataOption.interface.ts
zewa666: searchTerms is already the parsed result of your regex. that might have implications on how the searchValue gets split. e.g `he*lo` becomes `['he', 'lo']` but it swallows the star that I'd need for my LIKE search as an example if star is the wildcard.

ghiscoding: ahh ok yeah, if it's the same as the value used by the regex then yeah it's a string, I just wrote it without looking at the code and just making sure it would work with all type of filters :)

packages/odata/src/interfaces/odataOption.interface.ts
jr01: I very much like that I can now specify a custom operator on a column and have a fully flexible filter implementation in the (odata) backendServiceApi. This would perfectly solve my use case.

In my code base I could have different kinds of custom operators and I can use a `params.whatEver` to distinguish between these.

Naming is hard. `filterOverride` is OK I think. `customFilter` or `customOperatorFilter` perhaps lean more towards `OperatorType.custom`.

packages/odata/src/interfaces/odataOption.interface.ts
jr01: For my use case I would only need columnDef. I currently also use searchTerms, but I can (better) grab the multi-select values from columnDef.filter.collection.value's.

packages/odata/src/interfaces/odataOption.interface.ts
ghiscoding: @jr01 just so you know, since I'm not sure if you followed all our threads, I added a new `OperatorType.custom` (or the string 'Custom' as `OperatorString`) in PR #1526. The reason I added it was to provide a valid operator to the user, it's not used anywhere but it's available so that you can use it and it will compile without any errors. This is yet to be released and will require the new Angular-Slickgrid major 8.x and I'm waiting for Angular 18 to be released any day now. 

> Naming is hard. `filterOverride` is OK I think. `customFilter` or `customOperatorFilter` perhaps lean more towards `OperatorType.custom`.

hmm I think you're right that it will always be used with the `Operator.Custom` but I'm wondering if it always be the case? What if the user want to modify the logic of an existing operator, then that won't be a custom operator anymore (though I guess that would be super rare since I would prefer to receive PR to fix it in the code directly but who knows maybe OData v5 comes and changes things again).

zewa666: well actually I do have a use cass for overloading existing operators. if an eq search is performed on a datetime field I'd typically like to ignore the time part but only filter on the date part. similarly a calendarweek eq search should be transformed to between of cw start and cw end

ghiscoding:
hey guys, so Angular 18 just came out (an hour ago), so I might release Angular-Slickgrid 8.0 tonight and release a `minor` in the next few days when we're satisfied with this PR (I'll also have to see if we can duplicate this to GraphQL Service as well).

zewa666:
sounds great, I'll try to get the tests and docs in place.

examples/vite-demo-vanilla-bundle/src/examples/example09.ts
ghiscoding: so I took a look at the code and I now understand what you meant, but I'm not sure if it's worth doing anything about it. I mean that would require some refactoring and I assume that most user would simply use translated values. What is your taught on this? If it should be changed then it would be better as a separate PR to separate the concerns

zewa666: its somewhat unexpected to behave this way. but yep this is certainly for another PR as it doesnt have anything to do with this topic

ghiscoding: ok I've opened PR #1541 which should fix this issue

ghiscoding: I assumed that the PR was ok so I merged it, you can merge master branch to get it. However I would prefer to use something like `{ operatorAlt: '%%', descAlt: 'SQL Like' }` as alternate text, it seems more obvious to me :)

Apart from that, I think this comment is "resolved"

packages/odata/src/interfaces/odataOption.interface.ts
ghiscoding: > In my code base I could have different kinds of custom operators and I can use a `params.whatEver` to distinguish between these.

@jr01 I don't think that you'll be able to have multiple Custom Operators, if I recall correctly then the operator type must be unique (which is `Custom` in this case), so that won't work (the select dropdown also needs to have unique values for it work properly). Do you need that many custom operators? I'm not really sure that adding a bunch of `Custom1`, `Custom2` is good either, so...  also remember that `OperatorType` is a fixed list and it won't compile if you deviate

zewa666: I guess he meant different custom operators across multiple columns. thats at least how I read it :)

jr01: Yes I meant different custom operators across multiple columns :-)

zewa666:
I got the example properly updated, docs and tests added.

I'm not really sure what to do on the GraphQL side of things so I'd leave that up to you.

ghiscoding:
Yeah no worries, I'll take a look later to see if it's doable or not to update GraphQL as well. I spotted a couple of typos in your docs when I read it quickly last, I'll take another look at it and add comments. Thanks a lot for all of that, always nice to get contributions for your side :)

docs/backend-services/OData.md
```diff
@@ -221,6 +221,30 @@ Navigations within navigations are also supported. For example `columns: [{ id:
 The dataset from the backend is automatically extracted and navigation fields are flattened so the grid can display them and sort/filter just work. The exact property that is used as the dataset depends on the oData version: `d.results` for v2, `results` for v3 and `value` for v4. If needed a custom extractor function can be set through `oDataOptions.datasetExtractor`.
 For example if the backend responds with `{ value: [{ id: 1, nav1: { field1: 'x' }, { nav2: { field2: 'y' } } ] }` this will be flattened to `{ value: [{ id: 1, 'nav1/field1': 'x', 'nav2/field2': 'y' } ] }`.
 
+### Override the filter query
+
+Column filters may have a `Custom` operator, that acts as a placeholder for you to define your own logic. To do so, the easiest way is to provide the `filterQueryOverride` callback int the OdataOptions. This method will be called with `OdataFilterQueryOverrideArgs` to let you decide dynamically on how the filter should be assembled.
```
ghiscoding, line 226: small typo in "callback int the OdataOptions" on the "int"

docs/backend-services/OData.md
```diff
@@ -221,6 +221,30 @@ Navigations within navigations are also supported. For example `columns: [{ id:
...
+
+E.g. you could listen for a specific column and the active OperatorType.custom in order to switch the filter to a matchesPattern sql like search:
```
ghiscoding, line 228: I typically like to see SQL as uppercase so "SQL LIKE search" seems better to me :)

docs/backend-services/OData.md
```diff
@@ -221,6 +221,30 @@ Navigations within navigations are also supported. For example `columns: [{ id:
...
+
+```ts
+backendServiceApi: {
+  options: {
+    filterQueryOverride: (args) => {
+      const { fieldName, columnDef, columnFilterOperator, searchValue } = args;
+      if (columnFilterOperator === OperatorType.custom && columnDef?.id === 'name') {
+        let matchesSearch = (searchValue as string).replace(/\*/g, '.*');
+        matchesSearch = matchesSearch.slice(0, 1) + '%5E' + matchesSearch.slice(1);
```
ghiscoding, line 237: what is the escape of `%5E`? Maybe add a comment mentioning what and why we're escaping?

examples/vite-demo-vanilla-bundle/src/examples/example09.ts
```diff
@@ -131,6 +143,16 @@ export default class Example09 {
           enableCount: this.isCountEnabled, // add the count in the OData query, which will return a property named "__count" (v2) or "@odata.count" (v4)
           enableSelect: this.isSelectEnabled,
           enableExpand: this.isExpandEnabled,
+          filterQueryOverride: (args) => {
+            const { fieldName, columnDef, columnFilterOperator, searchValue } = args;
+            if (columnFilterOperator === OperatorType.custom && columnDef?.id === 'name') {
+              let matchesSearch = (searchValue as string).replace(/\*/g, '.*');
+              matchesSearch = matchesSearch.slice(0, 1) + '%5E' + matchesSearch.slice(1);
```
ghiscoding, line 150: same comment as above, it might be good to add a comment of what we're escaping?

test/cypress/e2e/example09.cy.ts
```diff
@@ -310,6 +310,29 @@ describe('Example 09 - OData Grid', () => {
...
 
+    it('should perform filterQueryOverride when operator "%%" is selected', () => {
+      cy.get('.search-filter.filter-name select').find('option').last().then((element) => {
+        cy.get('.search-filter.filter-name select').select(element.val());
+      });
+
+      cy.get('.search-filter.filter-name')
+        .find('input')
+        .clear()
+        .type('Jo%yn%er');
```
ghiscoding, line 321: awesome, nice to see SQL LIKE stuff in the demo 🚀

ghiscoding (changes requested):

ghiscoding:
@jr01 would you mind doing a quick review as well if you have time to spare? Thanks :)

zewa666:
I'll address the review feedback. thx

ghiscoding (approved):

jr01:
LGTM

docs/backend-services/OData.md
```diff
@@ -221,6 +221,30 @@ Navigations within navigations are also supported. For example `columns: [{ id:
...
+
+Column filters may have a `Custom` operator, that acts as a placeholder for you to define your own logic. To do so, the easiest way is to provide the `filterQueryOverride` callback in the OdataOptions. This method will be called with `OdataFilterQueryOverrideArgs` to let you decide dynamically on how the filter should be assembled.
+
+E.g. you could listen for a specific column and the active OperatorType.custom in order to switch the filter to a matchesPattern SQL LIKE search:
+
+```ts
+backendServiceApi: {
+  options: {
+    filterQueryOverride: (args) => {
+      const { fieldName, columnDef, columnFilterOperator, searchValue } = args;
```
ghiscoding, line 234: while reimplementing this in GraphQL, I noticed that these 2 lines can be simplified into 1 line by simply deconstructing the object directly in the function argument

```ts
filterQueryOverride: ({ fieldName, columnDef, columnFilterOperator, searchValue }) => {
```

examples/vite-demo-vanilla-bundle/src/examples/example09.ts
```diff
@@ -131,6 +145,16 @@ export default class Example09 {
           enableCount: this.isCountEnabled, // add the count in the OData query, which will return a property named "__count" (v2) or "@odata.count" (v4)
           enableSelect: this.isSelectEnabled,
           enableExpand: this.isExpandEnabled,
+          filterQueryOverride: (args) => {
+            const { fieldName, columnDef, columnFilterOperator, searchValue } = args;
```
ghiscoding, line 149: while reimplementing this in GraphQL, I noticed that these 2 lines can be simplified into 1 line by simply deconstructing the object directly in the function argument

```ts
filterQueryOverride: ({ fieldName, columnDef, columnFilterOperator, searchValue }) => {
```

ghiscoding (changes requested):
I left some new comments to look into and I also asked to replace `OdataFilterQueryOverrideArgs` with a more generic interface that I can reuse in the GraphQL Service as well since they have the exact same content as suggested in PR #1549

ghiscoding (approved):

PR #1538, opened 2024-05-22
feat(core): add new `preventDragFromKeys` grid option, fixes #1537

ghiscoding:
- the issue was brought in discussion #1537, the root cause was the mouse drag sometime kicks in when the user selects a few row by using the Ctrl+click combo, however in rare occasion the user might move by even a single pixel and that sends an onDrag event which the SlickCellRangeSelector picks up when then sends a new event `onCellRangeSelecting` and then the SlickRowSelectionModel assumes it's a new range from a mouse drag and override the previous range. However we should really prevent this mouse drag from happening when the user is pressing the Ctrl/Shift keys to avoid having the issue brought in #1537

### DIFF
--- a/packages/common/src/core/__tests__/slickInteractions.spec.ts
+++ b/packages/common/src/core/__tests__/slickInteractions.spec.ts
@@ -52,7 +52,7 @@ describe('Draggable class', () => {
     dg.destroy();
   });
 
-  it('should NOT trigger dragInit event when user is pressing mousedown with Ctrl key and is considered forbidden', () => {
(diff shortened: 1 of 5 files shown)


ghiscoding:
@zewa666 I wouldn't mind a review on this PR since I know you use this feature too and want to make sure it's safe to do so. See #1537 for more info

zewa666:
I've tried out the stackblitz PR review and did so for both Example 17 and Example 19 where I had somewhat similar issues. Both work fine. I couldn't replicate the shift behavior as mentioned in the other discussion but holding shift and clicking around in general causes weird text selections to happen so I didn't give that much thought anyways. But since it's configurable I don't see any big issues with not including it as default since the consumer can always decide which keys to take

PR #1541, opened 2024-05-23
fix(filters): Compound Operator alt texts should work with custom list

ghiscoding:
- instead of applying the Compound Operator alternate texts from `compoundOperatorNumeric` or `compoundOperatorString` only, we should instead apply the alt texts after either a custom list or a fixed list from mentioned functions.


ghiscoding:
cc @zewa666

PR #1543, opened 2024-05-24
feat(export): add missing `valueParserCallback` dataContext & new demo

ghiscoding:
- the `excelExportOptions.valueParserCallback` was missing the dataContext which can be useful when the exported data is not the current cell value but another cell value, for example our cell is Total but is a calculated value from 2 other column dataContext (Price & Qty)
- fix a bug found with new demo, the `excelExportOptions.style` wasn't being applied on regular data cell
- add a new Example 23 to demo Excel Export with Formula
- the new demo was created with the following logic (there are 3 calculated fields: SubTotal, Taxes & Total)
  - (in UI) - use Custom Formatters for calculated fields
  - (Excel Export) - use Excel Formula for calculated fields (via `excelExportOptions.valueParserCallback`)

#### TODOs
- [x] unit tests
- [x] add Cypress E2E tests (for UI only though, we'll test the export to Excel manually)
- [x] `excelExportOptions.filename` doesn't seem to work, need to investigate
- [x] make grid editable with editor styling
- [x] center alignment doesn't seem to work with custom export formatter
- [x] improve existing docs
- [x] might need to test with grouping, see what the new dataContext is filled with, I think it will be the group row object
  - I forgot that for Groups, we actually have `groupTotalsExcelExportOptions` which I had to improve to do everything I wanted in this demo :)
- [x] add more color styling to show full features
- [x] also calculate all Totals Sums via Excel Formulas as well

#### use Excel Formulas to calculate Totals by using other dataContext props
image

#### use Excel Formulas to calculate Group Totals 
image

### DIFF
--- a/examples/vite-demo-vanilla-bundle/src/examples/example23.ts
+++ b/examples/vite-demo-vanilla-bundle/src/examples/example23.ts
@@ -1,9 +1,23 @@
-import { type Column, Editors, type GridOption, Formatters, FieldType, type Formatter, } from '@slickgrid-universal/common';
+import {
+  type Aggregator,
+  Aggregators,
+  type Column,
+  Editors,
+  FieldType,
+  type Formatter,
+  Formatters,
+  type GridOption,
+  type Grouping,
+  GroupTotalFormatters,
+  type SlickGrid,
+  type SlickGroupTotals,
+} from '@slickgrid-universal/common';
 import { Slicker, type SlickVanillaGridBundle } from '@slickgrid-universal/vanilla-bundle';
 import { ExcelExportService } from '@slickgrid-universal/excel-export';
 
 import { ExampleGridOptions } from './example-grid-options';
 import './example23.scss';
+import { BindingEventService } from '@slickgrid-universal/binding';
 
 interface GroceryItem {
   id: number;
@@ -16,14 +30,8 @@ interface GroceryItem {
   total: number;
 }
 
-/**
- * Check if the current item (cell) is editable or not
- * @param {*} dataContext - item data context object
- * @param {*} columnDef - column definition
- * @param {*} grid - slickgrid grid object
- * @returns {boolean} isEditable
- */
-function checkItemIsEditable(_dataContext, columnDef, grid) {
+/** Check if the current item (cell) is editable or not */
+function checkItemIsEditable(_dataContext: GroceryItem, columnDef: Column, grid: SlickGrid) {
   const gridOptions = grid.getOptions();
   const hasEditor = columnDef.editor;
   const isGridEditable = gridOptions.editable;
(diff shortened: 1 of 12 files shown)


ghiscoding:
@zewa666 
So I know we talked about this a while ago and I always wanted to eventually go in depth on the subject of using Formula in Excel Export, so here it is... just because.... we can 🚀 😆 

Note that I was missing the `dataContext` argument in `excelExportOptions.valueParserCallback`, so I added it in here and tagged this PR as a feature because I don't think it would have been possible to do without the data context in this case (meaning creating a formula by using other props of the data context)

cc @jr01

zewa666:
damn don't show this to my boss 🤣

need to take some proper time to check this PR out but so far this looks amazing

examples/vite-demo-vanilla-bundle/src/examples/example23.ts
```diff
@@ -0,0 +1,259 @@
...
+      { id: i++, name: 'BBQ Chicken', qty: 1, taxable: false, price: 12.33 },
+      { id: i++, name: 'Chicken Wings', qty: 12, taxable: true, price: .53 },
+      { id: i++, name: 'Drinkable Yogurt', qty: 6, taxable: true, price: 1.22 },
+      { id: i++, name: 'Milk', qty: 3, taxable: true, price: 3.11 },
+    ];
+
+    return datasetTmp;
+  }
+
+  generatePhoneNumber(): string {
```
ghiscoding, line 252: we won't need this

ghiscoding:
@zewa666 cherry on top 🍒, I even got the demo to calculate Grouping Totals Sum (see both print screens above), your boss really shouldn't see this lol 🚀 

However please note that it takes a fair amount of code to get it all working, I also assume that perf is probably impacted as well a bit, but anyway the point was to show that it's possible. I assume that generic functions and code reuse should help to decrease the number of lines, I know that I have repeated code and that is to keep it readable and understandable by the users who looks at the code.

zewa666:
so much good stuff in this example. I didnt know about setting the metadata for excel exports. thats really great. where can I read up on the options? do you perhaps habe a link?

ghiscoding:
@zewa666 indeed you can do a lot with the cell value parser, I updated the associated [cell value parser](https://ghiscoding.gitbook.io/slickgrid-universal/grid-functionalities/export-to-excel#cell-value-parser) docs, you can find some info there. But I got a lot of these ideas when I migrated Excel-Builder to native code, for example the Excel Formulas I really got it from there when I read more about the metadata, so you'll get more info by going to the [excel-builder-vanilla](https://github.com/ghiscoding/excel-builder-vanilla) repo, read the docs and also try all the demos which I purposely made them WYSIWYG (what you see in the UI, you'll get in the export). In summary, whatever you can do in Excel-Builder, you should be able to do in SlickGrid as well 🚀

PR #1545, opened 2024-05-24
fix(filter): Date Filter should trigger Grid State change with Backspace

### LINKED ISSUE #1544: New DateFilter's clear does not remove the components value or call stateChange
### Describe the bug

I'm updating to Angular v18 and latest Angular-Slickgrid v8. During that I found that it seems there is a bug with clearing the value from the new VanillaJS datepicker.

besides not unsetting the components actual internal value, it also never calls a potential state change event, which I typically use to persist the selection/filters in localStorage

I'll check into that and see what might be the cause

### Reproduction

* open example https://ghiscoding.github.io/Angular-Slickgrid/#/trading
* set an execution time
* press backspace to clear it
* (no state change called if you modify and add the listener)
* close calender
* re-open
* notice that the value is still selected

Angular

### Environment Info

```shell
| (framework used)    | Angular (latest) |
```

### Validations

ghiscoding:
- fixes #1544

### DIFF
--- a/examples/vite-demo-vanilla-bundle/src/examples/example12.ts
+++ b/examples/vite-demo-vanilla-bundle/src/examples/example12.ts
@@ -639,7 +639,7 @@ export default class Example12 {
   }
 
   handleOnGridStateChanged(event) {
-    // console.log('handleOnGridStateChanged', event?.detail ?? '')
+    // console.log('handleOnGridStateChanged', event?.detail ?? '');
     const gridState = event?.detail?.gridState;
     if (Array.isArray(gridState?.rowSelection.dataContextIds)) {
       this.isMassSelectionDisabled = gridState.rowSelection.dataContextIds.length === 0;
--- a/packages/common/src/filters/dateFilter.ts
+++ b/packages/common/src/filters/dateFilter.ts
@@ -143,16 +143,17 @@ export class DateFilter implements Filter {
     // clear date picker + compound operator when Backspace is pressed
     this._bindEventService.bind(this._dateInputElm, 'keydown', ((e: KeyboardEvent) => {
       if (e.key === 'Backspace') {
-        this.clear(true);
+        this.clear(true, false);
       }
     }) as EventListener);
   }
 
   /** Clear the filter value */
-  clear(shouldTriggerQuery = true) {
+  clear(shouldTriggerQuery = true, shouldTriggerClearEvent = true) {
     if (this.calendarInstance) {
-      this._clearFilterTriggered = true;
+      this._clearFilterTriggered = shouldTriggerClearEvent;
       this._shouldTriggerQuery = shouldTriggerQuery;
+      this._currentValue = '';
       this.searchTerms = [];
       if (this._selectOperatorElm) {
         this._selectOperatorElm.selectedIndex = 0;
@@ -160,6 +161,12 @@ export class DateFilter implements Filter {
       if (this.calendarInstance.input) {
         this.calendarInstance.settings.selected.dates = [];
         this._dateInputElm.value = '';
+        this.calendarInstance.update({
+          dates: true,
+          month: true,
+          year: true,
+          time: true,
+        });
       }
     }
     this.onTriggerEvent(new Event('keyup'));


zewa666:
hmm not sure but this doesn't seem to work for me on Example10 at least

clear_bug

ghiscoding:
hmm I'll have to take another look at it later. Have you done anything troubleshooting on the code I changed? Like adding breakpoints or console logs?

zewa666:
I was looking at it from the website samples (https://ghiscoding.github.io/slickgrid-universal/#/example10)

ooh ... hold on I guess it's not deployed there yet. that could be the reason :)

ghiscoding:
@zewa666  hahaha yeah the demo is only being deployed on releases. I did that to avoid showing unreleased features 😉

ghiscoding:
> But anyways there seems to be more issues with the DatePicker, like I can now only paste formats in ISO (2020-01-07) format but not others like (07/01/2020) or (01.07.2020). So I need to dig deeper on that

So 1 thing to note is that the new lib can **only work** with Date in ISO format and also the associated input and its format is all done by code that I implemented (since again the external lib only accepts ISO format). I haven't tried copy+paste so that would explain but I thought that I've set the input as read only, so how did you paste a date? Should we really accept a paste though? I mean it's a date picker, so technically you should always use the picker UI

zewa666:
exactly the paste from Excel use case. My idea is anyways to allow only 3 specific formats. if its none of that I want to paste an Invalid Date forcing the user to fix it.

so ideally I could have an entry point where I do the conversion from X to ISO or error/null, so the lib can keep focusing on ISO, which I very like btw. Where would you see that fit

ghiscoding:
I'm not sure how the Excel Copy Buffer works, I added console log on the date editor `serializeValue()` and `loadValue()` but it doesn't go in there when pasting. It probably makes sense since at that point we don't have actually have the editor opened, so it must be coming from somewhere else but in that case it might just be a straight copy+paste without validation of any kind? You probably more than I do on how the plugin works since you worked on extending Example 19 with that kind of validation!?

zewa666:
I didnt have that feature yet and sat down to implement it since the old parser actually acted a but weird from different locales based excel sheets. hence I'm actually quite happy the new one just accepts ISO and forces it on the dev to convert tbe copy buffer to the matching format.

I guess I'll go with onBeforePasteCell and simply do the conversion there if the target column to be pasted from the range is a date editor.

ghiscoding:
yeah ok, I'll let you submit an update of the docs if you don't mind since you know the subject more than I do. Just 1 thing to note that is that when parsing, I purposely used `en-US` in my code because since I'm in Canada, I had a conversion problem when running locally (I had `en-CA`) vs CI (`en-US`) and the format was slightly different making my PR fails in CI, so to avoid format differences you can set a static locale to use in Tempo. The problem I had in my case is that on my local it was showing `a.m.` but in the CI it was showing as `am` (I even opened an [issue](https://github.com/formkit/tempo/issues/57) in Tempo and got clarification from the maintainer)... so in summary, it's just some "gotcha" to be informed of.

https://github.com/ghiscoding/slickgrid-universal/blob/aa0a74f339568ed58c605b74a1a4eea262b077a5/packages/common/src/filters/dateFilter.ts#L304-L305

zewa666:
yeah being from Austria I'm used to all sorts of weird stuff with i18n related conversion issues between en-US and de-AT. seen that one you mentioned and I think I've also proposed fixing everything CI related to en-US, at least thats what I used to do with Aurelia-I18n

ghiscoding:
I just edited my comment above with the following comments which explains it all I guess...

Finally the thing to note is that Tempo tries to use [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) as much as possible to keep the lib small, which is why I had some format differences

zewa666:
btw whats your recommended approach to testing out a non-released master merge with my angular-slickgrid app? just forcefully npm install straight from github-master vs npm? or should I do a local build of master repo and npm link?

ghiscoding:
my approach is very uncommon and crude, I run a full Slickgrid-Universal build and then copy the `dist` into each Angular-Slickgrid `node_modules/@slickgrid-universal/` packages that changed, it's crude and long. Also since I prefer to stay on Windows 10 for the time being, I don't have Tabs in file explorer so I use [Q-Dir](http://q-dir.com/) (as in Quad Directory Explorer) a small Windows freeware that allows me to use multiple tab in the same window and also allows to save your view in a favorite so it's a lot easier to manage with this freeware (see this youtube [video](https://www.youtube.com/watch?v=CUEiJYoqgvU) about it)

Another approach that is typically recommended is to use npm link (something like along this [article](https://medium.com/@AidThompsin/how-to-npm-link-to-a-local-version-of-your-dependency-84e82126667a)), it's probably better with yarn or pnpm though... but 1 of the reason I'm not using this approach is that there's no easy On/Off switch to remove this link, so once it's linked it kind of bypasses the original and released code. 

if you find something better, please let me know, for now I'll keep using 1st approach

zewa666:
i thought going with npm pack https://dev.to/scooperdev/use-npm-pack-to-test-your-packages-locally-486e

ghiscoding:
oh cool I wasn't aware of that one, but you probably need to find out the full path to get it working and update all packages, it shouldn't be too much of a problem. Have you tried the 3rd approach I mentioned above? Basically what you proposed earlier

zewa666:
nope, I actually prefer to stick to npm, thats the baseline for everyone, and only use pnpm in your repos. So that wont work for my app

ghiscoding:
if you want to go with npm pack, you could add npm scripts for each package (I've done that in Lerna-Lite for testing purposes) `"pack-tarball": "npm pack"`

PR #1549, opened 2024-05-28
feat(GraphQL): add `filterQueryOverride` to GraphQL Service

ghiscoding:
- similar approach to PR #1536 but for GraphQL Service
- adds a custom `filterQueryOverride` which provides a way for the user to override default filter operator and GraphQL query

image

### DIFF
--- a/docs/backend-services/graphql/GraphQL-Filtering.md
+++ b/docs/backend-services/graphql/GraphQL-Filtering.md
@@ -136,7 +136,7 @@ this.gridOptions = {
 
 ### Override the filter query
 
-Column filters may have a `Custom` Operator, that acts as a placeholder for you to define your own logic. To do so, the easiest way is to provide the `filterQueryOverride` callback in the GraphQL Options. This method will be called with `GraphqlFilterQueryOverrideArgs` to let you decide dynamically on how the filter should be assembled.
+Column filters may have a `Custom` Operator, that acts as a placeholder for you to define your own logic. To do so, the easiest way is to provide the `filterQueryOverride` callback in the GraphQL Options. This method will be called with `BackendServiceFilterQueryOverrideArgs` to let you decide dynamically on how the filter should be assembled.
 
 E.g. you could listen for a specific column and the active `OperatorType.custom` in order to switch the filter to an SQL LIKE search in GraphQL:
 
--- a/docs/backend-services/graphql/GraphQL-Filtering.md
+++ b/docs/backend-services/graphql/GraphQL-Filtering.md
@@ -2,6 +2,7 @@
 - [filterBy](#filterby)
 - [Complex Objects](#complex-objects)
 - [Extra Query Arguments](#extra-query-arguments)
+- [Override the filter query](#override-the-filter-query)
 
 ### Introduction
 The implementation of a GraphQL Service requires a certain structure to follow for `Slickgrid-Universal` to work correctly (it will fail if your GraphQL Schema is any different than what is shown below).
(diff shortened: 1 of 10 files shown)


ghiscoding:
@zewa666 so I think I got a similar approach to your OData PR #1536, however 1 major thing that makes it apart of that instead of returning a string, I'm asking the user to return an object of the shape `{ field: string; operator: string; value: any; }` which is what we end up with in the GraphQL query.

However, while doing this PR, I noticed that my `GraphqlFilterQueryOverrideArgs` is an exact copy of your `OdataFilterQueryOverrideArgs`, so I think it would be best if we merge them into 1 interface which could be named as `BackendServiceFilterQueryOverrideArgs` and because it's generic then I think you should also move it to a backend service interface like `backendServiceOption.interface.ts`. So would you mind doing that change to your PR? I can update my PR afterward by simply using that new interface.

I know you're not using GraphQL but rather OData, but still let me know if my PR makes any sense (I don't even have a GraphQL server to test with) but I followed this SO answer that inspired my approach: https://stackoverflow.com/a/37981802/1212166

zewa666:
sure I'll take a look tonight. btw can you target my branch as destination for the merge? I'm genuinly interested if thats possible, but no probs I can move the files around as you mentioned anyways.

with regards to testing perhaps a quick way to ramp up a server with auto graphql features would be to look into Hasura. or if you just want to spin up a demo on top of json files perhaps https://github.com/marmelab/json-graphql-server

docs/backend-services/graphql/GraphQL-Filtering.md
```diff
@@ -131,4 +132,27 @@ this.gridOptions = {
...
+
+### Override the filter query
+
+Column filters may have a `Custom` Operator, that acts as a placeholder for you to define your own logic. To do so, the easiest way is to provide the `filterQueryOverride` callback in the GraphQL Options. This method will be called with `GraphqlFilterQueryOverrideArgs` to let you decide dynamically on how the filter should be assembled.
+
+E.g. you could listen for a specific column and the active `OperatorType.custom` in order to switch the filter to an SQL LIKE search in GraphQL:
+
+> **Note** technically speaking GraphQL isn't a database query language like SQL, it's an application query language.
+> What that means is that GraphQL won't let you write arbitrary queries out of the box. It will only support the types of queries defined in your GraphQL schema.
+> see this SO: https://stackoverflow.com/a/37981802/1212166
```
zewa666, line 145: Depending on your GraphQL backend, chances are high a regex filter is supported. e.g here in Hasura https://hasura.io/docs/latest/queries/postgres/filters/text-search-operators/#_regex

so with that said, you could pass on the constructed Regex, like I'm doing in the OData sample, and have that perform the same feature as SQL LIKE would do. Keep in mind you're most likely looking for an case insenstive variante by either attaching `/i` or having an operator called `IRegex`  provided by your backend.

examples/vite-demo-vanilla-bundle/src/examples/example10.html
```diff
@@ -17,7 +17,8 @@ <h3 class="title is-3">
 </h3>
 <h6 class="title is-6 italic">
   <span class="text-red">(*) NO DATA SHOWN</span>
-  - just change any of Filters/Sorting/Pages and look at the "GraphQL Query" changing :)
+  - just change any of Filters/Sorting/Pages and look at the "GraphQL Query" changing.
+  Also note that the column Name has a filter with a custom %% operator that behaves like an SQL LIKE operator supporting % wildcards.
```
zewa666, line 21: same as above comment

zewa666 (commented):
aside from the small comment above LGTM

docs/backend-services/graphql/GraphQL-Filtering.md
ghiscoding: @zewa666 I wasn't aware of that, so what are you suggesting to change then? Should I reference your link as well or replace my link? I'd be happy to just copy+paste what you have in mind 😉

PR #1563, opened 2024-06-08
feat: paste multiline content as single cell

### LINKED ISSUE #1559: Quotes newlines/tabs excelcopybuffer
### Describe the bug

With the ExcelCopyBuffer I'd expect if a paste text is quoted with doublequotes, newlines within that text shouldn't break to new cells. Also when tabs are contained, the text after them is cut off.

I just wanted to check back with you whether you @ghiscoding  too think this is a wrong behavior. If so, I'll create a PR to fix that.

### Reproduction

* Go to example 19 (https://ghiscoding.github.io/slickgrid-universal/#/example19)
* Copy the following text into your editor of choice and replace \n with real newlines and \t with a tab key
```text
"this is\n
a \t newline\n
test"
```
* Check the result

Angular

### Environment Info

```shell
| (framework used)    | Angular |
```

### Validations

zewa666:
this changes the default behavior to paste multiline strings (formatted with quotes e.g by Excel) as a single cell as opposed to parsing it's content and performing multi-row pastes.

Additionally it adds an option to remove the quotes after paste as well as treat newlines with alternative characters.

fixes #1559


zewa666:
there are still some Cypress tests missing as well as a couple of docs.

packages/common/src/interfaces/excelCopyBufferOption.interface.ts
```diff
@@ -53,6 +52,11 @@ export interface ExcelCopyBufferOption<T = any> {
   /** option to specify a custom column header value extractor function */
   headerColumnValueExtractor?: (columnDef: Column<T>) => string | HTMLElement | DocumentFragment;
 
+  /** if the copied text starts and ends with a double-quote (Excel multiline string) replace newlines with the defined character. (default: false) */
+  replaceNewlinesWith?: string | false;
+
+  /** multiline strings copied from Excel are pasted with double quotes. Should those be removed? (default: false) */
+  removeDoubleQuotesOnPaste?: boolean;
```
ghiscoding, line 59: more options 😄 looking good

zewa666: oh boy yeah I feel bad for being responsible for so many of them 😅

ghiscoding:
so I just googled SlickGrid to see if any recent articles came out and I'm just speechless at this article, nearly all of it is wrong, I'm just like wow this is so wrong...  I had to share with someone this [article](https://trycatchdebug.net/news/1183790/angular-build-failure-with-angular-slickgrid) 🤣 

There's no mention of any the Migration Guides I wrote (he probably didn't read them either), neither the  Angular Update website and jumping so many versions in a single go is never recommended and finally all his infos of what to change are just so plain wrong.. reinstalling SlickGrid when I got rid of it... just wow! 🤦🏻‍♂️ oh and of course there's no "reply" button of any kind so there's no way to correct them either. I guess that I will never do business with them lol

---

apart from that, I pushed a couple of fixes and I think I'm done with everything, I'll wait for your PR and probably push next weekend again. It's now kind of hard to believe but I think I'm done with the project. I don't have any new features planned. 🚀

zewa666:
Its AI week for me. working with Semantic Kernel and building bots. so Grid is on hold but I plan to finish this one by the weekend.

first round of user tests came in and brought some good insights plus a bug report, which I'll create together with a PR.

Re article, it definitely wasnt too insightful but neither it felt offensive. as such I'd just take it easy ;)

having no new features is a good thing. should give us some room to even further harden the current code base and iron out some uncovered bugs.

zewa666:
For the love of testing ... I can't figure out how to copy and paste properly with Cypress. Essentially I wanted to test in the second test stub that if you paste the same as in the previous but this time quoted with double quotes, it should stay in one column.

Other than that, the PR is ready to go

ghiscoding:
Yeah some things are harder than needs to be in Cypress, just googling it, have you tried something similar to this SO Answer or the one following it?
https://stackoverflow.com/a/68871590/1212166 

For some of the Cypress tests I used the console to log certain things and inspect it (I mainly used it to test Grid State), so that might work too!? The only downside that I had with this approach is that Mocha (which is what Cypress use) doesn't have any function to check if an object contain 1 property, so I had to compare for the entire object equality and it's hard in some cases. 

https://github.com/ghiscoding/slickgrid-universal/blob/025888d0ceb0a29095b6036ab645de365bd08e8a/test/cypress/e2e/example09.cy.ts#L64-L67

But anyway if you think it's too complicate and we don't need these tests, I'm happy to merge as it is (at least we have the unit tests)

docs/grid-functionalities/excel-copy-buffer.md
```diff
@@ -126,4 +126,11 @@ this.gridOptions = {
 };
 ```
 
-This way you have full control, via the args parameters, to get a ref to the current row and cell being updated. Also keep in mind that while performing a buffered table paste (e.g three cols at once) and only one of them has a denying condition, the other cells paste will pass successfully.
\ No newline at end of file
+This way you have full control, via the args parameters, to get a ref to the current row and cell being updated. Also keep in mind that while performing a buffered table paste (e.g three cols at once) and only one of them has a denying condition, the other cells paste will pass successfully.
+
+### Handling quoted multiline pastes
+When copying Excel cells, where the value is a multiline text, the copy buffer will result in a quoted text buffer. Slickgrid will properly handle the case when this content itself contains `\n` characters, to not break onto multiple rows as well as not split on the next column if `\t` characters are contained.
```
ghiscoding, line 132: we should probably had a note that it requires version 5.2 and higher as a subtitle or a quoted note

```ts
#### requires v5.2.0 and higher

// OR

> Note: requires v5.2.0 and higher
```

test/cypress/e2e/example19.cy.ts
```diff
@@ -253,4 +253,29 @@ describe('Example 19 - ExcelCopyBuffer with Cell Selection', () => {
...
+        .should('have.text', 'this is');
+
+      cy.getCell(15, 14, '', { parentSelector: '.grid19', rowHeight: GRID_ROW_HEIGHT })
+        .should('have.text', 'a test');
+
+      cy.getCell(16, 14, '', { parentSelector: '.grid19', rowHeight: GRID_ROW_HEIGHT })
+        .should('have.text', 'with newlines');
+    });
+
+    it('should keep quoted content on the same cell', () => {
```
ghiscoding, line 278: probably better to remove this empty test if you can't add anything to test underneath ;)

zewa666:
I'll give it one more try tonight and otherwise remove the tests and finish the PR

zewa666:
nope, no chance. It's the mix of paste plus keydown listener that can't be simulated properly. So whatever, Unit tests will have to suffice here.

Rdy for merge from my side

ghiscoding:
okie dokie, I'm working on a new Example for Footer Totals Row and I'll done. I assume that you already pushed everything you wanted? I'll push a release by next week probably

PR #1575, opened 2024-06-14
feat(filters): add new optional `filterShortcuts` to Column Filter

### LINKED ISSUE #1569: Filter for blanks / non blanks

Similar to Excel it would be helpful if we could filter e.g. text columns to only show blank / non blank cells

https://docs.aspose.com/cells/net/how-to-filter-blanks-and-non-blanks/2.png

### Suggested solution

I'm not sure where to place that feature. One idea I had was about an advanced filter, signaled by a gear icon placed next to the filter, which would popup a menu. in there buttons for blank/non blank could be placed. so it would be somewhat similar to the command popup.

### Alternative

so far I can simulate non blanks by setting does not equal and `""` as filter value. but that feels a bit weird

### Additional context

_No response_

### Validations

- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

ghiscoding:
closes #1569
- user will be able to declare some Filter Shortcuts that will be added to the Column Header Menu that the shortcuts were defined, it can be declared via an array that must include at least a `title` (or `titleKey`) and a `searchTerms` array, lastly an optional `operator` can also be provided
- the available properties of these shortcut is a merge of Header Menu Item (except `command` and `action` which are reserved) and an addition of our required props which are the extra `searchTerms` and `operator` (which as you can guess are the same props as a Column Filter for example what we use in `preset`)
- here's a code sample of it
```ts
this.columnDefinitions = [{
    id: 'country', name: 'Country of Origin', field: 'country',
    filter: {
      model: Filters.inputText,
      filterShortcuts: [
        { title: 'Blank Values', searchTerms: ['A'], operator: '<', iconCssClass: 'mdi mdi-filter-minus-outline', },
        { title: 'Non-Blank Values', searchTerms: ['A'], operator: '>', iconCssClass: 'mdi mdi-filter-plus-outline', },
      ]
    },
  },
];
```

#### TODOs
- [x] add unit tests
- [x] add Cypress tests
- [x] add documentation
- [x] test UI with Locale translations

#### demo

brave_iNBMcswLvc

### DIFF
--- a/docs/column-functionalities/filters/compound-filters.md
+++ b/docs/column-functionalities/filters/compound-filters.md
@@ -9,6 +9,7 @@
 - [Update Filters Dynamically](input-filter.md#update-filters-dynamically)
 - [How to avoid filtering when only Operator dropdown is changed?](#how-to-avoid-filtering-when-only-operator-dropdown-is-changed)
 - [Custom Filter Predicate](input-filter.md#custom-filter-predicate)
+- [Filter Shortcuts](input-filter.md#filter-shortcuts)
 
 ### Description
 Compound filters are a combination of 2 elements (Operator Select + Input Filter) used as a filter on a column. This is very useful to make it obvious to the user that there are Operator available and even more useful with a date picker (`Vanilla-Calendar`).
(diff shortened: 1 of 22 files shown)


ghiscoding:
cc @zewa666 @jr01 

so the only thing I'm not sure is the position in the Column Header Menu, it's starting to be crowded a little. I think it's ok where I've put it but just in case, I also expose a configurable `filterShortcutsPositionOrder` (that's the only built-in one that I exposed and it's current position index is 55)

image

jr01:
Hey @ghiscoding - I think this looks pretty good and I don't feel it's crowded at all! Go for it!

zewa666:
great work, looks sweet. looking forward to this one

ghiscoding:
@zewa666 I released everything last night, we should be good for a while (I would think anyway 😄). Cheers and thanks for all

zewa666:
awesome, thx a lot gonna give it a try tomorrow

PR #1606, opened 2024-07-15
fix: revert to old behavior but properly generate picker labels

### LINKED ISSUE #1605: custom HTMLElement inside header loses event handler
### Describe the bug

This bug was first reported in Angular-Slickgrid but is really code related to Slickgrid-Universal, so I'm closing the Angular-Slickgrid issue and opening this new one with same info

> When I use a custom HTMLElement inside the name property of a column definition like this
> 
> ```
>   const div = document.createElement('div');
>   const button = document.createElement('button');
>   button.textContent = 'click';
>   div.appendChild(button);
>   button.addEventListener('click', () => alert('hi'));
> 
>     this.columnDefinitions = [
>       { id: 'title', name: div, field: 'title' },
> ```
> 
> In angular slickgrid the click event handler does not trigger whereas if I do the same in a normal slickgrid (6pac repo) grid the click event is fired.

### Reproduction

> https://stackblitz.com/edit/stackblitz-starters-akwtsl?file=src%2Fslickgrid-angular%2Fslickgrid-angular.component.html
> 

Angular

### Environment Info

```shell
| Angular-Slickgrid   | 8.3.1 |
| Slickgrid-Universal | 5.3.2 |
```

### Validations

zewa666:
fixes #1605

### DIFF
--- a/packages/common/src/core/slickGrid.ts
+++ b/packages/common/src/core/slickGrid.ts
@@ -593,8 +593,7 @@ export class SlickGrid<TData = any, C extends Column<TData> = Column<TData>, O e
         if (emptyTarget) {
           emptyElement(target);
         }
-        const node = options?.cloneNode ? val.cloneNode(true) : val;
-        target.appendChild(node);
+        target.appendChild(val);
       } else {
         // when it's already empty and we try to reassign empty, it's probably ok to skip the assignment
         const skipEmptyReassignment = options?.skipEmptyReassignment !== false;
@@ -2793,7 +2792,7 @@ export class SlickGrid<TData = any, C extends Column<TData> = Column<TData>, O e
         w = this.columns[i].width || 0;
 
         rule = this.getColumnCssRules(i);
-        if (rule.left) {          
+        if (rule.left) {
           rule.left.style.left = `${x}px`;
         }
         if (rule.right) {
--- a/packages/common/src/global-grid-options.ts
+++ b/packages/common/src/global-grid-options.ts
@@ -279,5 +279,10 @@ function pickerHeaderColumnValueExtractor(column: Column, gridOptions?: GridOpti
   if (headerGroup) {
     return headerGroup + columnGroupSeparator + column.name;
   }
+
+  if (column?.name instanceof HTMLElement || column?.name instanceof DocumentFragment) {
+    return column.name.textContent ?? '';
+  }
+
   return column?.name ?? '';
 }


packages/common/src/core/slickGrid.ts
```diff
@@ -593,8 +593,7 @@ export class SlickGrid<TData = any, C extends Column<TData> = Column<TData>, O e
         if (emptyTarget) {
           emptyElement(target);
         }
-        const node = options?.cloneNode ? val.cloneNode(true) : val;
```
ghiscoding, line 596: could you also remove the options prop above `cloneNode?: boolean;`  since it's no longer needed

packages/common/src/global-grid-options.ts
```diff
@@ -279,5 +279,10 @@ function pickerHeaderColumnValueExtractor(column: Column, gridOptions?: GridOpti
   if (headerGroup) {
     return headerGroup + columnGroupSeparator + column.name;
   }
+
+  if (column?.name instanceof HTMLElement || column?.name instanceof DocumentFragment) {
+    return column.name.textContent ?? '';
+  }
+
```
ghiscoding, line 286: so I had a very similar approach but I think it would be better to go with mine instead, because with yours, you forgot to deal with the line 280 

```suggestion
function pickerHeaderColumnValueExtractor(column: Column, gridOptions?: GridOption) {
  let colName = column?.name ?? '';
  if (colName instanceof HTMLElement || colName instanceof DocumentFragment) {
    colName = colName.textContent || '';
  }
  const headerGroup = column?.columnGroup || '';
  const columnGroupSeparator = gridOptions?.columnGroupSeparator ?? ' - ';
  if (headerGroup) {
    return headerGroup + columnGroupSeparator + colName;
  }
  return colName;
}
```

ghiscoding:
@zewa666 oh I'm not sure why you closed your PR? 

I had another possible problem with the `.textContent`, this will read all element texts but is that really ideal though? I mean I just tested this with an HTML element that has some text and a button with text and it gives me what is shown below (the button text is also pulled)

image

that is using the following code

```ts
const div = document.createElement('div');
    const button = document.createElement('button');
    button.className = 'button button-small';
    button.textContent = 'click';
    button.addEventListener('click', () => alert('hi'));
    div.appendChild(document.createTextNode('Column 1 '));
    div.appendChild(button);

    this.columnDefinitions = [
      {
        id: 'sel', name: div, field: 'num', width: 80, type: FieldType.number,
        excludeFromExport: true,
        resizable: true,
        filterable: true,
        selectable: false,
        focusable: false
      },
```

PR #1608, opened 2024-07-17
feat: support tooltips on icon formatters

zewa666:
adds the possibility to define tooltips via the title tag on icon formatters

### DIFF
--- a/packages/common/src/formatters/__tests__/iconFormatter.spec.ts
+++ b/packages/common/src/formatters/__tests__/iconFormatter.spec.ts
@@ -20,4 +20,12 @@ describe('the Icon Formatter', () => {
     const result = iconFormatter(0, 0, input, { field: 'user', params: { formatterIcon: icon } } as Column, {}, {} as any);
     expect((result as HTMLElement).outerHTML).toBe(`<i class="${icon}" aria-hidden="true"></i>`);
   });
+
+  it('should always return a <i> with the title attribute provided in the "title" property from "params"', () => {
+    const input = null;
+    const title = 'This is a title';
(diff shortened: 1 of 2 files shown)


zewa666:
this is just a minor feature but something I stumbled on the other day when working on a new feature.

zewa666:
Supabase is quite interesting though I'd personally prefer Hasura. Have been using it with super great success for a larger project and would prefer that everytime again. The thing is, once persistance and handling CRUD becomes a free thing along with permissions etc, all handled by fundamental DB principles, nothing beats the productivity boost. Plus, I guess the GraphQL layer should nicely integrate with Slickgrid, though I have to admit I haven't yet tried so.

ghiscoding:
ah yeah you mentioned it last time when we did the Header Menu thing. I think I'd also personally prefer GraphQL over OData, so Hasura make sense in this case. Thanks for the info

ghiscoding:
@zewa666 
I think that the biggest challenge with Hasura would be to create the structure that I defined for the Pagination/Sorting/Filtering. I assume that's all doable in Hasura and perhaps we could create a new package in universal to support Hasura following this structure that I require. I'm just guessing here since I haven't tried anything and barely read their docs

zewa666:
havent yet looked into slickgrids graphql backend but I guess it should be doable to adjust things to match Hasuras style.

the biggest pro with Hasura where it live-queries, aka graphql subscriptions but with server push on change. that alone is so very cool it could be worth the hassle (think of your realtime example)

Hasura cloud offers a free tier. as for db go for something like https://neon.tech/pricing

PR #1611, opened 2024-07-18
fix: provide all search values

zewa666:
meh there was an issue where suddenly on multi-selects I didn't get the value of the second and onwards selection

### DIFF
--- a/examples/vite-demo-vanilla-bundle/src/examples/example09.ts
+++ b/examples/vite-demo-vanilla-bundle/src/examples/example09.ts
@@ -145,9 +145,9 @@ export default class Example09 {
           enableCount: this.isCountEnabled, // add the count in the OData query, which will return a property named "__count" (v2) or "@odata.count" (v4)
           enableSelect: this.isSelectEnabled,
           enableExpand: this.isExpandEnabled,
-          filterQueryOverride: ({ fieldName, columnDef, columnFilterOperator, searchValue }) => {
+          filterQueryOverride: ({ fieldName, columnDef, columnFilterOperator, searchValues }) => {
             if (columnFilterOperator === OperatorType.custom && columnDef?.id === 'name') {
-              let matchesSearch = (searchValue as string).replace(/\*/g, '.*');
+              let matchesSearch = searchValues[0].replace(/\*/g, '.*');
               matchesSearch = matchesSearch.slice(0, 1) + CARET_HTML_ESCAPED + matchesSearch.slice(1);
               matchesSearch = matchesSearch.slice(0, -1) + '$\'';
 
--- a/examples/vite-demo-vanilla-bundle/src/examples/example10.ts
+++ b/examples/vite-demo-vanilla-bundle/src/examples/example10.ts
@@ -213,13 +213,13 @@ export default class Example10 {
             field: 'userId',
             value: 123
           }],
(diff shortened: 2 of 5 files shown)


zewa666:
hmmm do you think its ok to change the signatures for graphql as well @ghiscoding?

ghiscoding:
oh I see, that was a mistake in the implementation, yes it should be plural since we always deal with a `searchTerms` array and yes we should update the GraphQL and anywhere else. This feature is relatively new, so it should be ok to update it and that seems to be why the build failed in your PR as well

ghiscoding:
There's another line to change it seems

> Error: src/examples/example10.ts(216,79): error TS2339: Property 'searchValue' does not exist on type 'BackendServiceFilterQueryOverrideArgs'.

zewa666:
yeah just saw it ... not sure if I fixed it properly though as the Graphql example doesn't show any results

ghiscoding:
> yeah just saw it ... not sure if I fixed it properly though as the Graphql example doesn't show any results

the GraphQL example only shows the query string, so technically just take a look at the query to see if it work as intended or not. That is also what the E2E tests are testing

ghiscoding:
looks good, I'll go ahead and merge. 

Have you seen my new draft PR #1609? I'm not sure if I'll wait for this PR to be finished before pushing a new release or simply go without it

zewa666:
yeah I saw it, didnt have time to properly check it out, plan to do so at home. sounds definitely interesting if only for feature comparison reasons to the big player. I personally feel though that virtual scrolling has far more drawbacks compared to pagination in server-scenarios. nevertheless, I'll give it certainly a try.

next week I'll finally start to combine AI together with the grid and I'm super excited about the potential use cases. the slickgrid project at work made a huge impact so people feel fine to invest a more time 🎉

ghiscoding:
> next week I'll finally start to combine AI together with the grid and I'm super excited about the potential use cases. the slickgrid project at work made a huge impact so people feel fine to invest a more time 🎉

glad to hear this, I rarely hear good stories, I usually hear more about the issues haha

For the release, I think there's enough to go ahead with a new one this weekend and I also want to take my time for that Infinite Scroll feature and it will take time to add tests and also add the feature for GraphQL as well (currently testing with OData since I have test data) and maybe infinite scroll for local dataset (not sure yet)

zewa666:
wont say no to that ;)

to give you an idea where new potential features might go with our app (just hacked this together in an hour with a local Mistral LLM and context of the whole row)

grid_ai_yeah

ghiscoding:
So your last column as a Formatter with a button for the AI, is that it? I haven't much of AI stuff at my work yet but I should dig more into this new tech, it's definitely the future going forward. So is that AI button reading the new row data you've input and provide suggestions or whatever else with that data? It seems quite interesting. It reminds me a bit when I added a translate button for users with different language and they wanted to see it translated on the fly

Side note, my draft PR is also Example 27, so I'm not sure if you wanted to add this new example to my repo but if so, we'll for sure conflict :D

zewa666:
dont worry I just took whatever number I saw free. Your example with on the fly translation is actually quite close to it. What one does with it depends on how the prompt is built. In my case I've just instructed it to improve the failure description by using so far text + title.

the interesting part is going to be able to create the prompt along with a template, which allows interpolating the active or other rows to provide context.

I'll play around with it next week and let you know more

PR #1612, opened 2024-07-18
fix: queryselector should support numeric ids

zewa666:
I just tried to integrate the composite editor and it failed with the error that `[data-editor=1234]` is an invalid selector. Guess what, my column ids are numeric strings :)


ghiscoding:
haha I didn't think about that special case, thanks again :)

PR #1615, opened 2024-07-19
fix: don't use Document Fragment in SlickCheckbox for Salesforce Usage

good-ghost:
Salesforce does not supoort DocumentDefragment, so add condition preventDocumentFragmentUsage for checkboxSelectorColumn.


ghiscoding:
are you having some problems with the creation of a Pull Request? Let me know if you need help

ghiscoding:
reopening since your Pull Request seems valid to me. Have you tested it in Salesforce? if you want to bundle it and update the zip file, you can run this pnpm command from the root
```sh
pnpm run bundle:zip
```

ghiscoding (approved):

ghiscoding:
@good-ghost I'll go ahead and merge this since I think this is correct now. I'll push a new release today and will update the zip file as well by the end of the day. Thanks for your contribution

Issue #1618, opened 2024-07-23
Composite Editor manipulating DataContext via editor save method not persisted


zewa666:
### Describe the bug

When using the CompositeEditor it renders also my custom editor. this Foreign Key lookup editor does modify not only the bound column but also others in the dataContext (label + value) in its save command.

When I do this via cell edits, all is good and leaving the cell will modify both the bound column and my modified additional column.

If I do the same within the onSave of the CompositeEditor (e.g `dataContext["foo"] = "bar"` it works as well, just not in that editors save command as mentioned.

### Reproduction

as described above

### Which Framework are you using?

Angular

### Environment Info

```shell
| Executable          | Version |
| ------------------- | ------- |
| (framework used)    | VERSION |
| Slickgrid-Universal | VERSION |
| TypeScript          | VERSION |
| Browser(s)          | VERSION |
| System OS           | VERSION |
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/ghiscoding/slickgrid-universal/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] Read the [Wikis](https://github.com/ghiscoding/slickgrid-universal/wiki).
- [X] Check that there isn't [already an issue](https://github.com/ghiscoding/slickgrid-universal/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/ghiscoding/slickgrid-universal/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

ghiscoding:
I don't really understand your problem, since with the Composite Editor you're supposed to use the Composite Editor `onSave` callback and then use either the 1st arg `formValues` or the last arg `dataContextOrUpdatedDatasetPreview`. Have you read the docs that I wrote? It's been a while since I've look at the code

```ts
this.compositeEditorInstance?.openDetails({
  onSave: (formValues, _selection, dataContextOrUpdatedDatasetPreview) => {
    // save it
  }
}
```

zewa666:
I did yeah and as said it works if done in the onSave callback. yet its a super weird unexpected behavior that it does work in cell edit mode from custom editors save method but the same editor and  save function running inside composite mode behaves different

ghiscoding:
[...]

PR #1619, opened 2024-07-24
chore: reproduce bug

zewa666:
!!!DO NOT MERGE, JUST FOR ERROR REPRODUCTION AND SEEING IT IN STACKBLITZ!!!

Here's the repro of https://github.com/ghiscoding/slickgrid-universal/issues/1618.
if you take a look at the example12 you'll notice that the duration column now has a custom editor in use. If you change the value of a duration cell, you'll notice that the cost column should be updated to 1234657 as well after cell leave.

If you do the same via the composite editor however, that's not happening.

### DIFF
--- a/examples/vite-demo-vanilla-bundle/src/examples/example12.ts
+++ b/examples/vite-demo-vanilla-bundle/src/examples/example12.ts
@@ -10,6 +10,7 @@ import {
   type Formatter,
   Formatters,
   type GridOption,
+  InputEditor,
   type LongTextEditorOption,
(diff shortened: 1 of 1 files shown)


ghiscoding:
I was trying to update and rebase the PR, I've never done that, but I guess it's only doable by the author of the PR? 

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/keeping-your-pull-request-in-sync-with-the-base-branch

PR #1621, opened 2024-07-25
perf: use SortableJS ESM import to save 1kb

ghiscoding:
I'm not sure if it's worth it or not but we seem to be using 1kb from the zip when using the ESM build of SortableJS. However the thing that is not as great is that `@types/sortablejs` is not recognized when using 
`import Sortable from 'sortablejs/modular/sortable.core.esm.js';`

So I had to use something like below with a `@ts-ignore` on the ESM import
```ts
// @ts-ignore: use the ESM imports for smaller build, it however doesn't play nice with @types/sortablejs
import Sortable from 'sortablejs/modular/sortable.core.esm.js';
import type { Options as SortableOptions, SortableEvent } from 'sortablejs'; // from @types/sortablejs
```

#### before
image

#### after
image

### DIFF
--- a/packages/common/src/core/slickGrid.ts
+++ b/packages/common/src/core/slickGrid.ts
@@ -1,4 +1,3 @@
-import Sortable, { type SortableEvent } from 'sortablejs';
 import { BindingEventService } from '@slickgrid-universal/binding';
 import {
   classNameToList,
@@ -13,6 +12,9 @@ import {
   isDefinedNumber,
   isPrimitiveOrHTML,
 } from '@slickgrid-universal/utils';
+// @ts-ignore: use the ESM imports for smaller build, it however doesn't play nice with @types/sortablejs
+import Sortable from 'sortablejs/modular/sortable.core.esm.js';
+import type { Options as SortableOptions, SortableEvent } from 'sortablejs'; // from @types/sortablejs
 
 import {
   type BasePubSub,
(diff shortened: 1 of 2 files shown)


ghiscoding:
hmm that breaks Jest unit tests since Jest only supports CJS, perhaps I should revisit this if (when) I move from Jest to Vitest

zewa666:
well you could potentially make Jest play nice with ESM. https://jestjs.io/docs/ecmascript-modules
I tried that though quite a while ago and it was a mess. Hence why I moved to vitest

ghiscoding:
> well you could potentially make Jest play nice with ESM. https://jestjs.io/docs/ecmascript-modules I tried that though quite a while ago and it was a mess. Hence why I moved to vitest

I tried that with Lerna-Lite and it was also a large mess and I also decided to use Vitest on Lerna-Lite and was happy with the change. The only downside is that for the case of Lerna-Lite, Vitest is really slow on Windows when using a lot of `child_process` that Lerna uses, but that won't be the case here so yeah I want to migrate the project to Vitest eventually but ocnsidering that I have close to 5k unit tests, that will be a challenge. Anyway, it's super low priority.

zewa666:
glad to hear its not me being the only one having trouble with getting esm play nice with jest. 😅

ghiscoding:
closing for now, might revisit if and when we migrate to Vitest to support ESM

PR #1644, opened 2024-08-17
feat: migrate back to Vanilla-Calendar-Pro

ghiscoding:
- migrate to [`vanilla-calendar-pro`](https://github.com/uvarov-frontend/vanilla-calendar-pro) and drop the custom fork `vanilla-calendar-picker` (which was always meant to be temporary) that I had created until all features I needed were going to be merged into the original project Vanilla-Calendar-Pro and we are just about to be there.
- it's very close to be ready but it's still missing a couple of small fixes
- NOTE: the Types are not imported the same way, they must be imported `from 'vanilla-calendar-pro/types'` instead of the default import

### DIFF
--- a/docs/column-functionalities/editors/date-editor-(vanilla-calendar).md
+++ b/docs/column-functionalities/editors/date-editor-(vanilla-calendar).md
@@ -5,9 +5,7 @@
 - See the [Editors - Wiki](../Editors.md) for more general info about Editors (validators, event handlers, ...)
 
 ### Information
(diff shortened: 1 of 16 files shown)


ghiscoding:
[Vanilla-Calendar-Pro](https://github.com/uvarov-frontend/vanilla-calendar-pro) now has **all** features and fixes that I had in my fork, so we can now officially switch back to that repo.. woohoo 🎉

Issue #1634, opened 2024-08-07
Issue with enableExcelCopyBuffer and Hidden Columns Causing Selection Offset


hwlv:
### Describe the bug

When using the enableExcelCopyBuffer option set to true in slickgrid-universal, I’ve encountered an issue where selecting a range of cells causes the selection area to be offset if there is a hidden column in the grid. It appears that the hidden column is being included in the calculations, leading to discrepancies between the mouse position and the selected cells.

### Reproduction

1. Set enableExcelCopyBuffer to true.
2. Add a column with the hidden property set to true.
3. Attempt to select a range of cells that includes both visible and hidden columns.

<img width="1116" alt="image" src="https://github.com/user-attachments/assets/6028d4d6-e6a7-438a-bfb8-4b22f4f0a24a">

### Which Framework are you using?

React

### Environment Info

```shell
| Executable          | Version |
| ------------------- | ------- |
| (framework used)    | VERSION |
| Slickgrid-Universal | VERSION |
| TypeScript          | VERSION |
| Browser(s)          | VERSION |
| System OS           | VERSION |
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/ghiscoding/slickgrid-universal/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] Read the [Wikis](https://github.com/ghiscoding/slickgrid-universal/wiki).
- [X] Check that there isn't [already an issue](https://github.com/ghiscoding/slickgrid-universal/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/ghiscoding/slickgrid-universal/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

ghiscoding:
cc @zewa666

ghiscoding:
The `hidden` column property was something that was added not that long ago in the `6pac/slickgrid` project but it is not not fully supported. The Grid Presets is much more supported to hide columns

PR #1646, opened 2024-08-21
fix: SlickCellExternalCopyManager should use DataView

ghiscoding:
- while investigating for issue #1634, I found the problems shown below
- so it's a bit surprising that we didn't catch this earlier but the plugin was only coded for SlickGrid usage to get read/write data and without the DataView, but Slickgrid-Universal is all built with the DataView usage only and so we should really fix this to support both
- add a new `DataWrapperService` to make it easier to call the correct functions depending on our usage SlickGrid or SlickDataView

### DIFF
--- a/packages/common/src/extensions/__tests__/slickCellExternalCopyManager.spec.ts
+++ b/packages/common/src/extensions/__tests__/slickCellExternalCopyManager.spec.ts
@@ -5,7 +5,7 @@ import type { Column, GridOption, OnEventArgs, } from '../../interfaces/index';
 import type { SlickCellSelectionModel } from '../slickCellSelectionModel';
 import { SlickCellExternalCopyManager } from '../slickCellExternalCopyManager';
 import type { InputEditor } from '../../editors/inputEditor';
-import { SlickEvent, SlickEventData, type SlickGrid, SlickRange } from '../../core/index';
+import { type SlickDataView, SlickEvent, SlickEventData, type SlickGrid, SlickRange } from '../../core/index';
 import { type BasePubSubService } from '@slickgrid-universal/event-pub-sub';
 
 const pubSubServiceStub = {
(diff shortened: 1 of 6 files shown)


ghiscoding:
cc @zewa666

zewa666:
the new wrapper might come in really handy. could you shortly elaborate on what the difference between the dataview vs old approach is?

ghiscoding:
Welcome back @zewa666 😉 

So the differences between SlickGrid and the SlickDataView is mainly that SlickGrid is very basic in data functionalities, it has functions to add/remove/select but that's about it (if you wanted to just create a simple grid to display things then just using SlickGrid was enough, i.e. this [basic example](https://6pac.github.io/SlickGrid/examples/example1-simple.html)). On the other hand, the original author also created an optional DataView to add more functionalities like Grouping and Aggregators, which SlickGrid itself doesn't have, and all its indirect features (group/totals/nested groups). Other features that the DataView has are custom Filter & Sort, local Pagination and probably more. Also the theory is that the DataView was created like a plugin, any user could come and create it's own and just swap it out whenever he wants, like the [slick.remotemodel-yahoo.ts](https://github.com/6pac/SlickGrid/blob/master/src/slick.remotemodel-yahoo.ts) extends on this idea... so considering that the DataView has much more functionalities, including the Grouping, I decided from the start to only use the DataView in my projects on the expense of being a bit larger (it has 1800 lines of code), I never cared about the size but some people do (hence why some people still prefer to use the original project).

And finally below, is one of the issue that I found when investigating the other issue mentioned above, the `grid.getData()` returns the dataset when using plain SlickGrid **but** when we add a DataView (it always does in our case because like I said above, I always add it) and so `grid.getData().length` will probably throw because that doesn't exist in the DataView, then other calls like `grid.getDataItem(i)` will return nothing because the dataset is not in SlickGrid but rather in the DataView, so there's about 10 of these calls that you can see in this PR. As for the wrapper, I only added the ones that were used by this plugin and typically most user will never use or care about this but I prefer to keep the plugins in sync and to keep the grid functions like the original ones so that if I there are PRs on that side, I could copy these code changes on my side too (for example the auto-scroll when dragging cell/row selection (this [example](https://6pac.github.io/SlickGrid/examples/example-auto-scroll-when-dragging.html)), was added afterward and so I took the code and applied it to my plugins afterward)

I hope you don't mind long explanation 😉 

https://github.com/ghiscoding/slickgrid-universal/blob/b231d8915beee1925298141d313e733ebaac2235/packages/common/src/extensions/slickCellExternalCopyManager.ts#L276-L281

zewa666:
awesome. could you perhaps release that as a (next) version? that way I could more easily give it a try in our app if you're up for some testdrive

ghiscoding:
> awesome. could you perhaps release that as a (next) version? that way I could more easily give it a try in our app if you're up for some testdrive

I'm going to do another release this weekend mostly because of the other PR #1644 since all my PRs on Vanilla-Calendar-Pro were accepted and merged (about a dozen of them), so I'm quite happy about that. There also other small fixes here and in Angular-Slickgrid that are unreleased, so yeah another release for next weekend

PR #1652, opened 2024-08-24
fix: getComputedStyle should be called from the window object

ghiscoding:


ghiscoding:
doesn't to work as intended, let's close it

PR #1662, opened 2024-09-02
fix: pasting multiple new rows

zewa666:
I guess this feature was actually never properly tested but when pasting new rows at the bottom, with the defined `newRowCreator`, the previously auto created empty rows were not having an individual unique id, causing the subsequent actions to fail.

This changes it so that the burden is on the dev to provide the actual new row addition. At least that way he/she can control the newly created ids.

If you see this ok I would actually also update the docs to desribe the `newRowCreator` behavior.

### DIFF
--- a/packages/common/src/extensions/slickCellExternalCopyManager.ts
+++ b/packages/common/src/extensions/slickCellExternalCopyManager.ts
@@ -280,11 +280,7 @@ export class SlickCellExternalCopyManager {
 
     // ignore new rows if we don't have a "newRowCreator"
     if ((availableRows < destH) && typeof this._addonOptions.newRowCreator === 'function') {
-      const d = this._dataWrapper.getDataItems();
-      for (addRows = 1; addRows <= (destH - availableRows); addRows++) {
-        d.push({});
-      }
-      this._dataWrapper.setDataItems(d);
+      this._addonOptions.newRowCreator(destH - availableRows);
       this._grid.render();
     }
 
--- a/packages/common/src/extensions/slickCellExternalCopyManager.ts
+++ b/packages/common/src/extensions/slickCellExternalCopyManager.ts
@@ -280,7 +280,15 @@ export class SlickCellExternalCopyManager {
 
     // ignore new rows if we don't have a "newRowCreator"
     if ((availableRows < destH) && typeof this._addonOptions.newRowCreator === 'function') {
-      this._addonOptions.newRowCreator(destH - availableRows);
+      const rowsToAdd = destH - availableRows;
+      const rowsBeforePaste = this._dataWrapper.getDataLength();
+      this._addonOptions.newRowCreator(rowsToAdd);
+      const rowsAfterPaste = this._dataWrapper.getDataLength();
+
+      if (rowsAfterPaste !== rowsBeforePaste + rowsToAdd) {
+        console.warn(`[Slickgrid-Universal] The "newRowCreator" did not add the correct amount of rows, it should add "${rowsToAdd}" rows but it added "${rowsAfterPaste - rowsBeforePaste}" rows`);
+      }
+
       this._grid.render();
     }
 
--- a/packages/common/src/extensions/__tests__/slickCellExternalCopyManager.spec.ts
+++ b/packages/common/src/extensions/__tests__/slickCellExternalCopyManager.spec.ts
@@ -642,14 +642,17 @@ describe('CellExternalCopyManager', () => {
       });
 
       it('should Copy, Paste and run Execute clip command', (done) => {
-        const mockNewRowCreator = jest.fn();
+        let nextAddNewRowId = 0;
+        const mockNewRowCreator = jest.fn((count: number) => {
+          for (let i = 0; i < count; i++) {
+            gridStub.getData<SlickDataView>().addItem({
+              id: nextAddNewRowId--
+            });
+          }
+        });
         const mockOnPasteCells = jest.fn();
-        const renderSpy = jest.spyOn(gridStub, 'render');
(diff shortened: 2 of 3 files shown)


zewa666:
oh and btw I did also find one more place where the composite editor selection went wrong due to missing quotes

ghiscoding:
On the Composite Editor [docs](https://ghiscoding.gitbook.io/slickgrid-universal/grid-functionalities/composite-editor-modal#create-new-item) when adding a new item, I went with 2 possible approaches
- user can optionally provide his own id via `insertNewId` option
- user can omit the id, if so the id will be dataset length + 1

```ts
this.compositeEditorInstance?.openDetails({
      headerTitle: modalTitle,
      modalType,
      // insertNewId: 1234, // you can provide a custom Id (defaults to last Id+1)
      // insertOptions: { position: 'bottom' }, // if you wish to add the item to the bottom (defaults to top of the grid)
      onError: (error) => alert(error.message), // you should define how to deal with error coming from the modal

      // you can optionally provide an async callback method when dealing with a backend server
      onSave: (formValues, selection, dataContext) => {
        // simulate a backend server call which returns true (successful) after 30sec
        return new Promise(resolve => setTimeout(() => resolve(true), 500));
      }
    });
```

I never use this feature, so you're probably best to think what the correct approach is but still I think we can do the same as what I've done in Composite for that one. 

On my side, I started working on migrating from Jest to Vitest, I still have 270 that are failing (out of 5000) which I need to debug, so that will keep me busy for a while. I think Jest is faster than Vitest on Windows however, but still I want to drop CJS on next major (probably won't be until next year though)

zewa666:
oh thats great news with regards to Vitest, let me know if I can help.

the idea with the default length+1 does not really make sense as the user is supposed to provide a newRowCreator which should handle the job. so the code as was would actually fail bc it would do a double addition. As for id generation I think its kinda dangerous to assume defaults as the ids might already be jagged due to deletes, and in combo with softdeletes create issues.

ghiscoding:
ah well ok, like I said, I don't use this feature so much, so go ahead with what you think is best. But for the id generation, I think the dataset length + 1 is still an ok approach by default and if that is not correct for some then just use the optional `insertNewId` which seems fine for me. Anyway if the `newRowCreator` is required then it's another story.

zewa666:
but how would I infer that the newRowCreator function wasnt creating rows and than doing it as fallback with length+1? please take another look at the diff, I hope from the conditional surrounding the change it should clarify what I meant

packages/common/src/extensions/slickCellExternalCopyManager.ts
```diff
@@ -280,11 +280,7 @@ export class SlickCellExternalCopyManager {
 
     // ignore new rows if we don't have a "newRowCreator"
     if ((availableRows < destH) && typeof this._addonOptions.newRowCreator === 'function') {
-      const d = this._dataWrapper.getDataItems();
-      for (addRows = 1; addRows <= (destH - availableRows); addRows++) {
-        d.push({});
-      }
-      this._dataWrapper.setDataItems(d);
```
ghiscoding, line 287: but if you remove this, then how will the item be added to the dataset? You're forcing the user to do that part itself, is that it? isn't this a breaking change though?

zewa666: it is but the old behavior didnt work either as it clashed with existing ids, more so it actually created elements with an undefined id

ghiscoding: can we somehow add some console warning in the case that the user forget to add it?

ghiscoding: @zewa666 I've opened my PR for the Vitest migration and since it's touching every test files, it will certainly conflict with your PR. So I'd like to merge your PR before Vitest but would it be possible to add a console warning like I mentioned just above?

zewa666: sure will try to do so today

zewa666: took a bit longer but now it's in. Was it something like this you meant?

ghiscoding: The console warning I asked for was mainly for what you wrote above, if the user forgets to do it himself, then show them a warning that they didn't. I'm not sure if your new console covers this or not, you know more than I do on that part. 

> and yes, I'd expect the user to do it by himself as thats actually the only safe way to handle it imo

Side note, this new console will be missing test coverage, you could also ignore it with Istanbul (or v8 for upcoming Vitest), though if you can add a unit test then use something like below for spying on the console

https://github.com/ghiscoding/slickgrid-universal/blob/e97017aa786259fb947cff39f1ce4204c0cfedee/packages/utils/src/__tests__/domUtils.spec.ts#L90-L93

zewa666: I guess thats the only way to check that, cant make any assumptions about how the resulting ids should look like.

yep will update the tests. thx for the sample console check

ghiscoding (approved):

ghiscoding:
looks good, merging :)

The next step will be to update & merge my other Vitest migation PR (probably tonight). 
@zewa666 I assume that issue #1618 is still out of the loop for now, right?

zewa666:
yep feature got postponed. first conditional columns and footer rows are on my plate ;)

we can close the PR as well if you prefer and I'll reopen once back at it

ghiscoding:
that's fine, I just like to have updates on the status once in a while, sorry to ask too often 😉 

BTW, I got some small questions about Vitest, but I'll ask them in the Vitest PR #1663

PR #1666, opened 2024-09-07
chore: add SlickCellExternalCopyManager test coverage

ghiscoding:
@zewa666 oops it looks like I merged the PR a little too early, the code below is not being covered. I did add coverage for the first line but the condition below is not covered. It actually looks like it's not even possible to go into the `if` condition, because the `addRows` variable is assigned to `0` and so how can it even go into the `if` condition?

is the code shown below still needed?

image

### DIFF
--- a/packages/common/src/extensions/__tests__/slickCellExternalCopyManager.spec.ts
+++ b/packages/common/src/extensions/__tests__/slickCellExternalCopyManager.spec.ts
@@ -651,7 +651,9 @@ describe('CellExternalCopyManager', () => {
           .mockReturnValueOnce([new SlickRange(0, 1, 2, 2)])
           .mockReturnValueOnce(null as any);
 
+        let clipCommand;
         const clipboardCommandHandler = (cmd) => {
+          clipCommand = cmd;
           cmd.execute();
         };
         plugin.init(gridStub, { clearCopySelectionDelay: 1, clipboardPasteDelay: 1, includeHeaderWhenCopying: true, clipboardCommandHandler, newRowCreator: mockNewRowCreator, onPasteCells: mockOnPasteCells });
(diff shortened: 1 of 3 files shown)


zewa666:
hmmm havent seen that line. where is that code from? also do I need to do anything specific to get the new tests to run or merely pnpm install/bundle and than run test again?

ghiscoding:
@zewa666 I did some more research and the code came from the original SlickGrid but was no longer covered after your PR because you removed the `newRowCreator` which was the only place that was incrementing the `addRows` variable. But now that the creator is remove, this code is totally unreachable (`addRows` is always set to 9 and so the `if` condition is always `false`).

Here's the code before your PR removed

https://github.com/ghiscoding/slickgrid-universal/blob/7b485962024a4ce336ebe80d223ddba4b7e80824/packages/common/src/extensions/slickCellExternalCopyManager.ts#L281-L289)

and here's your PR changes

image

so should I just get rid of code since it's unreachable or does that previous `for` loop still had something that we should add back?

zewa666:
thats really weird, as far as I get the code was merely dropping and re-adding the same rows. is that to cause a rerender on a fresh new array instance?

I've tried my fix first locally and so far we havent noticed any issues but who knows what this was for. I'll check that out once more but from a quick first look i'd say off it goes that unused code 😅

ghiscoding:
yeah sure, let me remove it and you'll have more chances to test it over in the coming weeks so if something is wrong I'm sure you'll provide a PR because like I said before, I'm not really using this plugin myself. 😉

zewa666:
copy and paste is so huge for my company as our tool is here to replace all these renegade excel files. so naturally people try to copy and paste the hell back and forth.

Imagine how awesome it is to debug issues there, like the last one where randomly 2 cols werent pasted but the others were. just to find out that the excel had 3 merged cells, the copy buffer though still contains the individual cols and therefore tabs 🤣

PR #1670, opened 2024-09-11
perf: Reduce iterative time complexity in Tree Data, fixes #1668

### LINKED ISSUE #1668: Reducing time complexity
### Describe the bug

We found that the due to the below code, the time complexity is O(n square) which can be reduced to O(n)

Modified code:
```
 DataView.getAllSelectedFilteredItems = function () {
      /* istanbul ignore if */
      if (!Array.isArray(this.selectedRowIds)) {
        return [];
      }

      const selectedRowIdSet = new Set(this.selectedRowIds); // Add new set 
      const intersection = this.filteredItems.filter((a) => selectedRowIdSet.has(a[this.idProperty]));
      return (intersection || []);
    }
```

### Reproduction

For large data set, the page hange for 1 minute

React

### Environment Info

```shell
```

### Validations

Vsinghal339-source:
fixes #1668

Optimize getAllSelectedFilteredItems and preSelectedRowIdsChangeFn with Set lookup and reducing time complexity from O(n square) to O(n)

### DIFF
--- a/packages/common/src/core/slickDataview.ts
+++ b/packages/common/src/core/slickDataview.ts
@@ -1453,7 +1453,8 @@ export class SlickDataView<TData extends SlickDataItem = any> implements CustomD
           } else {
             if (preserveHiddenOnSelectionChange && gridOptions.multiSelect) {
               // remove rows whose id is on the list
-              rowIds = this.selectedRowIds?.filter((id) => args.ids.indexOf(id) === -1);
+              const argsIdsSet = new Set(args.ids);
+              rowIds = this.selectedRowIds?.filter((id) => !argsIdsSet.has(id));
             } else {
               rowIds = [];
             }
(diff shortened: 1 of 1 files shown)


Vsinghal339-source:
@ghiscoding Please review this PR

ghiscoding:
It looks good, thanks a lot for the contribution 😀 

Just out of curiosity, you said it was previously hanging for a minute. How long does it take after the fix, is it just a few seconds now?

Vsinghal339-source:
for 40k records, the code was iterating over 40k * 40K recoreds which takes approx 30-40 seconds. We tried with 60k records and time exponentially increased to 1 minutes. But now it comes down to 400 ms.

Note: Even with new approach of Set(), although time is reduced to 400ms but it will still impact the sudden CPU spikes. We need to create the Set() of data only once and save it in globally under parent object and incrementally modify it. With this approach we can further optimise the app and reduce time from 400ms to < 1-2 ms

ghiscoding:
Great, thanks for the explanation and of course if you find further improvement that can be made to improve performance then yes please contribute more PRs 😉

Vsinghal339-source:
I would like to know when the next release will be made. Current version of the tag is v5.6.1. When will the newest tag (v5.6.2) be made?

ghiscoding:
next weekend or the week after

PR #1675, opened 2024-09-12
fix(TreeData): Reset the childrens prop when unflattening dataset in case it is being reused

### LINKED ISSUE #1657: Tree Data Grid x Parent/Child Relation Dataset x Failures on subsequent component mount
### Discussed in https://github.com/ghiscoding/slickgrid-universal/discussions/1655

<div type='discussions-op-text'>

<sup>Originally posted by **tnaum-ms** August 28, 2024</sup>
I'm currently investigating an issue that arises when I mount my SlickGrid Tree Data as a React component for the second time. The problem occurs when I toggle between different data view formats in my application. Specifically, switching away from the Tree Data view and then returning to it results in a failure of the Tree Data Component.

The data set I’m working with follows a Parent/Child structure, where the `id` and `parentId` fields are of type `string`.
_(It would be helpful to include a note in the documentation indicating the preferred or faster data input method.)_

The SlickGrid Tree Data Grid throws the following error:

> Each data element must implement a unique 'id' property

Upon investigation, I discovered that on the second mount (not just a second render), one of the internal data mutation functions introduces duplicates into my data set. This causes the `ensureIdUniqueness` function to fail.

The issue seems to be related to the function located here:
https://github.com/ghiscoding/slickgrid-universal/blob/ce8b4000cf72fa78293e50fba886ae883c08f39b/packages/common/src/services/utilities.ts#L158

This specific loop appears to be introducing duplicates:
https://github.com/ghiscoding/slickgrid-universal/blob/ce8b4000cf72fa78293e50fba886ae883c08f39b/packages/common/src/services/utilities.ts#L173-L187

I'm continuing to investigate the root cause, but since this issue only occurs on the second mount, I haven’t yet identified any obvious side effects. If this is something you’ve encountered before, I would greatly appreciate any insights or suggestions you could share.

Thank you for your assistance.
Best, Tomasz
 </div>

jano-kucera:
fixes #1657 again

Recreate the children array from scratch everytime instead of entity lookups.

There were some lint errors which I had to fix to be able to commit this.

### DIFF
--- a/packages/common/src/services/utilities.ts
+++ b/packages/common/src/services/utilities.ts
@@ -193,7 +193,10 @@ export function unflattenParentChildArrayToTree<P, T extends P & { [childrenProp
   // make them accessible by guid on this map
   const all: any = {};
 
-  inputArray.forEach((item: any) => all[item[identifierPropName]] = item);
+  inputArray.forEach((item: any) => {
+    all[item[identifierPropName]] = item;
+    delete item[childrenPropName];
+  });
 
   // connect childrens to its parent, and split roots apart
   Object.keys(all).forEach((id) => {
@@ -205,13 +208,7 @@ export function unflattenParentChildArrayToTree<P, T extends P & { [childrenProp
       if (!(childrenPropName in p)) {
         p[childrenPropName] = [];
       }
-      const existIdx = p[childrenPropName]?.findIndex((x: any) => x[identifierPropName] === item[identifierPropName]);
-      if (existIdx >= 0) {
-        // replace existing one when already exists (probably equal to the same item in the end)
-        p[childrenPropName][existIdx] = item;
-      } else {
-        p[childrenPropName].push(item);
-      }
+      p[childrenPropName].push(item);
       if (p[collapsedPropName] === undefined) {
         p[collapsedPropName] = options?.initiallyCollapsed ?? false;
       }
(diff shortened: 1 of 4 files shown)


packages/common/src/services/utilities.ts
```diff
@@ -193,7 +193,10 @@ export function unflattenParentChildArrayToTree<P, T extends P & { [childrenProp
   // make them accessible by guid on this map
   const all: any = {};
 
-  inputArray.forEach((item: any) => all[item[identifierPropName]] = item);
+  inputArray.forEach((item: any) => {
+    all[item[identifierPropName]] = item;
+    delete item[childrenPropName];
```
ghiscoding, line 198: the only thing I wonder is why are you deleting **after** and not **before** the assignment, so why not inverse lines 197 & 198 ?

jano-kucera: These lines are interchangeable without any affect on each other.
 
Line 197 maps the item reference under the set id field.
Line 198 removes the children property from the item. 

I've added the line here as it was already a place where all items were iterated already.

ghiscoding: ok thanks, let's merge and hope this fixes it all :)

ghiscoding:
This is now release, but I even found another huge perf improvement via PR #1681 but of course I found it after I had already released everything lol 😆 So I'll wait a little before releasing this last perf, everything else is released. Thanks for the contribution. Cheers ⭐

PR #1679, opened 2024-09-13
feat: matching border color for range selector in darkmode

zewa666:
I needed a new project to get to know the react wrapper of Slickgrid Universal better. So why not build a whole VSCode extension right ? ;)

cell-selector-darkmode

Also, I used that as the first time to try out the darkMode in more detail and noticed this little gotcha when selecting cells via mouse (at the end of the gif). This PR fixes it.

### DIFF
--- a/packages/common/src/core/__tests__/slickGrid.spec.ts
+++ b/packages/common/src/core/__tests__/slickGrid.spec.ts
@@ -1159,6 +1159,15 @@ describe('SlickGrid core file', () => {
 
       expect(rowSelectSpy).toHaveBeenCalled();
     });
+
+    it('should clear previous selection model when calling setSelectionModel() with a different model', () => {
+      const cellSelectionModel = new SlickCellSelectionModel();
+
+      grid = new SlickGrid<any, Column>(container, [], columns, { ...defaultOptions, darkMode: true });
+      cellSelectionModel.init(grid);
+
+      expect(cellSelectionModel.cellRangeSelector.addonOptions.selectionCss).toBe('2px solid white');
+    });
   });
 
   describe('Node Getters', () => {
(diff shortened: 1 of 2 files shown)


zewa666:
ouch ... turns out this ones much harder than thought since the grid isn't yet available in the ctor. The `!` was misleading me to believe its already initiated.

Hm I could recreate the SlickCellRangeSelector inside init, or expose a public selectionCss override so we can keep the instance.

ghiscoding:
Wow this is quite interesting, there's so much we can do with custom extensions in VSCode nowadays, that's really something 😮 

> ouch ... turns out this ones much harder than thought since the grid isn't yet available in the ctor. The `!` was misleading me to believe its already initiated.
> 
ahh yeah that's my fault, the thing is that I often use this pattern because of these plugins have no choice but to execute the `init()` method, so technically speaking, the grid is already defined except in the constructor. If I make it optional then it adds optional chaining everywhere in the code even though we know that every plugin must be initialized no matter what. So I kinda went with `!` because of that assumption but like you said, it's not available in the constructor. The thing too is that adding optional chaining everywhere will increase the code size, maybe less nowadays with newer browser supporting optional chaining by default but still (and I'm targeting`--target es2021` for both ESM & CJS so maybe it's ok)

> Hm I could recreate the SlickCellRangeSelector inside init, or expose a public selectionCss override so we can keep the instance.

it seems to work, so why not! 
Just let me know when you're done with the PR, it looks fine from what I can see and if you want to refactor things around `!` then feel free (though like I said, I did this in most of the plugins, that is the assumption of every plugin must be initialized)

zewa666:
yeah seams so at least, hadnt had any issues so far.

I'll take one more look at it tomorrow and let you know when rdy to merge. perhaps I can find more stuff while at it

ghiscoding:
I'll go ahead and merge, just because I'm going ahead with a new release. Thanks

zewa666:
one thing I havent checked but suspect is still missing is to react to option changes. but nevermind  that can also be another PR

Issue #1684, opened 2024-09-18
Datepicker with manual input


zewa666:
### Clear and concise description of the problem

The new vanilla calender is working great, in combo with copy & paste you're also able to quickly fill out values. Sometimes though, its simply easier to change the date by actually typing it in. Currently slickgrid doesn't provide this feature.

### Suggested solution

When a date cell editor is opened, there should be an option to configure readonly mode. If disabled, users can enter dates in the input field and they get synced back to the date control.

![image](https://github.com/user-attachments/assets/43fc2674-d584-4b82-9133-e0fbd43d0948)

### Validations

- [X] Follow our [Code of Conduct](https://github.com/ghiscoding/slickgrid-universal/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] Read the [Wikis](https://github.com/ghiscoding/slickgrid-universal/wiki).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

ghiscoding:
Ah yeah that's probably a decent addition to go with, in Flatpickr we had the `allowInput`

> Allows the user to enter a date directly into the input field. By default, direct entry is disabled.

https://github.com/ghiscoding/slickgrid-universal/blob/11b565e0e2623d20669b838263fb1638f53ea5dc/packages/common/src/editors/__tests__/dateEditor.spec.ts#L222-L226

I think it might be ok to add this as a new option into the VanillaCalendarOption interface, however I think it should stay as disabled or unset by default

https://github.com/ghiscoding/slickgrid-universal/blob/3c6ce204dd9b48091b4c49c5c07b5799b3af34b9/packages/common/src/interfaces/vanillaCalendarOption.interface.ts#L10-L15

zewa666:
I didnt know about the FlatPickr option but agree that it should be kept as allowEdit and yep it should stay false as default for backwards compat. I'll update the PR with those info

PR #1685, opened 2024-09-19
feat: allow overriding readOnly behavior of dateEditor

zewa666:
So this is a general idea of how I thought of tackling the realted issue https://github.com/ghiscoding/slickgrid-universal/issues/1684

No tests or docs yet as I first would like to validate what you think about the feature request @ghiscoding

### DIFF
--- a/examples/vite-demo-vanilla-bundle/src/examples/example11.ts
+++ b/examples/vite-demo-vanilla-bundle/src/examples/example11.ts
@@ -175,7 +175,7 @@ export default class Example11 {
         type: FieldType.date, outputType: FieldType.dateIso,
         filterable: true,
         filter: { model: Filters.compoundDate },
-        editor: { model: Editors.date, massUpdate: true, disabled: false, editorOptions: { allowInput: true } },
+        editor: { model: Editors.date, massUpdate: true, editorOptions: { allowInput: true } },
       },
       {
         id: 'finish', name: 'Finish', field: 'finish', sortable: true, minWidth: 80,
--- a/examples/vite-demo-vanilla-bundle/src/examples/example11.ts
+++ b/examples/vite-demo-vanilla-bundle/src/examples/example11.ts
@@ -175,7 +175,7 @@ export default class Example11 {
         type: FieldType.date, outputType: FieldType.dateIso,
         filterable: true,
         filter: { model: Filters.compoundDate },
-        editor: { model: Editors.date, massUpdate: true },
+        editor: { model: Editors.date, massUpdate: true, disabled: false, readOnly: false },
       },
       {
         id: 'finish', name: 'Finish', field: 'finish', sortable: true, minWidth: 80,
@@ -293,7 +293,7 @@ export default class Example11 {
             }
           } else if ((event.target as HTMLElement).classList.contains('mdi-check-underline')) {
             this.slickerGridInstance?.gridService.updateItem({ ...dataContext, completed: true });
-            alert(`The "${dataContext.title}" is now Completed`);
+            alert(`The "${dataContext.start}" is now Completed`);
           }
         }
       },
--- a/examples/vite-demo-vanilla-bundle/src/examples/example11.ts
+++ b/examples/vite-demo-vanilla-bundle/src/examples/example11.ts
@@ -175,7 +175,7 @@ export default class Example11 {
         type: FieldType.date, outputType: FieldType.dateIso,
         filterable: true,
         filter: { model: Filters.compoundDate },
-        editor: { model: Editors.date, massUpdate: true, disabled: false, readOnly: false },
+        editor: { model: Editors.date, massUpdate: true, disabled: false, editorOptions: { allowInput: true } },
       },
       {
         id: 'finish', name: 'Finish', field: 'finish', sortable: true, minWidth: 80,
@@ -293,7 +293,7 @@ export default class Example11 {
             }
           } else if ((event.target as HTMLElement).classList.contains('mdi-check-underline')) {
             this.slickerGridInstance?.gridService.updateItem({ ...dataContext, completed: true });
-            alert(`The "${dataContext.start}" is now Completed`);
+            alert(`The "${dataContext.title}" is now Completed`);
           }
         }
       },
--- a/packages/common/src/editors/dateEditor.ts
+++ b/packages/common/src/editors/dateEditor.ts
@@ -186,7 +186,7 @@ export class DateEditor implements Editor {
           title: this.columnEditor && this.columnEditor.title || '',
           className: inputCssClasses.replace(/\./g, ' '),
           dataset: { input: '', defaultdate: this.defaultDate },
-          readOnly: this.columnEditor.readOnly === false ? false : true,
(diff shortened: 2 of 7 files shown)


packages/common/src/editors/dateEditor.ts
```diff
@@ -186,7 +186,7 @@ export class DateEditor implements Editor {
           title: this.columnEditor && this.columnEditor.title || '',
           className: inputCssClasses.replace(/\./g, ' '),
           dataset: { input: '', defaultdate: this.defaultDate },
-          readOnly: this.columnEditor.readOnly === false ? false : true,
+          readOnly: this.columnEditor.editorOptions?.allowEdit === true ? true : false,
```
ghiscoding, line 189: probably easier to do `!!(this.columnEditor.editorOptions?.allowEdit)`

zewa666: passing on options to other options I'm typically careful since true is not the same as truthy but you're probably right here

zewa666:
need to update the docs and e2e, but first I need to get the kids to sleep ;)

packages/common/src/editors/dateEditor.ts
ghiscoding: I've used this `!!(var)` for years, it's an easy shortcut to convert the output to a boolean

zewa666: thats right but `"false" == true`  and can happen relatively easy as consumer of the grid. chances are the user meant false just mistankely added quotes

ghiscoding: ahhh if that happens then yes I agree 😆

test/cypress/e2e/example11.cy.ts
```diff
@@ -22,6 +22,20 @@ describe('Example 11 - Batch Editing', () => {
     cy.get('h3').should('contain', 'Example 11 - Batch Editing');
   });
 
+  it('should input values directly in input of datepicker', () => {
+    cy.get(`[style="top: ${GRID_ROW_HEIGHT * 0}px;"] > .slick-cell:nth(5)`)
+      .click();
+
+    cy.get(`.input-group-editor`)
+      .focus()
+      .type('{backspace}'.repeat(10))
```
ghiscoding, line 31: might be easier to simply use Ctrl+Backspace

examples/vite-demo-vanilla-bundle/src/examples/example11.ts
```diff
@@ -175,7 +175,7 @@ export default class Example11 {
         type: FieldType.date, outputType: FieldType.dateIso,
         filterable: true,
         filter: { model: Filters.compoundDate },
-        editor: { model: Editors.date, massUpdate: true },
+        editor: { model: Editors.date, massUpdate: true, disabled: false, editorOptions: { allowInput: true } },
```
ghiscoding, line 178: is there a particular reason to add the `disabled: false`?

zewa666: just an oversite from a manual test. good catch

test/cypress/e2e/example11.cy.ts
zewa666: {ctrl}{backspace} didn't work for me. But I doubt there is any con with this approach. So if you're fine I'd just keep it that way

ghiscoding: ok sure whatever works is fine, just thought the alternative was shorter but if it doesn't work then too bad

ghiscoding: the E2E test still fails though, should we use `{enter}` key after typing the date?

zewa666:
hmm not sure why the tests now failed ... is this some random sideeffect?

ghiscoding:
> hmm not sure why the tests now failed ... is this some random sideeffect?

I don't think so, I saw it failing at least twice in your PR, so I think it's a legitimate failure. Does it pass locally?

packages/common/src/editors/dateEditor.ts
ghiscoding: @zewa666 ah this is probably why it fails, it looks like you forgot rename this readonly to `allowInput`

zewa666:
yes it did, and the above mentioned readOnly is fine, thats the attribute that gets attached to the input.

I'll take a look again later today

zewa666:
alright the issue was as expected, those other tests looked for the number of modified cells, which got messed up due to my test starting at the begin and adding another one. Moved it to the end and now all should be fine

ghiscoding:
> alright the issue was as expected, those other tests looked for the number of modified cells, which got messed up due to my test starting at the begin and adding another one. Moved it to the end and now all should be fine

ah yeah that's the side effect of configuring Cypress like I did where the sequence is important. That is actually discouraged by Cypress, but I prefer to use this approach to avoid having tests that are super long because the Cypress default approach is to make every tests independent and I don't like that part so..

@zewa666  On the a different subject, I had a user who opened an issue in Angular-Slickgrid complaining that date sorting was slow for 50k rows and I closed the issue because I thought I couldn't improve it further and also because the person is using an old version... however, most of my examples are using Date object which is much quicker, but when we use another format (i.e. US format) then the date has to be parsed and then convert to timestamp to sort and that is in fact slowing down a lot the sort (25k rows took a few seconds to sort). The bad thing is that the default browser sort will probably read the same row multiple times, meaning parsing the same date multiple times. So my question is have you ever use any other Sort than the default browser `.sort()`? From what I found so far, some users provide suggestions but it's usually an entire new sort function which is not connectable to the default browser sort and that is probably not going to work with my project. Any suggestions? I'll search some more but ideally I would like to override the default browser sort, which I'm already doing, but use a different type of sort for large dataset (quick sort, or whatever else)

zewa666:
closed the issue, sry for wrong tagging.

as for the sorting performance, since I'm using server side backends I cant really tell. what I can tell is that sorting more than 50k rows on the client sounds like a huge UX issue. really if its more than ~100 rows I highly question the use of client side at all. thats where the server excels

ghiscoding:
yeah well you know, there's always a ton of users who still prefer to have everything loaded locally for whatever reasons (even if there's not valid reasons). I even saw 1 guy mentioning in SlickGrid repo that he loads few million rows which seems overkill to me, but if SlickGrid makes it usable then yeah you'll have users that go for it... 

I think that there's couple of reasons why pagination is not always good to use, there are some features that aren't supported with Pagination and those are Tree Data, Row Detail and most notably Grouping and this later one is probably a huge one for some

Anyway, I'll search some more and see if I can find anything that can improve perf, it's mostly just the Dates that are slow, sorting string or numbers are typically quite fast but Dates requiring parsing which slows it down a lot.

zewa666:
could you perhaps create a sample repro to look at? I dig its about changing to a custom sorter but hard to follow without concrete examples

ghiscoding:
Yeah I modified Example 2 with US Dates and that's when I noticed the much slower sort after I click on the Add 25K rows button and then try to sort by Start/Finish dates. I can open a draft PR with that after I update the Cypress tests that will surely fail, I'll do that later today after work hours :)

PR #1690, opened 2024-09-23
fix: home and end keys stop propagation on input based editors, closes #1686

### LINKED ISSUE #1686: Home/End keys in active editor
### Describe the bug

When a cell editor is focused, the arrow keys left and right are intercepted and cell navigation disabled. This is good as it allows to navigate within the Inputs text. Same works with CTRL + Left/Right to jump to start/end of the editor.

what does not work is hitting the pos1 (jumps to first cell of row) or end key (jumps to last cell of the active row)

I wonder whether this is an expected behavior or a bug. I can provide a PR if this indeed is an unintended behavior

### Reproduction

see above

Angular

### Environment Info

```shell
```

### Validations

zewa666:
closes https://github.com/ghiscoding/slickgrid-universal/issues/1686

### DIFF
--- a/packages/common/src/editors/__tests__/autocompleterEditor.spec.ts
+++ b/packages/common/src/editors/__tests__/autocompleterEditor.spec.ts
@@ -196,30 +196,19 @@ describe('AutocompleterEditor', () => {
       expect(editor.getValue()).toBe('male');
     });
 
-    it('should dispatch a keyboard event and expect "stopImmediatePropagation()" to have been called when using Left Arrow key', () => {
-      const event = new (window.window as any).KeyboardEvent('keydown', { key: 'ArrowLeft', bubbles: true, cancelable: true });
-      const spyEvent = vi.spyOn(event, 'stopImmediatePropagation');
+    ["ArrowLeft", "ArrowRight", "Home", "End"].forEach((key: string) => {
+      it(`should dispatch a keyboard event and expect "stopImmediatePropagation()" to have been called when using ${key} key`, () => {
+        const event = new (window.window as any).KeyboardEvent('keydown', { key, bubbles: true, cancelable: true });
+        const spyEvent = vi.spyOn(event, 'stopImmediatePropagation');
 
-      editor = new AutocompleterEditor(editorArguments);
-      const editorElm = divContainer.querySelector('input.editor-gender') as HTMLInputElement;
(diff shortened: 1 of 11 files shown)


zewa666:
oh and lol, I've just realized that my previous pr had sever naming missuses due to lack of typings. so the first commit fixes that

ghiscoding:
> oh and lol, I've just realized that my previous pr had sever naming missuses due to lack of typings. so the first commit fixes that

ah yeah, I also realized some time ago that any commits with a conventional keyword will add a new changelog row. I made that mistake couple times where I have multiple commits with multiple `fix: ` prefix and was wondering why I had so many changelogs for the same PR 😆

packages/common/src/editors/__tests__/autocompleterEditor.spec.ts
```diff
@@ -196,30 +196,19 @@ describe('AutocompleterEditor', () => {
       expect(editor.getValue()).toBe('male');
     });
 
-    it('should dispatch a keyboard event and expect "stopImmediatePropagation()" to have been called when using Left Arrow key', () => {
-      const event = new (window.window as any).KeyboardEvent('keydown', { key: 'ArrowLeft', bubbles: true, cancelable: true });
-      const spyEvent = vi.spyOn(event, 'stopImmediatePropagation');
+    ["ArrowLeft", "ArrowRight", "Home", "End"].forEach((key: string) => {
```
ghiscoding, line 199: just another possible suggestion, I almost never used this and it was in fact added by another user but did you know about the [`describe.each`](https://vitest.dev/api/#describe-each)? It's another way of writing the same kind of test looping, see this one for example

https://github.com/ghiscoding/slickgrid-universal/blob/265a04f2a266fc6f9a54d5cd9b994e2039b6d886/packages/odata/src/services/__tests__/grid-odata.service.spec.ts#L1647-L1664

zewa666: I think we talked about that one already. I still fail to see a clear benefit to the simple to read and understand [].forEach but if you prefer the describe.each syntax I can change that

ghiscoding: ahh I might have forgot, I'm fine with your approach, I just wanted to mention it in case you didn't know but it looks like you already know about it, so all good :)

PR #1704, opened 2024-10-04
fix: reposition column picker to always be visible

roberbnd:
**What**
Add value to property "left" if we haven't space available in the viewport
default is 10 to maintain the same behavior if it is not necessary

**Why**
when you click on the edge of the grid the columns options is not totally visible

### DIFF
--- a/packages/common/src/extensions/slickColumnPicker.ts
+++ b/packages/common/src/extensions/slickColumnPicker.ts
@@ -221,8 +221,15 @@ export class SlickColumnPicker {
   protected repositionMenu(event: DOMMouseOrTouchEvent<HTMLDivElement> | SlickEventData): void {
     const targetEvent: MouseEvent | Touch = (event as TouchEvent)?.touches?.[0] ?? event;
     if (this._menuElm) {
+      const availableWidth = window.innerWidth - (this._menuElm.getBoundingClientRect().right + targetEvent.pageX)
+      let leftPosition = 10
+
+      if (availableWidth < 0) {
+        leftPosition = leftPosition + 60 + Math.abs(availableWidth)
+      }
+
       this._menuElm.style.top = `${targetEvent.pageY - 10}px`;
-      this._menuElm.style.left = `${targetEvent.pageX - 10}px`;
+      this._menuElm.style.left = `${targetEvent.pageX - leftPosition}px`;
       this._menuElm.style.minHeight = findWidthOrDefault(this.addonOptions.minHeight, '');
       this._menuElm.style.maxHeight = findWidthOrDefault(this.addonOptions.maxHeight, `${window.innerHeight - targetEvent.clientY}px`);
       this._menuElm.style.display = 'block';


packages/common/src/extensions/slickColumnPicker.ts
```diff
@@ -221,8 +221,15 @@ export class SlickColumnPicker {
   protected repositionMenu(event: DOMMouseOrTouchEvent<HTMLDivElement> | SlickEventData): void {
     const targetEvent: MouseEvent | Touch = (event as TouchEvent)?.touches?.[0] ?? event;
     if (this._menuElm) {
+      const availableWidth = window.innerWidth - (this._menuElm.getBoundingClientRect().right + targetEvent.pageX)
+      let leftPosition = 10
+
+      if (availableWidth < 0) {
+        leftPosition = leftPosition + 60 + Math.abs(availableWidth)
```
ghiscoding, line 228: where does the 60 come from? it's not very clear what this calculation does. It's also not covered by unit test, it would be need to be covered with a new test. I also just don't understand what this does, please provide either a print screen or animated gif of what this fixes

ghiscoding:
@roberbnd please provide more details and necessary unit test (I can help if necessary). I would like more info on where you took the number `60` in your calculation or is that a random number? We shouldn't have fixed numbers in code, it's hard to find and debug (it could be moved to a constant on top of the file or as a new grid menu option). So if you could please provide more details, print screen or animated gif would also be useful to understand the problem. Thanks

roberbnd:
@ghiscoding I want to avoid this:
image
as you can see the table options "columns" is in the edge.

60 is a "random" number after I was testing how much I need to move the ` <div> ` to the left

about testing I didn't know it was necessary for this PR, I am working on that part but I dont get it what is the command only one file execute packages/common/src/extensions/__tests__/slickColumnPicker.spec.ts
I already did `pnpm install`
image

ghiscoding:
I thought this was for the Grid Menu (which already has an auto-positioning) but I just realized that it's for the Column Picker which doesn't have any auto-positioning. However your code seems more like a patch for your use case, I think what could be better is to actually add an auto-positioning the same way that I've done for the Header Menu (which has an left/right position depending on which side has more space available), so I think we should go for that approach instead. If you take a look at the Header Menu, here's the code that does this left/right positioning 

https://github.com/ghiscoding/slickgrid-universal/blob/6491d51ad146b7e6028fe020686cda76f455a000/packages/common/src/extensions/slickHeaderMenu.ts#L168-L195

We could reuse this code but since I also like to stay DRY (don't repeat yourself), I think we should move this into a separate function that could be called by both (probably in `extensionCommonUtils.ts`). However this might feel like a lot of work, so I could take it over if that's the case... or you could still contribute, up to you. 

and here's how to add breakpoint unit tests, it probably only works if you use VSCode though

Code_-_Insiders_FO5EYgWs1e

ghiscoding:
@roberbnd I got it working as above with just couple lines, if you want to contribute, you could apply these changes 

```diff
repositionMenu() { 
  // auto-positioned menu left/right by available position
+ const gridPos = this.grid.getGridPosition();
+ const menuWidth = this._menuElm.clientWidth || 0;
+ let menuOffsetLeft = targetEvent.pageX || 0;
+  if (gridPos?.width && (menuOffsetLeft + menuWidth >= gridPos.width)) {
+    menuOffsetLeft = menuOffsetLeft - menuWidth;
+  }

  this._menuElm.style.top = `${targetEvent.pageY - 10}px`;
- this._menuElm.style.left = `${targetEvent.pageX - 10}px`;
+ this._menuElm.style.left = `${menuOffsetLeft}px`;
  // ...
}
```

and then add this unit test to cover the change

```diff
const gridStub = {
  applyHtmlCode: (elm, val) => elm.innerHTML = val || '',
+  getGridPosition: vi.fn(),
  getColumnIndex: vi.fn(),

// go to line 365, add unit test just before "onColumnsReordered event"

+    it('should reposition menu to the left when no available space on the right', () => {
+      vi.spyOn(gridStub, 'getGridPosition').mockReturnValue({ left: 50, top: 0, right: 0, bottom: 200, height: 22, width: 300, visible: true });
+      vi.spyOn(gridStub, 'getColumnIndex').mockReturnValue(undefined as any).mockReturnValue(1);
+
+      control.init();
+
+      const groupElm = createDomElement('div', { className: 'slick-column-name' });
+      gridStub.onPreHeaderContextMenu.notify({ node: groupElm, grid: gridStub }, { ...new SlickEventData(), preventDefault: vi.fn(), target: groupElm, pageX: 305 } as any, gridStub);
+      vi.spyOn(control, 'createPickerMenu').mockImplementation(() => {
+        if (control.menuElement) {
+          Object.defineProperty(control.menuElement, 'clientWidth', { writable: true, value: 122 });
+          return control.menuElement;
+        }
+        return document.createElement('div');
+      });
+      gridStub.onPreHeaderContextMenu.notify({ node: groupElm, grid: gridStub }, { ...new SlickEventData(), preventDefault: vi.fn(), target: groupElm, pageX: 305 } as any, gridStub);
+      Object.defineProperty(control.menuElement, 'clientWidth', { writable: true, value: 122 });
+      expect(control.menuElement?.style.left).toBe('183px');
+    });

    describe('onColumnsReordered event', () => {
      it('should reorder some columns', () => {
}
```

brave_ARmd1gxHoq

ghiscoding:
Since I didn't hear anything back, I went ahead and modified your PR and superseded it as a new PR #1707 
Also since I checked out your PR, you will still be added as a contributor. 

Thanks for opening this.

PR #1772, opened 2024-12-11
feat: dynamically create grid from imported CSV data

ghiscoding:


ghiscoding:
@zewa666 so I was looking for a new feature so that I could push a new `minor` release for Slickgrid-Vue and thought of creating this new example. It's most probably a common scenario and perhaps that might help you with your recent comment about creating a grid from a dynamic form!? What do you think? 😉 

Also a reminder, all PRs now include the Stackblitz link, so you can even give it a try live if you want :)

zewa666:
we do that for our OData provided Grid/ColumnConfigs so thats a super valuable example. our current use case though is to somewhat replicate the composite editor form but instead with ngx-formly in order to support rendering various grid definitions. e.g the grid shows the overview but dblckick on a row will open a modal with 1..n related tables based on forms.

PR #1788, opened 2024-12-30
fix(core): add missing key shortcuts for navigation & cell selections

ghiscoding:
- add a few missing key shortcuts to navigate in the grid
  - Ctrl+Home will now go to top left (coord 0, 0)
  - Ctrl+End will now go to bottom right of the grid
- add a few missing key shortcuts for cell selection
  - Shift+Ctrl+ArrowLeft, now same as Shift+Home, to select cells from current cell until starting index on same row
  - Shift+Ctrl+ArrowRight, now same as Shift+End, to select cells from current cell until end index on same row
  - Shift+Ctrl+ArrowUp to select cells from current cell up to first row on same column
  - Shift+Ctrl+ArrowDown to select cells from current cell down to last row on same column
  - Ctrl+a (or A) to select all cells in the grid


ghiscoding:
cc @zewa666

PR #1798, opened 2025-01-18
feat: add final rowspan implementation

ghiscoding:
So I've been working over a month on this, initially based on the @GerHobbelt's fork from his initial rowspan [commit](https://github.com/GerHobbelt/SlickGrid/commit/397eab80ed43a7048b8c6554ffb03a4d7b74a2e8) and after a few bug fixes, I was surprised to see it working for the most past **but** it had particular negative side effects that made me think if I could improve it further. I had a few problems with that first approach and they are mostly all related to each other:
1. it required to loop through the entire dataset to build a 2D array of **every cells** of that entire dataset with a 2D array that keeps `[[row, cell, colspan, rowspan]]`
   - it had a few nested `for` loops and I got concerned when I saw `for` loops 3 level down (scary😨)
3. this 2D array mapping was being built for **every grid** regardless if you had rowspan or not
4. and finally it required to keep this large 2D array in memory

So I decided to see if I could do it in a different way that would require less footprint (no 2D array) and less intrusive (it shouldn't impact regular grids perf at all) and I came up with this new approach. Instead of creating this large 2D array, what if we keep only the start/end indexes of each `rowspan` but instead of rows, we keep them by column indexes (basically the inverse of item metadata because that is row based)... and that is what I went with, the perf is quite good (about twice as fast as the first approach). Note that this 2nd approach still does require to loop through all dataset rows to extract every existing `rowspan` from all item metadata. So let's say that on the 1st column from the 2nd row, we have a `rowspan` of 5 then the mapping that we'll keep internally would be `{ 0: '1:6' }`... great but what if on top of the `rowspan`, we also have a `colspan` of 2 on the same cell? Then our mapping becomes `{ 0: '1:6', 1: '1:6' }` (basically duplicating the same start/end on the 2nd column). Also another concept is that if any cell has a `rowspan` child that becomes out of the viewport, we need to keep its parent rendered because that is the cell holding the height of it, however if eventually the intersect becomes out of the viewport then we'll remove it from our cache (so it might keep certain data rows a little longer in the DOM but that is required for the rowspan to work properly, i.e. our example has a rowspan that spreads across the grid until the bottom, so that parent row is basically never being removed from the cache because she always need to be displayed)

So why is this better than the first approach? Well this approach takes a lot less space in memory and if our rowspan is very large, let say 100 rows, then the mapping is nearly the same and we just need to update our ending for example `{ 0: '5:105' }` (5th row, spanning for 100 rows, so 100+5=>105 is the ending)... simple! The other bonus is that with this new approach, I do have to keep a mapping object but it's relatively small and the bonus is that I've put this under a grid option flag `enableRowSpan` so that regular grids have 0 perf impact as opposed to the previous approach that impacted **all** grids (just to be clear though, the perf impact wasn't that huge, for 500K rows it was around 300-400ms to build mapping for 1st approach and around 100-200ms for new approach and also note that it assuming only about 10 `rowspan` defined which is what the example demo has, so for grids with a lot more `rowspan` then expect the perf time to get slightly higher).

image

image


ghiscoding:
cc @zewa666 there it is RowSpan!!! Please do note that this comes with multiple constraints
- side effects will not keep anything in sync since metadata are based on grid row indexes... but that was also true for `colspan` as well
  - for example: Filtering/Sorting/Paging/ColumnReorder
  - in other words, the lib cannot know or guess how the user (you) will want to keep rowspan, so the user has implement his own logic 
- also Excel Export is currently broken, I will probably fix it after a new release just to push it more quickly

zewa666:
wow ... I didnt expect that to land so soon. I will definitely have a closer look at this next week since I'm working again on the grid. thx a lot for your efforts

ghiscoding:
yeah I was bored at the beginning of my Xmas holidays and gave a try to replicate the GerHobbelt's initial commit and within couple of hours I had something working as a POC, a few more days and most bugs were fixed but I thought I could improve perf and... that's how you get hooked back into the game, it never stops! haha, anyway that is pretty much the last big feature that was missing. :)

zewa666:
yeah, I think I heard that one before 😁

PR #1820, opened 2025-01-25
fix(resizer): add `autoResize.autoHeight` to resize by dataset length

ghiscoding:
The auto-resize is great but it had a flaw that always bothered me but wasn't sure on how to fix it. The problem is that it always resized the grid with all available viewport space regardless of the dataset and in some cases when the dataset was small, it showed a full size grid in the UI with in some cases a rather large empty whitespace area at the bottom of the grid. This PR fixes it once and for all. We always had the grid option `autoHeight` which was good when the user knows its dataset is small, but most of the time we don't know dataset length at all, it could be small today but large next month, so we can't use `autoHeight` grid option for that reason.... this PR adds the grid option `autoResize.autoHeight` (yes it's the same flag name but it's not under the same property), it's enabled by default but user could disable it if he wants and it doesn't do anything unless `enableAutoResize` is enabled which is why it's under that group. Another thing that is useful to know is that it's re-validated every time the dataset count changes (by either new data assignment OR data filtering). Lastly we should also note that the we have a `autoResize.minHeight` to avoid resizing too small (it was always in place but I increased it to 250px)

image

#### after fix

brave_wvRzO9334t

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
 # never EVER use CRLF, Windows
-* text eol=lf
+* text=auto eol=lf
--- a/demos/vue/src/components/Example28.vue
+++ b/demos/vue/src/components/Example28.vue
@@ -154,6 +154,7 @@ function defineGrid() {
 
   gridOptions.value = {
     autoResize: {
+      autoHeight: false,
       container: '#demo-container',
       rightPadding: 10,
     },
--- a/demos/vue/src/components/Example28.vue
+++ b/demos/vue/src/components/Example28.vue
@@ -310,7 +310,7 @@ function deleteFile() {
     if (songIdx >= 0) {
       popFolderItem.files.splice(songIdx, 1);
       lastInsertedPopSongId.value = undefined;
-      // isRemoveLastInsertedPopSongDisabled.value = true;
+      isRemoveLastInsertedPopSongDisabled.value = true;
 
       // overwrite hierarchical dataset which will also trigger a grid sort and rendering
       datasetHierarchical.value = tmpDatasetHierarchical;
--- a/demos/vue/src/components/Example28.vue
+++ b/demos/vue/src/components/Example28.vue
@@ -309,7 +309,7 @@ function deleteFile() {
     if (songIdx >= 0) {
       popFolderItem.files.splice(songIdx, 1);
       lastInsertedPopSongId.value = undefined;
-      isRemoveLastInsertedPopSongDisabled.value = true;
+      // isRemoveLastInsertedPopSongDisabled.value = true;
 
       // overwrite hierarchical dataset which will also trigger a grid sort and rendering
       datasetHierarchical.value = tmpDatasetHierarchical;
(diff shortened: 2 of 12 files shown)


ghiscoding:
cc @zewa666

zewa666:
thats great. I have a reported bug with autoresize and using frozen columns which I need to verify but I will check out this PR together with it.

zewa666:
@ghiscoding oops sry, now I just realized that this is actually a breaking change.

In our app we're expecting the grid to take up all the height by default. I guess same would apply to the [trading example](https://ghiscoding.github.io/Angular-Slickgrid/#/trading), because if you filter there lets say by Quantity 12, the grid starts jumping reaaaaalllly ugly.

Anyways, `autoResize.autoHeight: false` is the workaround and its documented. but nevertheless this was a breaking change

ghiscoding:
I really don't consider this a breaking change but improvement could probably be made, personally I rarely deal with live data like the real time trading, so I didn't test much with that type of grid but isn't `autoResize.autoHeight: false` enough anyway? When you said breaking change, did u mean that this flag should have been set to true by default (maybe until next major) is that your intention? 

Talking about breaking change, recent Vitest version broke my tests, which I fixed last night, and they decided to go with a patch version (maybe because v3 was released not that long ago), even though what they've done is a real breaking change on my side (all my tests with `setTimeout` and `queueMicrotask` broke). So sometime it seems that people have different view on breaking changes 😆 

> Vitest reintroduced a breaking change that they had originally put in v3.0 but reverted and now reintroduced which broke my unit tests. The breaking change is in this Vitest [commit](https://github.com/vitest-dev/vitest/commit/167a98d77c8095288fcf193f8dfff0dd5cd44fa1) which is that ['setTimeout', 'setInterval', 'queueMicrotask'] are no longer mocked and must be added manually in our Vitest config. This PR fixes this so that I can keep up again with latest Vitest version

ghiscoding:
@zewa666 hmm I really don't see, and can't replicate, any of that jumping in any of the demos but if you still think false should be the default then feel free to submit a fix PR

zewa666:
Here's a gif of what I meant when performing filtering. Notice how the footer starts to jump up/down as you type.

2025-03-06_10h51_37

Since this is a new behavior in contrast to before it technically is a breaking change. Does it matter? Not really, hence I'd simply ignore it as the workaround explained in the release notes clearly points out how to revert to the old behavior.

zewa666:
I'm totally with you in that this kind of is the more standard way compared to a fixed height+ empty space. just wanted to note that technically its nevertheless a breaking change as the default behavior changed within a minor version.

I totally would leave things as they are though except set the autoHeight value to false on the live feed example bc for these type of apps a fixed height seems the more appropriate way to go as that wild jumping is nerve-wracking without you manually doing anything 😅

ghiscoding:
@zewa666 ohhhh after re-reading and correctly trying your first suggestion of filtering by Qty 12, now I totally understand what you meant! The grid height keeps changing height which is the annoying part. Damn, sorry for being late on the understanding 😅 

So yes indeed, it does make sense to turn it off on that demo. Just looking at the animated gif below makes me dizzy lol

msedge_aHbKy3MMVs

zewa666:
haha yeah thats whar I meant. anyways all good and once that sample is fixed all should be fine imo

PR #1839, opened 2025-02-07
fix: add an explicit check to bound element

### LINKED ISSUE #1835: Using Dropzone with Slickgrid breaks the pagination pageSize select
### Describe the bug

Hi there, I ran into a weird bug that was quite a challenge to debug. I've been adding Dropzone (via https://www.npmjs.com/package/ngx-dropzone-wrapper) in order to create a new file upload editor and it worked nicely so far. Today got a call that our pagination pageSize select is broken and changing values does not perform any action.

After loooong debugging I figured out that the select lacks the change event subscription. Diving deeper into the codebase I realized that `BindingEventService->bind` there is something like this happening:

```ts
    if ((elementOrElements as NodeListOf<HTMLElement>)?.forEach) {
      // multiple elements to bind to
      ...
    } else if (elementOrElements) {
      // single elements to bind to
      ...
    }
```
which looks totally fine ... except that Dropzone seems to have used some Webpack magic to add a monkey patch that will patch the native HTMLSelect to add a forEach. And guess what it returns? Right, it's option elements 🤦 

So thats why my select ended up without a change but its options did :)

Now to be frank I don't really see this as an issue of Slickgrid. My question is whether we should try to handle this in slickgrids codebase or if this is to be fixed in the upstream dependency?

Thankfully one of your previous releases introduced customPaginationComponent and I was able to simply extend the original one and reverse-patch the issue. so there is a workaround

```ts
@Component({
  template: ""
})
export class ModifiedPaginationComponent extends SlickPaginationComponent {
  override addEventListeners(): void {
    ...
    if (
      typeof (this._itemPerPageElm as ForEachPatchedSelect).forEach ===
      "function"
    ) {
      (this._itemPerPageElm as ForEachPatchedSelect).forEach = (
        cb: (element: HTMLElement) => void
      ) => cb(this._itemPerPageElm);
    }
    this._bindingEventService.bind(
      this._itemPerPageElm,
      "change",
      this.updateItemsPerPage.bind(this)
    );
  }
}
```

### Reproduction

* Add Slickgrid in an angular project
* Add ngx-dropzone-wrapper
* turn on pagination support
* notice the select of the pageSize not working due to missing change event binding

Angular

### Environment Info

```shell
```

### Validations

zewa666:
checks that the bound element is not an input with a monkey patched forEach. See https://github.com/ghiscoding/Angular-Slickgrid/discussions/1446#discussioncomment-10194750

closes #1835

### DIFF
--- a/packages/binding/src/__tests__/bindingEvent.service.spec.ts
+++ b/packages/binding/src/__tests__/bindingEvent.service.spec.ts
@@ -32,6 +32,21 @@ describe('BindingEvent Service', () => {
     expect(addEventSpy).toHaveBeenCalledWith('click', mockCallback, undefined);
   });
 
+  it('should not use forEach if the element is a single HTMLElement but with a monkey patched forEach', () => {
+    const elm = document.createElement('input');
+    div.appendChild(elm);
+    const forEachSpy = vi.fn();
+    (elm as any).forEach = forEachSpy;
+    const mockCallback = vi.fn();
+    const addEventSpy = vi.spyOn(elm, 'addEventListener');
+
+    service.bind(elm, 'click', mockCallback);
+
+    expect(service.boundedEvents.length).toBe(1);
+    expect(addEventSpy).toHaveBeenCalledWith('click', mockCallback, undefined);
+    expect(forEachSpy).not.toHaveBeenCalled();
+  });
+
   it('should be able to bind and unbindByEventName an event', () => {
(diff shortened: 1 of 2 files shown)


ghiscoding:
Great thanks :)

zewa666:
no probs. will check up on the upstream to see why they're actually using that

PR #1843, opened 2025-02-12
fix: some browsers show grid viewport too wide by 1px

ghiscoding:
- so in some rare cases, some browsers incorrectly show the horizontal scrollbar for the grid when it shouldn't. The weird thing is that the scrollbar is only showing for 1px. We can fix this by simply looking at the available viewport inner width (that is the viewport minus the scrollbar width), and if the width is the same then there's no need to apply this width to the grid pane (prior to this PR, the calculated inner width was always applied to the grid pane, but now it won't be applied when it doesn't need it.... or in other words, only apply the grid pane width when it is larger than the available viewport width)
- in my case it happened with Ms Edge, though it probably could be any browsers

#### before fix, some browsers showed the scrollbar

image

#### with the fix in the same browser

image

### DIFF
--- a/packages/common/src/core/slickGrid.ts
+++ b/packages/common/src/core/slickGrid.ts
@@ -1330,7 +1330,7 @@ export class SlickGrid<TData = any, C extends Column<TData> = Column<TData>, O e
 
   /** Get the grid canvas width */
   getCanvasWidth(): number {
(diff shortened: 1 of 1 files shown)


ghiscoding:
hmm so thanks to all our Cypress tests, I can conclude that this is not a good fix. I'll have to investigate some more

ghiscoding:
closing since this

PR #1849, opened 2025-02-25
feat: allow overriding the position of the textarea savebutton

zewa666:
as windows users, my customers argue the save/cancel button order is super important ... *sigh*

### DIFF
--- a/packages/common/src/editors/__tests__/longTextEditor.spec.ts
+++ b/packages/common/src/editors/__tests__/longTextEditor.spec.ts
@@ -177,6 +177,14 @@ describe('LongTextEditor', () => {
       expect(editor.editorDomElement.rows).toBe(8);
     });
 
+    it('should initialize the editor with cols & rows define in user editor options', () => {
+      mockColumn.editor!.editorOptions = { cols: 9, rows: 8 } as LongTextEditorOption;
+      editor = new LongTextEditor(editorArguments);
+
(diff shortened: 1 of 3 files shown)


ghiscoding:
wow ok, I was surprised to see that I didn't add this already, I'm 100% sure that I had thought of that but maybe just in my dreams lol and after looking for positioning it looks like I only have `dropSide` for ContextMenu and GridMenu.... oh well, better late than never 😆

ghiscoding:
@zewa666  Side note and totally out of context, have you ever used horizontal stacked bar? Some new user created [Streamlit-Slickgrid](https://github.com/[streamlit/streamlit-slickgrid](https://github.com/streamlit/streamlit-slickgrid?tab=readme-ov-file)?tab=readme-ov-file) for python it seems and they use Slickgrid-Universal and while taking a look I found that stacked bar in their [demo](https://slickgrid.streamlit.app/), that might be interesting for some.

On the other hand, I'm making progress with Row Detail inner grids, I'm adding tests now and there might be 1 little bug to look into. I hope to finish everything by the weekend and push a release that will include your PR as well. Cheers

image

zewa666:
well one of the next features is going to have client side expressions sent from the backend in order to render a traffic light like chart. so bar charts might be close, although I'd prefer an angular component formatter with apache echarts

PR #1880, opened 2025-03-26
fix: shift + a should not select all cells upwards

zewa666:
Hi there @ghiscoding,

it seems we introduced a bug back in the December release where SHIFT + A inside an editor would perform a cell selection instead of writing the capital A letter.

You can try it out on https://ghiscoding.github.io/Angular-Slickgrid/#/editor, editing a Country of Origin Name cell and trying to type SHIFT + A.

### DIFF
--- a/packages/common/src/extensions/slickCellSelectionModel.ts
+++ b/packages/common/src/extensions/slickCellSelectionModel.ts
@@ -186,8 +186,18 @@ export class SlickCellSelectionModel implements SelectionModel {
     this.setSelectedRanges([args.range]);
   }
 
-  protected isKeyAllowed(key: string): boolean {
-    return ['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown', 'PageDown', 'PageUp', 'Home', 'End', 'a', 'A'].some((k) => k === key);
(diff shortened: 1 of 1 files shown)


ghiscoding:
oh wow good catch, thanks for that, lgtm

zewa666:
I just wonder why the issue was only reproducible on the angular samples and not the slickgrid ones

ghiscoding:
@zewa666 there isn't many examples that have the copy plugin enabled, [Example 19](https://ghiscoding.github.io/slickgrid-universal/#/example19) has it enabled and it does reproduce what you mentioned.

zewa666:
ahh alrighty, yeah missed the point that its related to the copy plugin

ghiscoding:
@zewa666  now published and this is potentially the last release before the next major that I'm about to start working on (will probably create a `next` branch again)

zewa666:
alrighty. is that new major already going to be the shift of the angular wrapper to the mono repo?

ghiscoding:
> alrighty. is that new major already going to be the shift of the angular wrapper to the mono repo?

indeed that's the plan, and actually it should all be listed in #1866 and if you have any other ideas then feel free to comment it in the opened discussion. Cheers

PR #1890, opened 2025-04-02
Fix: error in CSS

wboevink:
importing the stylesheet in angular with
`@import '@slickgrid-universal/common/dist/styles/css/slickgrid-theme-default.css';`

throws an error
`✘ [ERROR] Missing closing } at .slickgrid-container .search-filter .operator.input-group-addon select [plugin angular-css]`

the `_variables.scss` included `!important` which caused the error, moved `!important` to the `.scss` file where the variable is used.

### DIFF
--- a/packages/common/src/styles/_variables.scss
+++ b/packages/common/src/styles/_variables.scss
@@ -641,7 +641,7 @@ $slick-editor-modal-validation-summary-font-style:          italic !default;
 /* Compound Filters */
 $slick-compound-filter-bgcolor:                             #ffffff !default;
 $slick-compound-filter-operator-select-font-family:         "Cascadia Mono", Consolas, "Lucida Console" !default; // use a monospace font so the operator descriptions are all aligned
-$slick-compound-filter-operator-select-font-size:           14px !important !default;
+$slick-compound-filter-operator-select-font-size:           14px !default;
 $slick-compound-filter-operator-select-border:              1px solid #{color.adjust($slick-primary-color, $lightness: 15%)} !default;
 $slick-compound-filter-operator-select-width:               25px !default;
 $slick-compound-filter-operator-border-radius:              4px 0 0 4px !default;
--- a/packages/common/src/styles/slick-plugins.scss
+++ b/packages/common/src/styles/slick-plugins.scss
@@ -801,7 +801,7 @@ li.hidden {
 
     select {
       font-family: var(--slick-compound-filter-operator-select-font-family, v.$slick-compound-filter-operator-select-font-family);
-      font-size: var(--slick-compound-filter-operator-select-font-size, v.$slick-compound-filter-operator-select-font-size);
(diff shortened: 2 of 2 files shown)


ghiscoding:
@wboevink so does that mean that we can't use `!important` inside a CSS Variable? Because if I look on unpkg [here](https://app.unpkg.com/@slickgrid-universal/common@5.13.3/files/dist/styles/css/slickgrid-theme-default.css), the code has this `font-size:var(--slick-compound-filter-operator-select-font-size,14px !important);` which puts the `!important` in the fallback section (2nd argument). But personally, I use SASS and I never override this one myself and I've never seen this error, so I'm wondering why you're seeing the error yourself. 

However, the suggestion you provide is not ideal because with this code, we can no longer override the `!important` since it will always be in the code (because you moved it outside the variable). It might be better to see if we can get rid of the important completely, I'm assuming this problem probably showed up in Bootstrap, which framework are you using then? Side note, I know that I have other `!important` in the same file, getting rid of them would be ideal but sometime it's hard to do.

ghiscoding:
@wboevink so it turns out that we can simply take a look at the demo website to find out that `!important` is not needed at all, because as mentioned above, I didn't know that `!important` wasn't valid in a CSS variable.... so the SASS compiler simply removes the `!important` and the font-size is override it anyway, so it was totally unneeded and could be removed

image

So could you please update your PR and remove the `!important` and also remove the other one for padding which is also unnecessary and works well without it

https://github.com/ghiscoding/slickgrid-universal/blob/12baea69d658c911305800574a662a7a9ea0ad6b/packages/common/src/styles/slick-plugins.scss#L792

If you could update your PR by removing these 2 `!important`, I could go ahead and merge it. It would be 2 less important flag to take care of in the future (better take care of them than later)

wboevink:
> and when you don't use SASS then the error shows up!?

I just used the compiled `.css` files. All of them had the same error. This is the part where the error occurs:

```css
.slickgrid-container .search-filter .operator.input-group-addon select{
    border:var(--slick-compound-filter-operator-select-border,1px solid #599bfe);
    font-family:var(--slick-compound-filter-operator-select-font-family,"Cascadia Mono",Consolas,"Lucida Console");
    font-size:var(--slick-compound-filter-operator-select-font-size,14px !important); 
    width:var(--slick-compound-filter-operator-select-width,25px)
}
```

The font size is just ignored as the fallback value of 'var()' is invalid.

wboevink:
> If you could update your PR by removing these 2 `!important` flags, I could go ahead and merge it. It would be 2 less flags to take care of in the future (better take care of them now than later)
> 
> So in the end, I would like you to remove the `!important` on both variables as mentioned above and I'll get rid of this variable in the next major version (which I started working on).

Thanks for the comprehensive answers! I just committed the requested changes.

ghiscoding:
Thanks :)

ghiscoding:
@wboevink this is now released, sorry for the delay, I was waiting to see if anything else came up and I also started working on the next major release. Thanks again for the contribution

PR #1891, opened 2025-04-04
chore: move Angular-Slickgrid into Slickgrid-Universal monorepo

ghiscoding:
- at first I tried to use the same 2 folders as the Vue wrapper (1. `demo/angular`, 2. `frameworks/angular-slickgrid`) but `ng-packagr` is a little slow to build the project (slower than Vite for sure) and also the fact that Angular requires to publish the library package `dist/` folder instead of typically the package root folder is causing other problems in the monorepo setup... So in the end, I moved all examples back into the `frameworks/angular-slickgrid/app/examples` folder, this way it remains fast & the library publish from dist folder isn't a problem anymore for exposing the library (in other words, I'll be using the same folder structure that I had in the previous Angular-Slickgrid repo since that seems to be the most optimal structure)


ghiscoding:
cc @zewa666

PR #1916, opened 2025-04-12
chore: bump TS target to ES2022

ghiscoding:


ghiscoding:
seems to regress with ES2022, let's stick with ES2021 then

PR #1928, opened 2025-04-24
feat: allow providing custom date format via base Date Formatter

ghiscoding:
Allow the user to provide a custom date format via a new basic date Formatter, the user must provide the format through `params.dateFormat` in his column `params`. Note that the input date `type` must still be provided the same as before, this PR is just an easier way to show different output format in the UI

For example

```ts
this.columns = [
  {
    id: 'start', name: 'Start', field: 'start',
    type: FieldType.date,
    formatter: Formatters.date, // base date formatter which requires "params.dateFormat"
    params: {
      dateFormat: 'MMM DD, YYYY' // ie: Dec 30, 2000
    },
  }
];
```

### Demo

with the code above, we have this working demo

image

### DIFF
--- a/packages/common/src/filters/dateFilter.ts
+++ b/packages/common/src/filters/dateFilter.ts
@@ -7,8 +7,8 @@ import type { IOptions } from 'vanilla-calendar-pro/types';
 import { FieldType, OperatorType, type OperatorString, type SearchTerm } from '../enums/index.js';
 import type { Column, ColumnFilter, Filter, FilterArguments, FilterCallback, GridOption, OperatorDetail } from '../interfaces/index.js';
 import { applyOperatorAltTextWhenExists, buildSelectOperator, compoundOperatorNumeric } from './filterUtilities.js';
-import { formatDateByFieldType, mapTempoDateFormatWithFieldType } from '../services/dateUtils.js';
-import { mapOperatorToShorthandDesignation } from '../services/utilities.js';
+import { formatDateByFieldType, mapTempoDateFormatWithFieldType, tryParseDate } from '../services/dateUtils.js';
+import { isColumnDateType, mapOperatorToShorthandDesignation } from '../services/utilities.js';
 import type { TranslaterService } from '../services/translater.service.js';
 import type { SlickGrid } from '../core/slickGrid.js';
 import { resetDatePicker, setPickerDates } from '../commonEditorFilter/commonEditorFilterUtils.js';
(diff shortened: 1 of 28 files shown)


ghiscoding:
@zewa666 so that's what I came up with to address custom date formats, I had to revisit Sorting & Filtering too, because without them working would have been confusing to the user and it's interesting to have them work too even if it increases the library size a tad more

packages/common/src/filters/dateFilter.ts
```diff
@@ -417,6 +423,23 @@ export class DateFilter implements Filter {
...
 
+  protected formatDate(
+    inputDate: Date | string,
+    column: Column,
+    inputFieldType: (typeof FieldType)[keyof typeof FieldType] | undefined,
+    outputFieldType: (typeof FieldType)[keyof typeof FieldType]
+  ): string {
+    let dateStr = '';
+    const params = column.params || {};
+    const paramOutputFormat = params?.outputFormat ?? params.format;
```
ghiscoding, line 434:
```suggestion
    const paramOutputFormat = params.outputFormat ?? params.format;
```

packages/common/src/services/collection.service.ts
```diff
@@ -234,8 +232,10 @@ export class CollectionService<T = any> {
   // -------------------
 
   protected getParseDateInfo(col: Column, preParseDateColumns: boolean | string): ParsingDateDetails | void {
+    const params = col.params ?? {};
+    const paramDateFormat = params.inputFormat ?? params.format;
     const fieldType = col.type || FieldType.string;
-    const dateFormat = mapTempoDateFormatWithFieldType(fieldType);
+    const dateFormat = paramDateFormat ?? mapTempoDateFormatWithFieldType(fieldType);
```
ghiscoding, line 238:
```suggestion
    const dateFormat = paramDateFormat || mapTempoDateFormatWithFieldType(fieldType);
```

packages/common/src/sortComparers/sortUtilities.ts
```diff
@@ -15,8 +15,17 @@ export function sortByFieldType(
...
 
   if (isColumnDateType(fieldType)) {
-    // @ts-ignore
-    sortResult = getAssociatedDateSortComparer(fieldType).call(this, value1, value2, sortDirection, sortColumn, gridOptions);
+    // if the column is a date type, the user can optionally pass date format inside column params
+    // when that happen, we can call the date sort function directly without having to first find associated date format by column type
+    let paramDateFormat = '';
+    if (sortColumn && isColumnDateType(fieldType)) {
+      const params = sortColumn?.params || {};
+      paramDateFormat = params?.inputFormat ?? params.format;
```
ghiscoding, line 23:
```suggestion
      paramDateFormat = params.inputFormat ?? params.format;
```

zewa666:
so how would that behave with a graphql/odata backend? would the outputDate control filter/sorting here as well?

ghiscoding:
Yeah what I had in mind was to make `inputFormat` equal `type` and `outputFormat` equal `outputType` and then I thought what about Filtering, Sorting and.... perhaps I was way over the top with this PR and I should scale it back...a lot! 

All I really wanted to do, was to provide an easy approach for the end user to display a custom format for dates without having to create custom formatter.

zewa666:
I can see the benefit for being able to quickly change the formats there without the need for custom formatter functions. sadly the reality is that more than often I find my anyways creating one due to sudden feature requests 😅

nevertheless this looks good except one thing I'd like to ask for how it behaved and that is copy&paste. enabling copy formatted value would respect this one as well I'd assume right?

ghiscoding:
> enabling copy formatted value would respect this one as well I'd assume right?

yes the logic is to check if there's a Formatter on the cell, if there is then run it to get formatted output and so that will behave the same as before

ghiscoding:
so here it is below with `exportWithFormatter` enabled and it works as intended. I also tried separately without the flag and as expected it just returns the original input cell value


Code_-_Insiders_jVsko6S0Jc

PR #1964, opened 2025-05-09
fix!: remove Arrow pointer from Custom Tooltip

ghiscoding:
- removing the Arrow pointer from the Custom Tooltip because it was often showing at the wrong place because of its fixed position, so I think it's better to just remove it completely and just keep tooltip auto-positioning and that should be more than sufficient.
- this also helps decreasing the build size by about 2kb (less code but also less CSS)

brave_TE0rYOIahi

### DIFF
--- a/demos/aurelia/src/examples/slickgrid/example33.scss
+++ b/demos/aurelia/src/examples/slickgrid/example33.scss
@@ -13,10 +13,6 @@ $button-border-color: #ababab;
 //   --slick-tooltip-border-color:             #252525;
 //   --slick-tooltip-border:                   2px solid #252525;
 //   --slick-tooltip-color:                    #ffffff;
-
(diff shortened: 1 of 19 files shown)


ghiscoding:
@zewa666 FYI, I'm removing the Tooltip Arrow as shown above, I think that's fine, right? Whenever I looked at the position of the arrow, it was often offset with the text because of its fixed position. I know there are libs to better follow the position and such, like popperJS, but the goal is really to keep my lib a bit light and no adding too many 3rd party deps

I'll keep the PR open and wait for your feedback before merging.

zewa666:
all good. I personally actually prefer this display style as it feels more like the native ones.

ghiscoding:
yeah it's a more modern look in some ways 😄 
thanks for the feedback

PR #1990, opened 2025-05-27
Support cell selection

wscherphof:
As discussed in #1987:
- Allow either the CellSelectionModel or the RowSelectionModel;
- Don't beak the selected cell ranges when using the CellSelectionModel;
- Provide an extra callback onDestroy, that always gets called when the editor is removed.

### DIFF
--- a/packages/common/src/interfaces/compositeEditorOpenDetailOption.interface.ts
+++ b/packages/common/src/interfaces/compositeEditorOpenDetailOption.interface.ts
@@ -104,6 +104,12 @@ export interface CompositeEditorOpenDetailOption {
    */
   onClose?: () => Promise<boolean>;
 
+  /**
+   * onDispose callback allows user to execute something when the modal is disposed (closed and removed from the DOM).
+   * This will be called regardless of whether there are changes done in the form or not.
+   */
+  onDispose?: () => void;
+
   /** onError callback allows user to override what the system does when an error (error message & type) is thrown, defaults to console.log */
   onError?: (error: OnErrorOption) => void;
 
--- a/packages/composite-editor-component/src/slick-composite-editor.component.ts
+++ b/packages/composite-editor-component/src/slick-composite-editor.component.ts
@@ -130,6 +130,9 @@ export class SlickCompositeEditorComponent implements ExternalResource {
     this._eventHandler.unsubscribeAll();
     this._bindEventService.unbindAll();
     this._formValues = null;
+    if (typeof this._options.onDispose === 'function') {
+      this._options.onDispose();
+    }
     this.disposeComponent();
   }
 


packages/common/src/interfaces/compositeEditorOpenDetailOption.interface.ts
```diff
@@ -104,6 +104,12 @@ export interface CompositeEditorOpenDetailOption {
    */
   onClose?: () => Promise<boolean>;
 
+  /**
+   * onDispose callback allows user to execute something when the modal is disposed (closed and removed from the DOM).
+   * This will be called regardless of whether there are changes done in the form or not.
+   */
+  onDispose?: () => void;
```
ghiscoding, line 111: ~why create a new event, or mainly why is `onClose` event not good enough? I would prefer to have less code... is it possible to update the `onClose` instead to make it more versatile?~

ahhhh nevermind, I just read the docs I wrote a while ago just above and yes now it makes sense but I wonder if maybe that `onClose` should be renamed to a more representative name in the future (that would be a breaking change though)

> For example, when the user opens the modal window and start changing a few inputs in the form but then decides to the form, this is when the `onClose` gets executed (and to be clear, it only gets executed when there's changes in the form and a close action is clicked, it won't execute when there's no changes).

package.json
```diff
@@ -94,7 +94,7 @@
     "node": "^20.17.0 || >=22.9.0",
     "pnpm": "10.x"
   },
-  "packageManager": "pnpm@10.10.0",
+  "packageManager": "pnpm@10.11.0+sha512.6540583f41cc5f628eb3d9773ecee802f4f9ef9923cc45b69890fb47991d4b092964694ec3a4f738a420c918a333062c8b925d312f42e4f0c263eb603551f977",
```
ghiscoding, line 97: this would have to be rolled back, I would prefer to keep pnpm minimum version to v10.10

docs/grid-functionalities/Composite-Editor-Modal.md
```diff
@@ -265,6 +265,16 @@ this.compositeEditorInstance?.openDetails({
...
 
+### onDispose
+The `onDispose` callback function notifies you that the editor is finished and being disposed. It gets called regardless of whether there are changes or not.
+```ts
+this.compositeEditorInstance?.openDetails({
+  headerTitle: 'Create Item',
+  modalType: 'create',
+  onDispose: () => console.log('editor removed'),
+});
+```
```
ghiscoding, line 276: that's good, though we also need to update all framework docs, so you should modify 5 docs in total (4 frameworks + 1 vanilla). You can easily find all other docs by doing a search across the entire project, for example search for this: "The `onError` callback function will execute anytime an error is thrown by the modal window."

wscherphof: See 2988d95bd3f97cea488329ea383bf5b839dcf378

packages/common/src/interfaces/compositeEditorOpenDetailOption.interface.ts
wscherphof: I agree the onClose is a bit tricky. Wouldn't go for the breaking change though.

wscherphof:
> it looks like all Cypress E2E tests are failing for all frameworks on the same Example 30 test...
> 
> image
> 
> We also need unit test to cover the new event, you could maybe add a spy to the `onDispose` on any of the test. I think a good place to put it might in be in this test below, you could add the `onDispose` check as well and an `expect` that it was called
> 
> https://github.com/ghiscoding/slickgrid-universal/blob/44a74203183a8a3ec74b35ef76768053d619a274/packages/composite-editor-component/src/slick-composite-editor.component.spec.ts#L695-L702

New test for onDispose in 480007502edeb95e1a667a906754b5c4074b201e.

E2E: removed the check on "no row selected after mass-selection" in fc28c6173bcef243c25882c7f4039e0ae8cef74e.

ghiscoding:
> E2E: removed the check on "no row selected after mass-selection" in [fc28c61](https://github.com/ghiscoding/slickgrid-universal/commit/fc28c6173bcef243c25882c7f4039e0ae8cef74e).
> EDIT: I found some other instances now; see [d97b8bd](https://github.com/ghiscoding/slickgrid-universal/commit/d97b8bdd6d5a2e6586ab801e2f6560a003667d9a), where I also set the selectedRanges to `[]` when modalType = `mass-update`, because the selection has no meaning to the editor in that case.

Yeah ok I did see your change and I agree on the Mass Update to remove all selections, but I wonder if removing all the E2E tests that were previously expecting the row selection to be cleared is the correct behavior. I mean, I wonder if your change is correct or if we should rather add a new option flag for this? I'm saying this because you're changing the behavior so in theory it's a slight breaking change.

@zewa666 you're also using the Composite Editor right? This PR would include a behavior change since it would now reapply the Row Selection that the user had before opening the Modal window. That's a behavior change because at the moment, when Mass Selection changes are applied and the user clicks Save, it currently clears the Row Selection... so I'm wondering if we should add a new flag to let the user decide on the behavior. What do you think?

below is the current behavior

msedge_HVjUnn4z6A

zewa666:
no we don't use the composite editor since our requirements are to show related grids, hence we went with using the column definitions and so on but building a custom form with ngxformly with those.

I think the proposed change here is good, and while its a breaking change I feel like its actually more of a fix. I dont see a use case why the selection *should* be cleared. so I'd say its good 👍

ghiscoding:
alright thanks for the feedback, let's merge as it is then. Thanks a lot for the contribution, expect a new release over the weekend :)

PR #2009, opened 2025-06-17
feat(filter): allow processing empty search term via flag

ghiscoding:
- an empty filter search term will return everything and that is not being processed or saved by Grid State/Presets. So to deal with that I added `processEmptySearchTerms` as a flag that can be enabled as a `filter` option and with that enabled, 
- also note that we previously had a flag named `emptySearchTermReturnAllValues` but it was a bit lengthy and not exactly clear what it was meant to do, it was also working only on an ms-select filter, the new flag works for both ms-select and text filter. 
- Also note that the new `processEmptySearchTerms` flag is the complete inverse of `emptySearchTermReturnAllValues` and the latter is now a `@deprecated` flag (to be removed in next major)

### DIFF
--- a/docs/column-functionalities/filters/select-filter.md
+++ b/docs/column-functionalities/filters/select-filter.md
@@ -198,24 +198,26 @@ this.columnDefinitions = [
 ```
 
 ### How to filter empty values?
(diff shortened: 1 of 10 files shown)


ghiscoding:
I might have to roll this back or change it a bit since filtering whitespace should already work and probably be extended to Grid State/Presets. See previous PR #1570

PR #2026, opened 2025-06-25
fix: make sure Row Detail item exists before trying to access it

carlosdelros:
This Pull Request adds a check to ensure item exists before trying to access a property in it.

In situations where rows need to be updated like changing the page, filtering, sorting, etc. the event `onBeforeRemoveCachedRow` is triggered which in turn triggers the RowDetailViewPlugin's method handleRemoveRow. Sometimes the index provided to this method can't be found in the dataset so the line `const item = this.dataView.getItemByIdx(rowIndex)` yields undefined.

### DIFF
--- a/packages/row-detail-view-plugin/src/slickRowDetailView.ts
+++ b/packages/row-detail-view-plugin/src/slickRowDetailView.ts
@@ -785,7 +785,7 @@ export class SlickRowDetailView implements ExternalResource, UniversalRowDetailV
   protected handleRemoveRow(rowIndex: number): void {
     const item = this.dataView.getItemByIdx(rowIndex);
 
-    if (!item) {
+    if (item) {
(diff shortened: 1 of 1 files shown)


ghiscoding:
There's missing code coverage and linting formatting it seems, you could run `pnpm lint` or a full build via `pnpm build`. But for the code coverage, I would go with the inverse and use if true condition instead which is better for coverage and is actually using less lines of code...

```ts
  protected handleRemoveRow(rowIndex: number): void {
    const item = this.dataView.getItemByIdx(rowIndex);
    if (item) {
      // all the code
    }
}
```

ghiscoding:
looks good thanks

ghiscoding:
Released in [v9.3.0](https://github.com/ghiscoding/slickgrid-universal/releases/tag/v9.3.0), thanks for the contribution. ⭐

Issue #2028, opened 2025-06-26
CustomTooltip Drops Title on custom Filter components


zewa666:
### Describe the bug

Hey there @ghiscoding, long time no hear. Hope you're doing well.
Yesterday we've found a new side-effect bug with the Custom Tooltip Plugin when applied to our custom filter component.

So our foreign key component is used both as an editor and filter at the same time. It essentially hosts an ng-select component and allows to perform lookups and select one or many. In the case of multiple selection we found that our tooltip is always dropped for the first selected bubble indicator, but works for the rest of em.

![Image](https://github.com/user-attachments/assets/71ec2b33-6152-4fdb-83c4-963a3ecc4b45)

### Reproduction

https://github.com/user-attachments/assets/1a0992b4-1ad2-4317-967b-747a1ad317f8

The issue appears in the slickCustomTooltip.ts -> swapAndClearTitleAttribute method https://github.com/ghiscoding/slickgrid-universal/blob/2d18db3a7b90583dcf7939d8384522989731f7c5/packages/custom-tooltip-plugin/src/slickCustomTooltip.ts#L568-L573

where as a precaution, in case that the cell itself contains a title, it also gets removed, so that the native browser tooltip does not appear. The issue here is that the selector `this._cellNodeElm?.querySelector('[title]')` actually targets the first of any child elements, thus always the first bubble.

I've reproduced that by modifying the Example16 like in this diff:

[diff.patch](https://github.com/user-attachments/files/20924795/diff.patch)

### Which Framework are you using?

Vanilla / Plain JS

### Environment Info

```shell
| Executable          | Version |
| ------------------- | ------- |
| (framework used)    | VERSION |
| Slickgrid-Universal | VERSION |
| TypeScript          | VERSION |
| Browser(s)          | VERSION |
| System OS           | VERSION |
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/ghiscoding/slickgrid-universal/blob/master/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Wikis](https://github.com/ghiscoding/slickgrid-universal/wiki).
- [x] Check that there isn't [already an issue](https://github.com/ghiscoding/slickgrid-universal/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/ghiscoding/slickgrid-universal/discussions).
[...]

PR #2029, opened 2025-06-26
fix: only toggle title and data-slick-tooltip if value is defined

zewa666:
this PR makes sure that the toggling of the title and data-slick-tooltip attributes of the customTooltip plugin only happens if there is a value to set.

closes https://github.com/ghiscoding/slickgrid-universal/issues/2028

### DIFF
--- a/frameworks/angular-slickgrid/src/demos/examples/custom-angularComponentFilter.ts
+++ b/frameworks/angular-slickgrid/src/demos/examples/custom-angularComponentFilter.ts
@@ -14,13 +14,14 @@ import {
   type SlickGrid,
   unsubscribeAllObservables,
 } from '../../library';
+import { FilterNgSelectComponent } from './filter-ng-select.component';
 
 export class CustomAngularComponentFilter implements Filter {
   private _shouldTriggerQuery = true;
   private _subscriptions: Subscription[] = [];
 
   /** Angular Component Reference */
-  componentRef!: ComponentRef<any>;
+  componentRef!: ComponentRef<FilterNgSelectComponent>;
 
   grid!: SlickGrid;
   searchTerms: SearchTerm[] = [];
@@ -79,11 +80,11 @@ export class CustomAngularComponentFilter implements Filter {
         this.componentRef = componentOuput.componentRef;
 
         this._subscriptions.push(
-          componentOuput.componentRef.instance.onItemChanged.subscribe((item: any) => {
+          componentOuput.componentRef.instance.onItemChanged.subscribe((items: any[]) => {
             this.callback(undefined, {
               columnDef: this.columnDef,
               operator: this.operator,
-              searchTerms: [item.id],
+              searchTerms: items.map((item) => item.id),
               shouldTriggerQuery: this._shouldTriggerQuery,
             });
             // reset flag for next use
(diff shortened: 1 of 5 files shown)


zewa666:
while at it, I've also updated the custom angular filter and component to make use of signals, as these are seen as the goto solution for modern angular. So that might help others to get an idea of how to wrap their components as well.

ghiscoding:
> while at it, I've also updated the custom angular filter and component to make use of signals, as these are seen as the goto solution for modern angular. So that might help others to get an idea of how to wrap their components as well.

ah yeah I haven't worked with Angular in the past few months, years even, so I haven't had a chance to try out all of these things like Signal and Standalone Components. I guess that eventually we should switch all these examples to Standalone but I assume it might be a "all or nothing" kinda situation!? Or can we actually convert them 1 by 1 when still using legacy Angular structure?

zewa666:
yep, it should be possible to convert components one by one to both signals and standalone. ngModule style isn't going away ... still.

Ok I think this one should be good now, the new E2E test makes sure to test the tooltips as well and that way have a protection against a regression.

ghiscoding:
Great, I wouldn't mind a PR for one example conversion if you have a chance. I'm not exactly sure when I'll publish a new release since I spoke to the other Salesforce team and they're finally assigning time to update to the latest version 9 but they come from version 3 so there's a few small issues popping up and I'd like to fix them all before a new release. I guess the pkg-new-pr might be a good time to try it, I think it's something like this

```ts
"pnpm": {
  "overrides": {
    "@slickgrid-universal/common": "https://pkg.pr.new/ghiscoding/slickgrid-universal/@slickgrid-universal/common@2029"
  }
}
```

zewa666:
no probs as I typically just monkey patch or subclass until its released. which component/example would you have in mind for the update?

ghiscoding:
> which component/example would you have in mind for the update?

@zewa666  any one of them, basic or not isn't important, I'm just interested to see which couple of steps are necessary to do one at a time migration. Thanks :)

ghiscoding:
@zewa666 so how do you like the all-in-here (Universal + Angular + ....)? Kinda cool to have it all at hand right!? It's a bit longer to install but it's all there, I like it myself 🦖

zewa666:
oh yeah forgot to mention. its an absolute blessing. so much nicer to work with as I can right away test the effects both in vanilla and also angular. well done 🎉

ghiscoding:
> oh yeah forgot to mention. its an absolute blessing. so much nicer to work with as I can right away test the effects both in vanilla and also angular. well done 🎉

indeed love it too and on top of that Slickgrid-Universal is now the perfect name to represent all of that 🚀

ghiscoding:
Released in [v9.3.0](https://github.com/ghiscoding/slickgrid-universal/releases/tag/v9.3.0)

ghiscoding:
@zewa666 I noticed that this PR change with the modified Cypress test is a bit flaky and seems to be constently showing something like this (it always seems to require multiple pass)

image

so anyway, if you ever get a chance to look into it, that would be great to avoid the test being so flaky. Cheers

zewa666:
alright, I'll take a look at this one next week

PR #2066, opened 2025-07-23
feat: add columnSortByName option to ColumnPicker and GridMenu interfaces

mcallegario:
# Pull Request: Add columnSortByName Option to ColumnPicker and GridMenu Interfaces

## Summary
This pull request introduces a new `columnSortByName` feature that enables alphabetical sorting of columns in both the Column Picker and Grid Menu components. This enhancement addresses a critical usability issue where users struggle to locate specific columns in grids with many columns, significantly improving the user experience for complex data grids.

## Problem Statement
**In grids with many columns, users have difficulty finding the column they want to make visible.** When dealing with datasets containing dozens or hundreds of columns, the default column order (which often follows data structure or creation order) can make it extremely challenging for users to:
- Quickly locate a specific column in the picker list
- Efficiently manage column visibility in large datasets
- Navigate through extensive column lists without frustration

This becomes particularly problematic in enterprise applications where data grids commonly contain 50+ columns, leading to poor user experience and decreased productivity.

## Solution
The new `columnSortByName` option provides **alphabetical organization** of columns, making it intuitive and efficient for users to find any column regardless of the grid's size or complexity.

### 🆕 New Feature: Column Sorting by Name
- **Added `columnSortByName` boolean option** to `ColumnPickerOption` and `GridMenuOption` interfaces
- **Enhanced `populateColumnPicker()` function** in `extensionCommonUtils.ts` to support alphabetical column sorting
- **Implemented locale-aware sorting** using `localeCompare()` for consistent results across different languages and regions

### 🔧 Refactoring: Optimized Column Visibility Logic
- **Improved `handleColumnPickerItemClick()` function** to use a more efficient column lookup mechanism
- **Replaced index-based iteration** with column ID-based matching for better reliability
- **Enhanced checkbox-to-column mapping** using `dataset.columnid` for precise identification

### Column Sorting Algorithm
```typescript
// New sorting logic added to populateColumnPicker function
let sortedColumns = context.columns;
if (addonOptions?.columnSortByName === true) {
  // create a sorted copy of the columns array based on the "name" property
  sortedColumns = [...context.columns].sort((a, b) => {
    const nameA = a.name?.toString().toLowerCase() || '';
    const nameB = b.name?.toString().toLowerCase() || '';
    return nameA.localeCompare(nameB);
  });
}
```

### Usage Examples
```typescript
// Enable alphabetical sorting in Column Picker
const columnPickerOptions: ColumnPickerOption = {
  columnSortByName: true,
  columnTitle: "Available Columns"
};

// Enable alphabetical sorting in Grid Menu
const gridMenuOptions: GridMenuOption = {
  columnSortByName: true,
  columnTitle: "Manage Columns"
};

### DIFF
--- a/packages/common/src/extensions/extensionCommonUtils.ts
+++ b/packages/common/src/extensions/extensionCommonUtils.ts
@@ -65,9 +65,13 @@ export function handleColumnPickerItemClick(this: SlickColumnPicker | SlickGridM
     context._areVisibleColumnDifferent = true;
     const columnId = event.target.dataset.columnid || '';
     const visibleColumns: Column[] = [];
-    context._columnCheckboxes.forEach((columnCheckbox: HTMLInputElement, idx: number) => {
-      if (columnCheckbox.checked) {
-        visibleColumns.push(context.columns[idx]);
+
+    // Iterate through columns and add those that have checked checkboxes to visibleColumns
+    context.columns.forEach((column: Column) => {
+      const columnId = column.id;
+      const checkbox = context._columnCheckboxes.find((cb: HTMLInputElement) => cb.dataset.columnid === columnId.toString());
+      if (checkbox?.checked) {
+        visibleColumns.push(column);
       }
     });
 
@@ -162,7 +166,17 @@ export function populateColumnPicker(this: SlickColumnPicker | SlickGridMenu, ad
   const isGridMenu = context instanceof SlickGridMenu;
   const menuPrefix = isGridMenu ? 'gridmenu-' : '';
 
-  for (const column of context.columns) {
+  let sortedColumns = context.columns;
+  if (addonOptions?.columnSortByName === true) {
+    // create a sorted copy of the columns array based on the "name" property
+    sortedColumns = [...context.columns].sort((a, b) => {
+      const nameA = a.name?.toString().toLowerCase() || '';
+      const nameB = b.name?.toString().toLowerCase() || '';
+      return nameA.localeCompare(nameB);
+    });
+  }
+
+  for (const column of sortedColumns) {
     const columnId = column.id;
     const columnLiElm = document.createElement('li');
     if ((column.excludeFromColumnPicker && !isGridMenu) || (column.excludeFromGridMenu && isGridMenu)) {
--- a/packages/common/src/interfaces/columnPicker.interface.ts
+++ b/packages/common/src/interfaces/columnPicker.interface.ts
@@ -20,6 +20,9 @@ export interface ColumnPickerOption {
   /** Same as "columnTitle", except that it's a translation key which can be used on page load and/or when switching locale */
   columnTitleKey?: string;
 
+  /** Defaults to false, which will sort the columns by name */
+  columnSortByName?: boolean;
+
   /** Defaults to "Force fit columns" which is 1 of the last 2 checkbox title shown at the end of the picker list */
   forceFitTitle?: string;
 
--- a/packages/common/src/interfaces/gridMenuOption.interface.ts
+++ b/packages/common/src/interfaces/gridMenuOption.interface.ts
@@ -33,6 +33,9 @@ export interface GridMenuOption {
   /** Same as "columnTitle", except that it's a translation key which can be used on page load and/or when switching locale */
   columnTitleKey?: string;
 
+  /** Defaults to false, which will sort the columns by name */
+  columnSortByName?: boolean;
+
   /** Defaults to "left", which side to align the grid menu dropdown? */
   dropSide?: 'left' | 'right';
 

### CODE REVIEW

**Comment by ghiscoding** [comment]
If that new flag isn't defined, does it still use the default behavior? This change would most probably impact the Grid State/Preset were the column positions in the array is very important. Also please note that you'll need to add test coverage for this new feature. I'm also not sure what is wrong with the CI, it started failing yesterday on pnpm install, need to investigate 

cc @zewa666 what do you think of this feature?

**Comment by zewa666** [comment]
I think thats a great one but I'd even go a step further. allow to pass in a custom sort function so that the consumer can define how the columns get sorted.

**Comment by ghiscoding** [comment]
> I think thats a great one but I'd even go a step further. allow to pass in a custom sort function so that the consumer can define how the columns get sorted.

That's a great idea, I think with that in mind, I might actually prefer to only have a `sortColumns` function option and simply remove the flag, this would let the user sort however they want and keep the library quite lean (aka small) since that would be quite simple to implement. I've done similar approach in the ms-select library in this [PR](https://github.com/ghiscoding/multiple-select-vanilla/pull/358), it should be as simple as that PR, something along these lines and adding unit tests should be easy as well
```ts
if (this.columns && this.addonOptions.sortColumns) {
    this.columns = this.columns.sort(this.addonOptions.sortColumns);
}
```

I also updated the pnpm lock and that fixed the CI, it must have been corrupted somehow, CI is all good now so you might need to resync or rebase your branch. Thanks

mcallegario:
Thanks for the comments!

>If that new flag isn't defined, does it still use the default behavior?

Yes.


>his change would most probably impact the Grid State/Preset were the column positions in the array is very important.

Yes, I identified this need and also adjusted the code of the handleColumnPickerItemClick function, to preserve the order of the grid columns.

 ```typescript
   // Iterate through columns and add those that have checked checkboxes to visibleColumns
    context.columns.forEach((column: Column) => {
      const columnId = column.id;
      const checkbox = context._columnCheckboxes.find((cb: HTMLInputElement) => cb.dataset.columnid === columnId.toString());
      if (checkbox?.checked) {
        visibleColumns.push(column);
      }
    });
```


>I think thats a great one but I'd even go a step further. allow to pass in a custom sort function so that the consumer can define how the columns get sorted.

That's a great idea, I'll implement it.

**Comment by zewa666** [comment]
@mcallegario thx, thats really gonna add more value to it. while at it, would you mind updating the e2e tests to check for that along with a quick hint in the docs (md files)?

**Comment by mcallegario** [comment]
## 🔍 Code Analysis: Column Sorting Implementation

This PR introduces a column sorting mechanism using a custom sort function.

### 🎯 Core Implementation Details

#### **Enhanced `populateColumnPicker()` Function**
The main enhancement occurs in the `populateColumnPicker()` function where we've added intelligent column sorting:

```typescript
let sortedColumns = context.columns;
if (typeof addonOptions?.columnSort === 'function') {
  // create a sorted copy of the columns array based on the "name" property
  sortedColumns = [...context.columns].sort(addonOptions.columnSort);
}
````

## 📝 Demo Implementation 

This PR includes a practical demonstration of the new `columnSort` feature in the vanilla demo (`example01.ts`), showcasing how developers can implement alphabetical column sorting in real-world applications.

### Demo Code Example
The second grid in the demo now demonstrates the column sorting functionality:

```typescript
gridMenu: {
  // enable the "columnSort" option to sort columns by name
  columnSort: (item1: Column, item2: Column) => {
    const nameA = item1.name?.toString().toLowerCase() || '';
    const nameB = item2.name?.toString().toLowerCase() || '';
    return nameA.localeCompare(nameB);
  },
  // ... other grid menu options
}
```

ghiscoding (changes requested):

ghiscoding:
@mcallegario I left some comments and also note that since you modified the Example 01, its associated Cypress E2E test now fails which will require some changes I guess. I'll also mention again that I want full unit test coverage, so please add at least 1 or 2 unit tests to cover the new sort function. Also review the comment that @zewa666 left [above](https://github.com/ghiscoding/slickgrid-universal/pull/2066#issuecomment-3111950118). Thanks

Side note, the .md file that @zewa666 mentioned are the documentations available for the project, it would be helpful to have them updated. So the code analysis and demo you've put above should be moved to these docs
- https://github.com/ghiscoding/slickgrid-universal/blob/master/docs/grid-functionalities/Column-Picker.md
- https://github.com/ghiscoding/slickgrid-universal/blob/master/docs/grid-functionalities/Grid-Menu.md

Also note that all framework wrappers were moved in here, it would be helpful if you can modify all wrappers docs as well (if possible). Thanks again

zewa666:
while this may sound super nitpicky, rest assured the docs and especially e2e tests have covered our *** already a couple of times and protected against contradicting feature requests. Slickgrid is a huge lib and housekeeping becomes a super important task.

if you struggle with any bit let us know and we'll help out

packages/common/src/interfaces/columnPicker.interface.ts
```diff
@@ -20,6 +20,9 @@ export interface ColumnPickerOption {
   /** Same as "columnTitle", except that it's a translation key which can be used on page load and/or when switching locale */
   columnTitleKey?: string;
 
+  /** Custom sort function for the columns in the grid menu */
```
ghiscoding, line 23: oops super nitpicky but there's another typo here, this is the Column Picker interface and you're mentioning Grid Menu. 😉

packages/common/src/extensions/extensionCommonUtils.ts
```diff
@@ -65,9 +65,13 @@ export function handleColumnPickerItemClick(this: SlickColumnPicker | SlickGridM
     context._areVisibleColumnDifferent = true;
     const columnId = event.target.dataset.columnid || '';
     const visibleColumns: Column[] = [];
-    context._columnCheckboxes.forEach((columnCheckbox: HTMLInputElement, idx: number) => {
-      if (columnCheckbox.checked) {
-        visibleColumns.push(context.columns[idx]);
+
+    // Iterate through columns and add those that have checked checkboxes to visibleColumns
+    context.columns.forEach((column: Column) => {
```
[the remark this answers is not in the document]
ghiscoding: ok thanks for the explanation, seems to make sense

ghiscoding (changes requested):

packages/common/src/interfaces/columnPicker.interface.ts
ghiscoding: all good thanks :)

ghiscoding:
@mcallegario This is now published in [v9.6.0](https://github.com/ghiscoding/slickgrid-universal/releases/tag/v9.6.0), thanks again for another great contribution 🚀

PR #2095, opened 2025-08-08
chore(deps): update TypeScript to v5.9 and fix type issue

ghiscoding:


ghiscoding:
will have to close it and skip TypeScript because I couldn't get Angular-Slickgrid tests to pass with this change

PR #2132, opened 2025-08-30
chore(deps): migrate to Rolldown-Vite

ghiscoding:


ghiscoding:
definitely has a lot of problems with all frameworks, so let's close for now.

PR #2144, opened 2025-09-10
fix: never hide `Editors.longText` in Composite Editor modal

ghiscoding:
- we should never hide the `Editors.LongText` when it's a Composite Editor
   - this issue came up when doing multiple Row Selections and then clicking on Mass Update, the textarea wasn't showing
   - this only happened with the steps shown above and only in Mass Update (not on Mass Selections)
- add `isCompositeEditor` to all Editor instantiations and refactor Composite Editor Factory with better typing and assign use this new option as `isCompositeEditor: true` so that it's a lot easier to detect when each Editor is a Composite Editor

msedge_4v1m82esVD


packages/common/src/editors/longTextEditor.ts
```diff
@@ -185,13 +184,15 @@ export class LongTextEditor implements Editor {
   }
 
   hide(): void {
-    this._wrapperElm.style.display = 'none';
+    // only hide only when it's an inline editor (prevent hiding on Composite Editor)
```
ghiscoding, line 187: little typo, "only" was entered twice

PR #2153, opened 2025-09-12
feat: use Web Worker for Excel export to improve perf

gotstu:


ghiscoding:
Perf is pretty damn good, for "Example 2 - Grouping & Aggregators" with 50K rows, the regular Excel Export takes 1m12sec. and with this new Web Worker it takes 6sec. and for 500K rows it took 58sec. It's pretty impressive. 

There's a lot of code in here (>4500 LOC seems like a lot), I see that there's a fallback in case the Web Worker doesn't work but I'm not sure if want to keep this one in the same service or not. The other option would be to move this to another optional Service/Package but that would also mean less people testing it. Let's discuss it next week, we also need full test coverage, do we need the TESTING GUIDE and `demo/webWorkerDemo.html`?

cc @zewa666 
FYI, this new contributor is a coworker trying out a paid LLM Service. I've been trying to do this switch to Web Worker for a while but it was quite challenging with just copilot. 

EDIT

I think another approach could be to use an external library like [piscina](https://piscinajs.dev/) to more easily pass data and object I think to the web worker, it even works with Streams (which I added recently to the excel-builder lib) which could even improve perf further. I'm basically trying to avoid having to recreate all Formatters and everything in the Web Worker since I find it's not bullet proof and a bit risky and so I would rather we just pass instances to the Web Worker if possible

packages/excel-export/src/utils/workerManager.ts
```diff
@@ -0,0 +1,1321 @@
...
+
+    return await fallbackTask();
+  }
+
+  /**
+   * Create the worker script as a string
+   * This includes the entire worker implementation inline to avoid CSP issues
+   */
+  private createWorkerScript(): string {
+    return `
```
ghiscoding, line 278: wow ok, I don't really like this for 2 major reasons:
- adding functions as a string is quite risky and doesn't allow for proper TS Typings
- this `BUILT_IN_FORMATTERS_MAP` seems to only loop through built-in formatters, but what about Custom Formatters?

ghiscoding: I asked ChatGPT on how we could write functions and stringify them. The best option is to move that to a separate `.ts` file and then load it to the web worker by using Node read file. Below is what ChatGPT gave me

```ts
// src/worker-fns.ts
export function heavyCalc(a: number, b: number): number {
  return a ** b + Math.sqrt(a + b);
}
```
and then 
```ts
// src/main.ts
import { readFile } from 'node:fs/promises';          // only runs in the build step
import { fileURLToPath } from 'node:url';

// Resolve the source file as an URL (ESM‑compatible)
const fnPath = fileURLToPath(import.meta.resolve('./worker-fns.ts'));

// Build‑time async IIFE – this code is executed by the bundler, not at runtime
const workerSrc = await (async () => {
  const src = await readFile(fnPath, 'utf8');        // raw TypeScript source
  // The bundler will transpile the file to JavaScript before it reaches the browser
  return `
    ${src}
    self.onmessage = (e) => {
      const result = heavyCalc(e.data.a, e.data.b);
      self.postMessage(result);
    };
  `;
})();

const worker = new Worker(
  URL.createObjectURL(new Blob([workerSrc], {type: 'text/javascript'}))
);
```

gotstu:
Going to abandon this PR

ghiscoding:
> Going to abandon this PR

Yeah that was a good try but in the end, there's too many duplicate code (mainly all the Formatters/GroupingFormatters and all their associated functions).

PR #2162, opened 2025-09-17
feat: Drag-Fill capability and Hybrid SelectionModel plugin

ghiscoding:
Thanks to Ben (6pac) for starting this new feature in SlickGrid repo:

> New HybridSelectionModel combines cell and row selection models. It functions like CellSelectionModel except when certain conditions switch it into RowSelectionModel. If the options.rowSelectOverride function exists, then HybridSelectionModel calls it expecting a boolean return to determine whether to behave like RowSelectionModel. If this option is not set, then selection of a column that is a member of the array options.rowSelectColumnObjectArr will be treated with the behaviour of RowSelectionModel.
> 
> Drag and Replace option mirrors Excel as closely as possible. When a range of cells is selected, a grey box appears at the bottom right of the selection area, and dragging that grey box expands the selection area and replaces the newly selected cells will the contents of the existing cells. Excel specifically only allows expansion horizontally or vertically, but here we allow both. The current behaviour may be modified. Note that the spreadsheet example the FormulaEditor causes problems with this new feature because it creates a new CellRangeSelector each time a cell is edited. This editor needs to add an event to the existing grid's CellRangeSelector, or create a new one.

SlickPR_DragFill_Hybrid

### DIFF
--- a/demos/vanilla/src/app-routing.ts
+++ b/demos/vanilla/src/app-routing.ts
@@ -36,6 +36,8 @@ import Example32 from './examples/example32.js';
 import Example33 from './examples/example33.js';
 import Example34 from './examples/example34.js';
 import Example35 from './examples/example35.js';
+import Example36 from './examples/example36.js';
+import Example37 from './examples/example37.js';
 
 export class AppRouting {
   constructor(private config: RouterConfig) {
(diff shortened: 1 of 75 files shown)


ghiscoding:
and there you have it... done! 

It does increase the zip file by 3Kb but it should be decreasing back to similar size in the next major version because I expect to simply drop `SlickCellSelectionModel` & `SlickRowSelectionModel` and keep only `SlickHybridSelectionModel` in the future. For now we'll have all 3 models and user will have to enable this new Hybrid model via `enableHybridSelection` (not sure if I'll keep this flag in the future or not)

PR #2192, opened 2025-10-13
feat: remove multiple cellCssStyles using predicate

zewa666:
this allows to delete multiple cellCssStyles in one go by providing a predicate that returns a boolean based on the key and hash


zewa666:
the background for this feature is that our app is multiple cell markers like read-only and unsaved-changes etc. There is a feature where based on specific criteria, some of these need to be resetted alltogether. Currently, in order to do so, I need to access the private cellCssStyles property. This would provide a cleaner solution.

ghiscoding:
ok this looks like a good enhancement, if you could just review the comment I left then I think we can then merge it after. Thanks

ghiscoding:
looks good thanks :)

PR #2197, opened 2025-10-16
chore(deps): remove prettier-oxc plugin

ghiscoding:


ghiscoding:
it's 29sec without prettier-oxc plugin and 24sec with the plugin, so ok it might be ok to keep the plugin since I guess 5 sec. isn't that negligible (20% time decrease)

The next OXC Formatter promises to be extremely fast (I'm expecting < 5 sec.) even though it isn't ready yet (not even in Beta yet), so we'll have to wait couple more months.

Issue #2200, opened 2025-10-20
Cannot run the slick-grid react app


nawabfurquan:
### Describe the bug

Encountered the following error about regenerator-runtime while trying to run the application both in [Slick Grid React Demo](https://github.com/ghiscoding/slickgrid-react-demos) and in a fresh new React project using the [Quick Start Tutorial](https://ghiscoding.gitbook.io/slickgrid-react/getting-started/quick-start):

```
> slickgrid-react@4.0.0 dev
> vite


  VITE v7.1.6  ready in 582 ms

  ➜  Local:   http://localhost:8080/
  ➜  press h + enter to show help
✘ [ERROR] Could not resolve "regenerator-runtime/runtime.js"

    node_modules/slickgrid-react/dist/index.js:1:7:
      1 │ import 'regenerator-runtime/runtime.js';
        ╵        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "regenerator-runtime/runtime.js" as external to exclude it from the bundle,
  which will remove this error and leave the unresolved path in the bundle.

/Users/furquan/Documents/slickgrid-react-demos/bootstrap5-i18n-demo/node_modules/esbuild/lib/main.js:1477
  let error = new Error(text);
              ^

Error: Error during dependency optimization:

✘ [ERROR] Could not resolve "regenerator-runtime/runtime.js"

    node_modules/slickgrid-react/dist/index.js:1:7:
      1 │ import 'regenerator-runtime/runtime.js';
        ╵        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "regenerator-runtime/runtime.js" as external to exclude it from the bundle, which will remove this error and leave the unresolved path in the bundle.


    at failureErrorWithLog (/Users/furquan/Documents/slickgrid-react-demos/bootstrap5-i18n-demo/node_modules/esbuild/lib/main.js:1477:15)
    at /Users/furquan/Documents/slickgrid-react-demos/bootstrap5-i18n-demo/node_modules/esbuild/lib/main.js:946:25
    at /Users/furquan/Documents/slickgrid-react-demos/bootstrap5-i18n-demo/node_modules/esbuild/lib/main.js:1355:9
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  errors: [Getter/Setter],
  warnings: [Getter/Setter]
}

Node.js v23.7.0
```

**Points to note:**

[...]

PR #2201, opened 2025-10-20
chore(deps): remove regenerator-runtime dependency

nawabfurquan:
Removed the `regenerator-runtime` dependency from `slickgrid-react`. This dependency was not used and was causing runtime error due to missing dependency.
This PR fixes the issue: https://github.com/ghiscoding/slickgrid-universal/issues/2200

### DIFF
--- a/demos/react/package.json
+++ b/demos/react/package.json
@@ -66,7 +66,6 @@
     "cypress": "catalog:",
     "cypress-real-events": "catalog:",
     "globals": "catalog:",
-    "regenerator-runtime": "^0.14.1",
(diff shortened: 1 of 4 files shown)


ghiscoding:
oops I didn't see your PR in advance and it looks like we did the same changes. Thanks for that, please note that I'm not sure when exactly a new release will be pushed since I'm waiting for another fix by another user, so it might take some time but at least you can just install the dependency for the time being. Also if the project is useful to you, then you could upvote it if you like, thanks ⭐

PR #2207, opened 2025-10-27
fix: stops custom tooltip from flickering when hovered, fixes #2206

### LINKED ISSUE #2206: Flickering Custom Tooltips
### Describe the bug

With limited space, there are occassions where the tooltip hasn't enough space and renders in a slot where the cursor will manage to hover over the tooltip. When that happens, every move of the mouse causes a flickering re-render of the tooltip

I assume it has something to do with how the mouseover is handled for the tooltips but haven't yet dug deeper

### Reproduction

* Open Slickgrid Universal Example 16
* Open the devtools and reduce the visible area
* hover over the cells of the 3rd column
* 
Image

Angular

### Environment Info

```shell
```

### Validations

zewa666:
fixes #2206 

This fixes the issue in such a way, that a timeout is added to debounce multiple quick enter/leave event combos. Additionally a 3s total timeout is added to remove the tooltip

https://github.com/user-attachments/assets/f48f62c5-68ac-4660-bc0c-d36d0012bbe3

references https://github.com/ghiscoding/slickgrid-universal/issues/2206

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -222,7 +222,7 @@ importers:
         version: 5.8.3
       vitest:
         specifier: 'catalog:'
-        version: 4.0.15(@types/node@24.10.1)(@vitest/ui@4.0.15)(esbuild@0.25.11)(happy-dom@20.0.0)(jiti@2.6.1)(jsdom@27.2.0(postcss@8.5.6))(less@4.4.1)(sass-embedded@1.93.2)(sass@1.94.2)(terser@5.44.0)(yaml@2.8.2)
+        version: 4.0.15(@types/node@24.10.1)(@vitest/ui@4.0.15)(esbuild@0.25.11)(jiti@2.6.1)(jsdom@27.2.0(postcss@8.5.6))(less@4.4.1)(sass-embedded@1.93.2)(sass@1.94.2)(terser@5.44.0)(yaml@2.8.2)
 
   demos/aurelia:
     dependencies:
@@ -851,7 +851,7 @@ importers:
         version: 5.1.4(typescript@5.8.3)(vite@6.4.1(@types/node@24.10.1)(jiti@2.6.1)(less@4.4.1)(lightningcss@1.30.2)(sass-embedded@1.93.2)(sass@1.94.2)(terser@5.44.0)(yaml@2.8.2))
       vitest:
         specifier: 'catalog:'
-        version: 4.0.15(@types/node@24.10.1)(@vitest/ui@4.0.15)(esbuild@0.25.11)(happy-dom@20.0.0)(jiti@2.6.1)(jsdom@27.2.0(postcss@8.5.6))(less@4.4.1)(sass-embedded@1.93.2)(sass@1.94.2)(terser@5.44.0)(yaml@2.8.2)
+        version: 4.0.15(@types/node@24.10.1)(@vitest/ui@4.0.15)(esbuild@0.25.11)(jiti@2.6.1)(jsdom@27.2.0(postcss@8.5.6))(less@4.4.1)(sass-embedded@1.93.2)(sass@1.94.2)(terser@5.44.0)(yaml@2.8.2)
       yaml:
         specifier: ^2.8.2
         version: 2.8.2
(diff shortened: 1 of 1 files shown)


zewa666:
so this would be a quick shot at fixing the issue but adds a new behavior with the auto-close after 3s (which most likely should be configurable via customTooltip options).

What do you think about this approach?

ghiscoding:
wow this seems like an extreme usage lol... that seems like a lot of code, why not simply not show the tooltip if there's not enough space when doing the positioning of the tooltip? That would be much simpler and you wouldn't have to add listener to all types of mouse events like you just did. 

https://github.com/zewa666/slickgrid-universal/blob/0dd38bdf8a7093558648e503c84fba5d911d4141/packages/custom-tooltip-plugin/src/slickCustomTooltip.ts#L549-L555

zewa666:
the issue happens for us primarily due to grids with more columns than available width. if only a part of the last column is in the visible area, the available space is limited hence the mouseover. not showing the tooltip would feel more like broken UX in said case as a non tech savy user might not get the underlying reason for the missing tooltip

and yep, its quite some code, but essentially all it does is debouncing current events

ghiscoding:
> and yep, its quite some code, but essentially all it does is debouncing current events

but why not use the suggestion I made to do the check in the tooltip reposition function? I think, and would much prefer, less lines of code for what seems like a little check to do before positioning it.

zewa666:
oh ok didnt notice that on phone. will check that out tomorrow 👍

ghiscoding:
hey so I'm not sure if you'll have time to update this PR soon but I'm probably going to go with a release on Monday-Tuesday..ish since I think I'm done with everything including the new Hybrid Selection Model

zewa666:
ok so I tried modifying it as you suggested but that didn't really help. the flickering is back.

```
const gridContainer = this._grid.getContainerNode();
      const gridRect = gridContainer.getBoundingClientRect();
      const cellRect = this._cellNodeElm.getBoundingClientRect();

      
      if (cellRect.right > gridRect.right || cellRect.left < gridRect.left) {
        this._tooltipElm.remove();
        return;
      }
```

I still think the timeout is the solution to all of this, and that unfortunately takes a bit more of code. So while I could squash everything into the reposition method I feel my original approach is cleaner to read.

zewa666:
if you try to replicate the issue as I've shown in the video its pretty obvious. And yes while that example is overkill, the real situation can easily happen with the last column being cut-off by the viewport when there are more columns than width of the grid available

ghiscoding:
that's ok, it was worth the try, I trust you with the best solution. Just make sure to have enough test coverage. Thanks

ghiscoding:
> so this would be a quick shot at fixing the issue but adds a new behavior with the auto-close after 3s (which most likely should be configurable via customTooltip options).
> 
> What do you think about this approach?

and yes this should be a separate option and keep the previous default of always show tooltip

zewa666:
oki doki I think I finally got it, dang need to get prettier sorted on my PC, those linter issues otherwise are a nightmare 🤣

ghiscoding (changes requested):

packages/custom-tooltip-plugin/src/slickCustomTooltip.ts
```diff
@@ -484,6 +518,26 @@ export class SlickCustomTooltip {
     if (finalOutputText.toString()) {
       document.body.appendChild(this._tooltipElm);
 
+      if (this.addonOptions?.persistOnHover === true) {
+        this._tooltipElm.addEventListener('mouseenter', () => {
```
ghiscoding: let's use my custom service to bind such events, because with my service we can unbind all easily and avoid any event leaking

for example
```
import { BindingEventService } from '@slickgrid-universal/binding';

protected _bindEventService = new BindingEventService();

init() {
  this._bindEventService.bind(editorResetButtonElm, 'click', (e) => console.log('clicked', e));
}

dispose() {
  this._bindEventService.unbindAll();
}
```

zewa666:
@ghiscoding thx for the review. Some good catches you had there. Sry that it took so long but today was crazy tons of work with all those recent NPM attacks and now the new React one as well.

Btw I took the opportunity to pnpm audit this project and detected a critical for happy-dom, which seems to be caused by vitest. Perhaps updating all packages might be a good idea but I didn't want to do it as part of this PR

ghiscoding:
> @ghiscoding thx for the review. Some good catches you had there. Sry that it took so long but today was crazy tons of work with all those recent NPM attacks and now the new React one as well.
> 
> Btw I took the opportunity to pnpm audit this project and detected a critical for happy-dom, which seems to be caused by vitest. Perhaps updating all packages might be a good idea but I didn't want to do it as part of this PR

no worries, I wasn't aware of the new React attack. For Happy-Dom it comes from Vitest and I removed all usage of Happy-Dom that I had couple of weeks ago and I'm now using JSDOM only (finally). For all deps, I run Renovate on a schedule and I always merge them all before running any new release, there's also Dependabot that runs and identified couple of them recently (I had to add some pnpm overrides for some of them). So I think it's ok on the security audit, I always take care of them as much as possible (not really sure why Happy-Dom shows up though, it should be an optional dep)... I just merged the Renovate PR that was opened, it was including updates to React

ghiscoding (approved):

zewa666:
ok cool, lemme know if you want anything else sorted otherwise I'll try to fix up the issues caused by the merge this weekend and call it done

ghiscoding:
yeah sorry about that, it looks like I caused the conflict after merging the PR to update dependencies. I just fixed though, so it looks good now. I'll merge after CI run. 

Out of curiosity, are you also planning on fixing the other issue as well? Next week could be fine too, no pressure, just asking ;)

zewa666:
are you talking about the copy/paste issue? could take a look at that one yeah

ghiscoding:
> are you talking about the copy/paste issue? could take a look at that one yeah

yeah the copy selection issue, that's the only other one left :)

PR #2208, opened 2025-10-28
fix: add correct offset height/width margins

ghiscoding:
- need to readjust default values of SlickCellRangeDecorator offset height/width calculation to remove extra unwanted spacing
- it became a bit more obvious when working on new PR #2162 hence why I'd like to fix it now

### before
image

### after
image

### DIFF
--- a/packages/common/src/extensions/__tests__/slickCellRangeDecorator.spec.ts
+++ b/packages/common/src/extensions/__tests__/slickCellRangeDecorator.spec.ts
@@ -31,7 +31,7 @@ describe('CellRangeDecorator Plugin', () => {
         border: '2px dashed blue',
         zIndex: '9999',
       },
-      offset: { top: -1, left: -1, height: 2, width: 2 },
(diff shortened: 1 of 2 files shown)


ghiscoding:
cc @zewa666 it might be better to extend it in the global grid options in the future, but for now that will do

zewa666:
lol I think we monkey patched something similar already but I didnt feel its worth contributing back bc it felt like a specific issue we have 😅

ghiscoding:
> lol I think we monkey patched something similar already but I didnt feel its worth contributing back bc it felt like a specific issue we have 😅

great that will be 1 less monkey patch on your end 😉 The weird thing is that I wasn't able to replicate this issue in 6pac/slickgrid, so it might be related to my CSS which is different compared to original SlickGrid... so anyway it should be much better now on our end 🎉 

Side note, Ben merged and we released the new Hybrid Selection Mode with Drag-Fill yesterday, so I think that I will try to finish the PR #2162  before making my own release around next week. Cheers

PR #2210, opened 2025-10-28
fix: calculate grid's width using window.innerWidth only when parent-container autoresize is not enabled

### LINKED ISSUE #2209: Resizer service should not use window width if autoresize by container is enabled
### Describe the bug

In certain situations the resizer service is not able to figure out the container element's width so it defaults to use the `window.innerWidth`.

I think using `window.innerWidth` is ok for most cases, but if I have have set `autoResize: true`, `resizeDetection: 'container'` and `calculateAvailableSizeBy: 'container'` the resizer should not use window. If it cant find it it should be ok to use 0.

The relevant piece of code is here:
```
packages/common/src/services/resizer.service.ts 
   -> calculateGridNewDimensions 
   ->  Line 268 -> const availableWidth = getInnerSize(this._pageContainerElm, 'width') || window.innerWidth || 0;
```

### Reproduction

My application is using a multi-tab system. When you switch away from a tab, the tab's content is "cached" using Vue's `KeepAlive` feature. When I come back to the tab, during the rendering slickgrid's resizer service tries to calculate the width of the grid and I'm thinking there is some kind of race condition and it cant find it so it defaults to using window.innerWidth.

Vue

### Environment Info

```shell
| Vue                 |    3    |
| Slickgrid-Universal |.   9.9  |
```

### Validations

carlosdelros:
Fixes #2209 

Fix resizer.service to correctly compute grid width based on the parent container when autoresize is enabled.

Previously, the service used window.innerWidth, which could lead to incorrect sizing. This change mirrors the existing logic used for height calculation.

### DIFF
--- a/packages/common/src/services/resizer.service.ts
+++ b/packages/common/src/services/resizer.service.ts
@@ -256,30 +256,29 @@ export class ResizerService {
 
     let gridHeight = 0;
     let gridOffsetTop = 0;
-    let gridWidth = 0;
 
     // which DOM element are we using to calculate the available size for the grid?
     if (autoResizeOptions.calculateAvailableSizeBy === 'container') {
-      // uses the container's height and width to calculate grid's dimensions without any offset
+      // uses the container's height to calculate grid height without any top offset
       gridHeight = getInnerSize(this._pageContainerElm, 'height') || 0;
-      gridWidth = getInnerSize(this._pageContainerElm, 'width') || 0;
     } else {
       // uses the browser's window height with its top offset to calculate grid height
       gridHeight = window.innerHeight || 0;
       gridOffsetTop = gridElmOffset.top;
-
-      // uses the browser's window width to calculate grid width
-      gridWidth = window.innerWidth || 0;
     }
 
     const availableHeight = gridHeight - gridOffsetTop - bottomPadding;
+    const availableWidth =
+      getInnerSize(this._pageContainerElm, 'width') ||
+      (autoResizeOptions.calculateAvailableSizeBy !== 'container' && window.innerWidth) ||
+      0;
     const maxHeight = autoResizeOptions?.maxHeight;
     const minHeight = autoResizeOptions?.minHeight ?? DATAGRID_MIN_HEIGHT;
     const maxWidth = autoResizeOptions?.maxWidth;
     const minWidth = autoResizeOptions?.minWidth ?? DATAGRID_MIN_WIDTH;
 
     let newHeight = availableHeight;
-    let newWidth = autoResizeOptions?.rightPadding ? gridWidth - autoResizeOptions.rightPadding : gridWidth;
+    let newWidth = autoResizeOptions?.rightPadding ? availableWidth - autoResizeOptions.rightPadding : availableWidth;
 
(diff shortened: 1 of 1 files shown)


ghiscoding:
I thought you said you wanted to fix the resizer width when it uses the `container` type? It looks like there's a few E2E tests failing, I think you might need to put back the original code `getInnerSize(this._pageContainerElm, 'width') || window.innerWidth || 0;` in the `else` when it's not using `container`

~It looks like Example 14 from the Vanilla demos is using the `container` resize, so that one might need adjustments in the E2E test directly~ (wait that's not true, it's not `calculateAvailableSizeBy` but rather `resizeDetection` which is different), all other tests are probably failing because of the unexpected code change. 

https://github.com/ghiscoding/slickgrid-universal/blob/4e7841034572ac81cb90273d00c484f7ec98c84b/demos/vanilla/src/examples/example14.ts#L553-L557

ghiscoding:
Yeah that seems a lot better since all E2E tests are passing now... does that still fix your original issue? I'd be ok to merge with this update if it's ok on your side?

carlosdelros:
> I thought you said you wanted to fix the resizer width when it uses the `container` type? It looks like there's a few E2E tests failing, I think you might need to put back the original code `getInnerSize(this._pageContainerElm, 'width') || window.innerWidth || 0;` in the `else` when it's not using `container`
> 
> ~It looks like Example 14 from the Vanilla demos is using the `container` resize, so that one might need adjustments in the E2E test directly~ (wait that's not true, it's not `calculateAvailableSizeBy` but rather `resizeDetection` which is different), all other tests are probably failing because of the unexpected code change.
> 
> https://github.com/ghiscoding/slickgrid-universal/blob/4e7841034572ac81cb90273d00c484f7ec98c84b/demos/vanilla/src/examples/example14.ts#L553-L557

Sorry about that. I tried to mirror the logic used to calculate the height.

ghiscoding:
Yeah I think the width must be calculated a little different compare to height, it can't be exactly the same code though I don't fully recall why exactly... so anyway, is that ok on your side now? can this be merged?

I'll still do a quick check locally in an hour or so before merging

carlosdelros:
> Yeah that seems a lot better since all E2E tests are passing now... does that still fix your original issue? I'd be ok to merge with this update if it's ok on your side?
> 
> oh wait you went back to the same code?? why?

Thanks yes this fixes the issue.

Yeah, went back to just change this one line. 

```
const availableWidth =
      getInnerSize(this._pageContainerElm, 'width') ||
      (autoResizeOptions.calculateAvailableSizeBy !== 'container' && window.innerWidth) ||
      0;
```

I originally wanted to mirror the calculation done to the height, but those e2e tests break. So just changing the exact place where the issue is.

Thanks for the fast response.

ghiscoding:
looks good after testing locally, thanks for the contribution. Please note that I probably won't release until next week since I'm still working on another large PR (Hybrid Selection Mode). Cheers

ghiscoding:
now available in [v9.10.0](https://github.com/ghiscoding/slickgrid-universal/releases/tag/v9.10.0), thanks again for your contributions :)

PR #2223, opened 2025-11-05
chore: try oxlint jsPlugins with angular-eslint

ghiscoding:
try using oxlint with `"jsPlugins": ["@angular-eslint/eslint-plugin"]` for the `frameworks/angular` project. 

On windows I'm getting this error, let see if it works in CI with Ubuntu

```
> oxlint .

WARNING: JS plugins are experimental and not subject to semver.
Breaking changes are possible while JS plugins support is under development.
memory allocation of memory allocation of memory allocation of 4294967280 bytes failed
4294967280 ELIFECYCLE  Command failed with exit code 3221226505.

 *  The terminal process "C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe -Command pnpm run lint" terminated with exit code: 1.
```

### Edit

I think I'll use the temporary [suggestion](https://github.com/oxc-project/oxc/issues/14375#issuecomment-3493905826) provided in the OXC issue on Windows. I'll revisit it later when the issue is fixed.


overlookmotel:
Did you find that `@angular-eslint/eslint-plugin` worked as an Oxlint JS plugin? (apart from the OOM errors on Windows which we're aware of https://github.com/oxc-project/oxc/issues/14375)

ghiscoding:
@overlookmotel yes it works with your [suggestion](https://github.com/oxc-project/oxc/issues/14375#issuecomment-3493905826) in that same oxc github issue which you suggested to separate the linting into 2 tasks (which is what I did in this PR):
1. lint entire solution with `import` plugin but skip Angular code
2. lint only Angular code with `@angular-eslint/eslint-plugin` but without `import` plugin via a separate [`oxlintrc-angular.json`](https://github.com/ghiscoding/slickgrid-universal/pull/2223/files#diff-b23a160e8dcb002bb972d51a0107c8faebce406356ce187d5dfe102241a6a57c) config file. 

So I'll use that approach for now until https://github.com/ghiscoding/slickgrid-universal/pull/2223 is fixed. oxlint is so damn fast anyway that even with 2 passes, it's not even comparable with eslint. So I'm ok with this approach, it's slightly more code and time but it works and it's awesomely fast, so I'm good for now 🚀 

Note that my code was already linted from my previous eslint setup and I haven't really tried to change some code to verify that the jsPlugins does work. I think I had 1 or 2 warnings with oxlint, but I could verify more in depth if that is what you were asking to confirm?

overlookmotel:
Thanks for coming back.

When you have time, if you're able to put it through its paces a bit and report any problems, that'd be a *huge* help.

My next task is to build a plugin tester, but it turns out it's not really possible to automate because every repo uses different testing libraries etc, so I'll need to test plugins on a case-by-case basis - very laborious. In meantime, bug reports (or even "it works 🚀!") is really helpful.

If we can find bugs, they'll probably not be too hard to fix. The tricky part is hunting them down!

ghiscoding:
@overlookmotel hey sorry for the delay, I just made a quick test and yes `jsPlugins": ["@angular-eslint/eslint-plugin"]` seems to work as intended. I added this linting rule [`"@angular-eslint/no-forward-ref": "error"`](https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin/docs/rules/no-forward-ref.md) just for testing purposes and indeed I am getting the expected error from the angular eslint plugin. So it's working great from my end.... of course that would be so much better after the windows fix (so that I can run only 1 oxlint task instead of 2), but I can certainly wait. 😉 

Thanks for the great work and amazing perf 🚀 

image

overlookmotel:
Thanks very much for testing it and coming back to me. Really appreciate it.

overlookmotel:
OOM bug on Windows was fixed in https://github.com/oxc-project/oxc/pull/17013, which is in Oxlint 1.34.0, released last night.

Would *massively* appreciate feedback to confirm that the fix does indeed work. If it doesn't, please let us know on https://github.com/oxc-project/oxc/issues/14375.

PR #2228, opened 2025-11-10
chore: add react 18 support for slickgrid-react

nawabfurquan:
- Downgraded React version to 18 to support versions 18 and above
- Fixed type errors for the useRefs


ghiscoding:
Sorry but I have no intention of supporting or living in the past, if you can't upgrade to React 19 that is sad for you but still React 19 is over a year old already and this is a free Open Source project and I have no time or bandwidth to support older version of a framework... 

However, there are alternatives, if you can't upgrade to React 19, then just go with the previous v5.x of Slickgrid-React until you can eventually upgrade in the future. You just won't get the latest features, until you can upgrade but that should be limited

PR #2236, opened 2025-11-15
feat: add new `autoEditByKeypress` to open editor by typing a char, fix #2185

### LINKED ISSUE #2185: autoEdit false but activate specific editors on keydown

Hi @ghiscoding, haven't heard each others quite a while, hope you're doing well.

So its finally time I need to tackle the autoEdit mode for our app. As it is currently we're using autoEdit off, as turning it on interferes heavily with cell selections, copy&pastes (paste into editor instead of cell) etc.
Downside is, for simple text inputs you need to perform a double click for activation and and can only than start typing, which isn't really the best UX.

### Suggested solution

What I'd like to propose is another setting that goes along with autoEdit: false and allows to set, per column defintion, whether a keydown should immediately activate the cells underlying editor and bubble on the keydown event to it. This way, a user standing on a cell could start typing right-away as is the behavior in tools like Excel.

### Alternative

_No response_

### Additional context

_No response_

### Validations

- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

ghiscoding:
fixes #2185

When using and enabling both `autoEdit` and `enableExcelCopyBuffer`, starting a cell selection can conflict with the editing part. What this PR does, is to provide another grid option `autoEditByKey`, which only starts editing when user starts typing a character via the keyboard, this avoid conflicting with with cell selection.

#### TODOs
- [x] add unit tests
- [x] add Cypress tests
- [x] add documentation

#### Notes
- [x] not sure about the grid option name, for now I named it `autoEditByKeypress` (meaning start editing by keyboard char)
- won't do much with a Date Editor, it will open the date picker but that's about it (not that useful for that one, but dblClick still works too)
   - I think that to help the date picker, I should look into moving the focus from the grid cell to rather the date picker itself
   - I also think it might be good to do something similar to the slider editor (I think using left/right keys should move slider instead of moving to the next grid cell
- also I asked chatGPT to give me a regex to check if key is a character or not, it seems to work for the most part but I wonder if we should also add a grid option for that? It might not be necessary to get started though, maybe in the future if users have special chars that they want to add to the regex char validation (for now it should work with emoji, unicode, latin, ... good enough probably)

@zewa666 so this should fix your issue and it doesn't require too much code change. I'm not sure about the grid option name, feel free to provide feedback. Will that solve all your issues?

as you can see below, I focus on the cell, then start typing "123", Enter, .... and I assume that is the behavior you were looking for

brave_5FBheS9REP

### DIFF
--- a/demos/aurelia/src/examples/slickgrid/example32.html
+++ b/demos/aurelia/src/examples/slickgrid/example32.html
@@ -90,13 +90,13 @@
         OFF
       </button>
     </div>
-    <span class="ms-2"><code>autoEditByKey</code></span>
-    <div class="btn-group" role="group" aria-label="autoEditByKey">
+    <span class="ms-2"><code>autoEditByKeypress</code></span>
+    <div class="btn-group" role="group" aria-label="autoEditByKeypress">
       <button
         type="button"
         class="btn btn-outline-secondary btn-sm"
         data-test="auto-edit-key-on-btn"
-        click.trigger="toggleAutoEditByKey(true)"
+        click.trigger="toggleAutoEditByKeypress(true)"
       >
         ON
       </button>
@@ -104,7 +104,7 @@
         type="button"
         class="btn btn-outline-secondary btn-sm"
         data-test="auto-edit-key-off-btn"
-        click.trigger="toggleAutoEditByKey(false)"
+        click.trigger="toggleAutoEditByKeypress(false)"
       >
         OFF
       </button>
--- a/demos/aurelia/src/examples/slickgrid/example32.ts
+++ b/demos/aurelia/src/examples/slickgrid/example32.ts
@@ -687,8 +687,8 @@ export class Example32 {
     this.aureliaGrid.slickGrid?.setOptions({ autoEdit: state });
   }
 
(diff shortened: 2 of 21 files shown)


zewa666:
that is exactly it. you totally nailed it. tried the same but somehow my first keystroke always got swallowed so it ended up with 23Enter in your example. awesome work 🔥

ghiscoding:
@zewa666 are you ok with the grid option name? or do you have a better name, I'm trying to be short but at the same similar to `autoEdit`

zewa666:
its perfect. I would keep it as that 👍

ghiscoding:
@zewa666 so Ben (6pac) is suggesting to use `autoEditByKeypress`, which is a bit longer but also clearer. I was trying to keep it short but at the same time clear enough, we could also use `autoEditByKeying` but it might be less accurate. See this [comment](https://github.com/6pac/SlickGrid/pull/1162#issuecomment-3537502238) from PR duplication in `6pac/slickgrid`. Probably better to get feedback from everyone

zewa666:
As written over there, I really have no strong feelings about the naming and in the last hand would also always vouch for more readability.

PR #2263, opened 2025-12-03
feat: expose the OnSearchChangeEventArgs

zewa666:
this allows consumers of the onBeforeSearchChange hook to have proper types on the event details

### DIFF
--- a/packages/common/src/services/filter.service.ts
+++ b/packages/common/src/services/filter.service.ts
@@ -24,7 +24,7 @@ import type { RxJsFacade, Subject } from './rxjsFacade.js';
 import type { SharedService } from './shared.service.js';
 import { findItemInTreeStructure, getDescendantProperty, mapOperatorByFieldType } from './utilities.js';
 
-interface OnSearchChangeEventArgs {
+export interface OnSearchChangeEventArgs {
   clearFilterTriggered?: boolean;
   shouldTriggerQuery?: boolean;
   columnId: string | number;


zewa666:
a small improvement but helps with better types. no stress on this one, doesn't matter when it gets released

ghiscoding:
ah yeah a missed export indeed. The only thing I'm waiting is the tooltip fixes and the other issue for selection copy. Do you think that you'll have a chance to look at them before the end of the month?

zewa666:
right tomorrow

ghiscoding:
@zewa666 hey so I was wondering why or how you were going to use this and I realized that the new Angular output interface (and all other frameworks) are actually using the wrong interface for that event

https://github.com/ghiscoding/slickgrid-universal/blob/e6edfb50d86b73e190862f20419d25d25eda27d1/frameworks/angular-slickgrid/src/library/components/angular-slickgrid-outputs.interface.ts#L108

https://github.com/ghiscoding/slickgrid-universal/blob/e6edfb50d86b73e190862f20419d25d25eda27d1/frameworks/slickgrid-vue/src/components/slickgridVueProps.interface.ts#L93

https://github.com/ghiscoding/slickgrid-universal/blob/e6edfb50d86b73e190862f20419d25d25eda27d1/frameworks/slickgrid-react/src/components/slickgridReactProps.ts#L103

... so I assume that we should update them all. Do you want to create another PR or do you want me to do it?

ghiscoding:
actually I made it myself since I also found out that it was using the event interface wrapper (it's not a `SlickEvent` but rather a regular PubSub event and it's not the way of using them)

zewa666:
yeah thats definitely better as I resorted to $any($event)) in tenplate and explicit cast in controller

PR #2264, opened 2025-12-04
fix: use the correct event interface for `onBeforeSearchChange`

ghiscoding:
follow-up of PR #2263 fixes 2 issues
1. wrong event interface used on the event `onBeforeSearchChange`
2. wrong event type used 
   - it's not a `SlickEvent` (used with `$event.detail.args`) but rather a PubSub Event (`$event.detail`)

cc @zewa666


zewa666:
ahh good catch, havent noticed those Framework specific exporters yet. will keep em in mind for next time

Issue #2137, opened 2025-09-08
Pasting more than 2 rows inserts blank rows


pcadec:
### Describe the bug

Dear Ghislain,

When pasting 3 or more rows, then blank rows are inserted every other row.
Errors occurs when data is copied from Excel, but also when copied from Slickgrid itself.

I fixed it locally in the file /packages/common/src/extensions/slickCellExternalCopyManager.ts

Inside  decodeTabularData(grid, clipText)
I replaced 
    const clipRows = clipText.replace('\r\n', '\n').split(/[\n\f\r](?=(?:[^"]*"[^"]*")*[^"]*$)/);
By
    const clipRows = clipText.replaceAll('\r\n', '\n').split(/[\n\f\r](?=(?:[^"]*"[^"]*")*[^"]*$)/);

Thanks for your consideration on the topic

### Reproduction

Bug occured with Chrome version 139.0.7258.155  but NOT in Firefox 142.0.1

It can be reproducted directly on this demo page : 
https://ghiscoding.github.io/angular-slickgrid-demos/#/example03

When reproducing on the demo page, just click "ok" to all "This is a required field". These alert are thrown because of blank lines inserted, which can be seen when all the alerts are closed.

### Which Framework are you using?

Angular

### Environment Info

```shell
| Executable          | Version |
| ------------------- | ------- |
| (framework used)    | ANGULAR |
| Slickgrid-Universal | 9.8.0   |
| TypeScript          | 5.5.4   |
| Browser(s)          | CHROME 139|
| System OS           | WINDOWS 11 |
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/ghiscoding/slickgrid-universal/blob/master/.github/CODE_OF_CONDUCT.md)
- [x] Read the [Wikis](https://github.com/ghiscoding/slickgrid-universal/wiki).
- [x] Check that there isn't [already an issue](https://github.com/ghiscoding/slickgrid-universal/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/ghiscoding/slickgrid-universal/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

ghiscoding:
[...]

PR #2340, opened 2026-01-29
fix: support multiple edge cases copy pasting from excel

zewa666:
This supports now:
* Newlines stripping (users request)
* Proper Newlines within quotes handling
* proper quote handling in general (jagged quoted columns)
* removal/replacement of newlines and doubleqoutes on paste

closes https://github.com/ghiscoding/slickgrid-universal/issues/2137

### DIFF
--- a/packages/common/src/extensions/slickCellExternalCopyManager.ts
+++ b/packages/common/src/extensions/slickCellExternalCopyManager.ts
@@ -231,7 +231,7 @@ export class SlickCellExternalCopyManager {
 
   protected decodeTabularData(grid: SlickGrid, clipText: string): void {
     const columns = grid.getColumns();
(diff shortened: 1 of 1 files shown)


zewa666:
ok I think I managed to grab all the cases that I've seen so far, but that doesn't mean there aren't more to come. Our Excel power-users are creative ;)

The major change here is to inspect every individual cell by itself and treat it separately.

zewa666:
also not to forget to mention that @pcadec contribution with the replaceAll was way easier than my convoluted approach. I've implemented his one here instead

zewa666:
2026-01-29_09h27_51

ghiscoding:
awesome thanks a lot, I'll push a release in the weekend :)

PR #2363, opened 2026-02-09
fix: pre-set value for editor validation on paste

zewa666:
hey there @ghiscoding 

tbh I'm not sure why but with the latest update I've realized that the paste feature does no longer work for text/date fields defined with the `editor.required: true`. From the change you can see that the validation performs the check against the previous item state, so in order to make this work I needed to apply it in advance.

I really have no clue whether this used to work in a previous verison, can't remember if we checked that scenario explicitely, but I'd assume so. Moreover it only happens for TextEditors and the DateEditor, I couldn't replicate it on Booleans. Haven't checked the other editor types.

Any ideas what could have caused this? Was there perhaps a timing issue previously that caused this to work?

### DIFF
--- a/packages/common/src/extensions/slickCellExternalCopyManager.ts
+++ b/packages/common/src/extensions/slickCellExternalCopyManager.ts
@@ -198,7 +198,7 @@ export class SlickCellExternalCopyManager {
           commitChanges: noop,
         }) as Editor;
 
-        editor.loadValue({...item, [columnDef.field]: value});
+        editor.loadValue({ ...item, [columnDef.field]: value });
         const validationResults = editor.validate(undefined, value);
         if (!validationResults.valid) {
           const activeCell = this._grid.getActiveCell()!;


zewa666:
you can replicate the issue by applying this change to Example19 in the editor config.
```ts
editor: { model: Editors.text, required: i === 5 ? true : false },
```

now copy a text from somewhere else, place the cursor into a cell of ColumnF, hit ESC to leave the editor mode and hit CTRL+V

ghiscoding:
@zewa666 ~that might be because I'm migrating to the Hybrid Selection model which is now the only selection model available~... ah wait you meant the last stable release, not the `next` branch, then I don't know

hey can you apply these changes on the `next` branch instead? I don't want to push code to the `master` branch until I release the Beta release, Then I'll merge `next` to `master` right after the Beta. There's too many conflicts every time, it's starting to be annoying

zewa666:
yeah its affecting master, but right, I'll change target to next tomorrow. also need to come back to the other two open points

ghiscoding:
sounds good, I think I'll do v10 Beta tomorrow after your PRs... no pressure haha. But even then I'm already working on a feature for v10.1, so if you can't make it then that's ok too

zewa666:
closing in favor of https://github.com/ghiscoding/slickgrid-universal/pull/2368

PR #2368, opened 2026-02-10
fix: pre-set value for editor validation on paste

zewa666:
this fixes the issue that the new value is not present on the item when pasting, thus failing certain validations like required


ghiscoding:
Thanks

PR #2369, opened 2026-02-10
feat: bind css classes to slickgrid container

zewa666:
as discussed, this adds the possibility to bind css classes directly to the container inside the angular-slickgrid component.

In which section would you document this behavior?

### DIFF
--- a/frameworks/angular-slickgrid/docs/styling/styling.md
+++ b/frameworks/angular-slickgrid/docs/styling/styling.md
@@ -3,6 +3,7 @@
 - [Using CSS Variables](#using-css-variables-instead-of-sass)
 - [Bootstrap & Other Frameworks](#bootstrap-support)
 - [SVG Icons](#svg-icons)
+- [Injecting css classes into grid container](#injecting-css-classes-into-grid-container)
 
 ### CSS/SASS Styles
 Load the default Bootstrap theme style and/or customize it to your taste (customization can applied by using SASS)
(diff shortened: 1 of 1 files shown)


ghiscoding:
> In which section would you document this behavior?

good question, the only thing I could think of is to simply add it to the "Getting Started" maybe? Otherwise that would be a grid functionality but that's probably too much to add a doc for just this. Unless we add a new doc for Templates or something and add more template stuff in it as well.

Side note the Linting fails with Prettier, you might need to run the `pnpm prettier:write`

zewa666:
dang it, forgot to run it again. will fix it tomorrow with the doc update

zewa666:
I've just noticed you've switched to OXC away from ESLint. That most likely explains why I suddenly see no warnings in VSCode for formatting issues. I had to update the vscode workspace settings to set the path to prettier. Now at least the autofix by format document works but I still don't see any warnings. Do you have any idea whats necessary besides installing the OXC extension?

ghiscoding:
yeah I switched from ESLint to oxlint and will soon also switch from Prettier to oxfmt because it's way faster. I thought that installing the oxc extension was the only thing needed but I think it might be caused by the fact that I use a Prettier plugin with oxc, not sure if there's more to configure for it to properly reflect 

https://github.com/ghiscoding/slickgrid-universal/blob/44d9a5230b9cb57f5fecbf6e7b12b8ef9f3ab69d/package.json#L103

https://github.com/ghiscoding/slickgrid-universal/blob/44d9a5230b9cb57f5fecbf6e7b12b8ef9f3ab69d/.prettierrc#L7

you could try to remove the plugin to see if that works better. With the oxc plugin it's a bit faster but not that much, so it's ok to remove it. But anyway, that will all go away when I switch to oxfmt (oxc formatter)

Also adding the docs in the styling docs is the perfect spot 👍🏻

zewa666:
ok will try that. this PR otherwise is done from my side

PR #2371, opened 2026-02-11
feat(tooltip): add global tooltip observation for non-grid elements

ghiscoding:
This PR add the possibility for the SlickCustomTooltip plugin to work outside the SlickGrid world. This is not enabled by default since the end user might have separate kind of tooltip plugins outside of the grid. However the user can enable the feature via `observeAllTooltips: true` and we can also target specific elements in the DOM via `observeTooltipContainer: 'body'` (`'body'` is the default and this option won't work without the `observeAllTooltips`). 

This PR was partially vibe coded with copilot's help

### DIFF
--- a/demos/aurelia/src/examples/slickgrid/example33.html
+++ b/demos/aurelia/src/examples/slickgrid/example33.html
@@ -33,6 +33,22 @@
   <div class="col" style="margin-bottom: 20px">
     <label for="pinned-rows">Simulated Server Delay (ms): </label>
     <input type="number" id="server-delay" data-test="server-delay" style="width: 60px" value.bind="serverApiDelay" />
+    <button
+      class="ms-2 btn btn-outline-secondary btn-sm"
+      data-test="filter-empty-desc"
+      click.trigger="setFiltersDynamically('=')"
+      title="Apply filter to show only empty descriptions"
+    >
+      Filters Empty Description
+    </button>
+    <button
+      class="ms-2 btn btn-outline-secondary btn-sm"
+      data-test="filter-non-empty-desc"
+      click.trigger="setFiltersDynamically('!=')"
+      title="Apply filter to show only non-empty descriptions"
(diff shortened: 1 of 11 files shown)


ghiscoding:
@zewa666 I found this possible feature to render Tooltip also when outside the grid. Basically I encountered this problem while implementing the upcoming menu slot renderer and in my case I have a `title` defined in a Context or Header Menu item and it didn't render the tooltip (even when imported) as I expected because the `title` is on an element that is attached on the body instead of being within or under the grid itself. So the previous implementation was that it works for all SlickGrid cells, headers, etc... but it wasn't working for something outside the grid hence this PR.

So does that PR make sense? I admit that it was vibe coded but I had to tweak it a lot to be closer to what I really wanted. Let me know if that makes sense and if there's anything to change or update... so a quick PR review might be nice before I merge it since I'm sure it might also be useful to you as well. I'm not too sure about the option names, but couldn't think of anything better at the moment

zewa666:
I think you can't call it vibe coding if you review and apply refactorings afterwards, otherwise all my work with Junior devs would be vibe coding as well 🤣 

I think this looks solid, the only thing I'm questioning is whether there is a chance for multiple instead of a single container and if we should support that?

ghiscoding:
> I think you can't call it vibe coding if you review and apply refactorings afterwards, otherwise all my work with Junior devs would be vibe coding as well 🤣

haha yeah ok that makes me feel better lol

> I think this looks solid, the only thing I'm questioning is whether there is a chance for multiple instead of a single container and if we should support that?

~but how would that happen?~ ahhh nevermind I see what you mean, just having a button+title and inner icon+title is indeed failing and requires changes. 

However this problem should be solved in a separate PR since this issue probably always existed and after asking copilot to fix it, it does bring a lot of code change. So better have 2 PRs, let me merge this one

PR #2374, opened 2026-02-12
fix(plugin): SlickCustomTooltip should work with parent+child tooltips

ghiscoding:
the tooltip wasn't behaving properly when an element with tooltip had child that also had tooltip. The event bubbling was causing some issues and `title` were also being cleared when it shouldn't in a child element

#### parent & child tooltips
msedge_MBt06Yk790

#### still works with grid cell & multiple tooltips
msedge_F9wivEZLpR


ghiscoding:
@zewa666 alright, see animated gif above and maybe do quick PR review if you want. This fixes the issue you brought up in the other PR for inner (child) element tooltips. The example shows a tooltip on button+title and child icon+title which are the inner tooltip

PR #2381, opened 2026-02-16
fix(tooltip): allow observing multiple containers with tooltips

ghiscoding:
This fixes something I missed in previous comment from @zewa666 I just re-read it and now understood what you really meant, lol. Because earlier I thought you were talking about inner child tooltips (e.g. button+tooltip and inside it an icon+tooltip which wasn't working and I fixed that in PR #2374)... but in reality you just mean watching multiple containers, right?

> I think this looks solid, the only thing I'm questioning is whether there is a chance for multiple instead of a single container and if we should support that?

I simply had to switch from `.querySelector()` to `.querySelectorAll()` and ask the user to provide a CSV string to cover multiple containers. I can do that because of my built-in Binding Event Service that accepts `Element | Element[]` for binding and that's it, it just works :)

For example: 
```ts
customTooltip: { 
  observeAllTooltips: true, 
  observeTooltipContainer: '.grid-container, .grid-container2' 
}
```

### DIFF
--- a/docs/grid-functionalities/custom-tooltip.md
+++ b/docs/grid-functionalities/custom-tooltip.md
@@ -13,6 +13,7 @@
 - [How to delay the opening of a tooltip?](#how-to-delay-the-opening-of-a-tooltip)
   - [delay a tooltip with Formatter](#delay-a-tooltip-with-formatter)
   - [delay a Regular Tooltip](#delay-a-regular-tooltip)
+- [Tooltips Outside the Grid](#tooltips-outside-the-grid)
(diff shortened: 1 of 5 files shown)


ghiscoding:
@zewa666 so does that PR cover what you brought up earlier?

zewa666:
yeah I think this should exactly nail it

ghiscoding:
Great and BTW you can see it live now
https://ghiscoding.github.io/angular-slickgrid-demos/#/example33

PR #2527, opened 2026-04-14
fix: properly update lastSearchValue to support presetted filter cleanup

zewa666:
If the app startsup with presets (e.g from localStorage) and these contain filter presets for CompoundInputs, deleting all of them (CTRL+A; Backspace) won't trigger Backend Service re-fetches due to wrong undefined<>'' checks.

### DIFF
--- a/packages/common/src/filters/__tests__/compoundInputFilter.spec.ts
+++ b/packages/common/src/filters/__tests__/compoundInputFilter.spec.ts
@@ -251,6 +251,22 @@ describe('CompoundInputFilter', () => {
     expect(callbackSpy).toHaveBeenCalled();
   });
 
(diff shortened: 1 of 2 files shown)


zewa666:
I held off creating this PR earlier as I wanted to make sure its present with the latest version. Since I'm having time to update our app to the latest NG and Slickgrid-Universal/Angular version, now its confirmed to still exist

ghiscoding:
> deleting all of them (CTRL+A; Backspace)

I'm not sure that I understand, what do you mean by "deleting all of them"? is that clearing the compound dropdown? or deleting all filters? I can't picture it...

zewa666:
if you startup prefiltered, click into the fipter input and hit ctrl+a and backspace

ghiscoding:
ok thanks that seems fine and I guess it does make sense to always save the last search value

PR #2531, opened 2026-04-16
feat: add `autoScrollOnColumnResize` grid option (default true)

ghiscoding:
Defaults to true, auto-scrolls horizontally to keep the resized column visible (last column scrolls to max for legacy compatibility).

This was related to a PR opened in MSSQL extension, they added the feature through a SlickGrid plugin but I can do it directly in the core with less than 20 lines.

~**Note** this only applies to the last column resizing.~ actually let's make it auto-scroll for any column that goes outside the viewport on the right (not just on last column)

firefox_nkPL3ufTpU

### DIFF
--- a/packages/common/src/core/__tests__/slickGrid.spec.ts
+++ b/packages/common/src/core/__tests__/slickGrid.spec.ts
@@ -4641,8 +4641,8 @@ describe('SlickGrid core file', () => {
       grid = new SlickGrid<any, Column>(container, data, columns, { ...defaultOptions, forceFitColumns: false });
       grid.init();
 
-      // Simulate state for the branch: _isResizingLastColumn && hScrollDist && !vScrollDist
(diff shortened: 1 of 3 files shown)


ghiscoding:
cc @zewa666 @jr01

zewa666:
thats a nice one. while at it, on larger grids, if you reorder columns and reach the screens/grids border, it would be nice if that could scroll as well. e.g moving the last column to first position

ghiscoding:
> thats a nice one. while at it, on larger grids, if you reorder columns and reach the screens/grids border, it would be nice if that could scroll as well. e.g moving the last column to first position

not really sure what you mean but feel free to open a new PR, I externalized the scroll as `scrollToX()` not long ago, so that is basically what we can use to scroll all divs at the same time (preheader, header, data, footer, ...)

PR #2542, opened 2026-04-21
fix(regression): excelcopybuffer columnvisibility behavior changed in v10

### LINKED ISSUE #2541: ExcelCopyBuffer Cell Selection does not respect hidden columns
### Describe the bug

The getCellFromPoint() from the ExcelCopyBuffer, which drives its own HybridSelectionModel, iterates
`this.columns` (which in v10 includes hidden columns) to accumulate widths and find the cell index. Since these are not rendered it causes an offset when selection cells for a lets say next copy action. (see the repro gif)

### Reproduction

Image

Angular

### Environment Info

```shell
| (framework used)    | latest |
```

### Validations

zewa666:
fixes both the copy/paste behavior and the cell selection not respecting the new hidden columns

fixes https://github.com/ghiscoding/slickgrid-universal/issues/2541

### DIFF
--- a/packages/common/src/core/__tests__/slickGrid.spec.ts
+++ b/packages/common/src/core/__tests__/slickGrid.spec.ts
@@ -6986,14 +6986,14 @@ describe('SlickGrid core file', () => {
         expect(result5).toEqual({ cell: -1, row: -1 });
       });
 
-      it('should return hidden cells as well since skipping them would add an unwanted offset', () => {
+      it('should skip hidden columns and return last visible column cell when x coordinate falls beyond visible canvas', () => {
         grid = new SlickGrid<any, Column>(container, data, columns, { ...defaultOptions, enableCellNavigation: true });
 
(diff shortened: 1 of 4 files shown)


ghiscoding:
oops sorry, I thought I was done with regressions caused by the hidden columns, hopefully the last one. Thanks again, it looks all good :)

zewa666:
I'm sure some hidden gems are still around. good news is that we have 2 guys helping out testing the latest update, hence we're running quickly into those and can get em fixed

ghiscoding:
yeah that's good, we could also add more Cypress tests, but it's not the easiest to do for cell selection and dragging. Perhaps someday I'll migrate to Playwright (but Cypress is still easier to write)

Are you expecting to look at your other issue related to tree depth? I don't need it, but just curious of timeline.

Side note I'll push another version tonight, I had feedback from the MSSQL maintainers that required a fix for the column scroll (they also wanted scrolling even when falling outside the browser view). So I'm releasing it after working hours

PR #2544, opened 2026-04-22
fix: use absolute instead of running index to count position

zewa666:
sry this issue somehow sneaked in during testing. ofc the returned indexes should be the absolute ones from the total collection

### DIFF
--- a/packages/common/src/core/slickGrid.ts
+++ b/packages/common/src/core/slickGrid.ts
@@ -6384,7 +6384,7 @@ export class SlickGrid<TData = any, C extends Column<TData> = Column<TData>, O e
         continue;
       }
       w += this.columns[i].width as number;
-      cell++;
+      cell = i + 1;
     }
     cell -= 1;
 
--- a/packages/common/src/core/slickGrid.ts
+++ b/packages/common/src/core/slickGrid.ts
@@ -6380,11 +6380,10 @@ export class SlickGrid<TData = any, C extends Column<TData> = Column<TData>, O e
 
     let w = 0;
     for (let i = 0; i < this.columns.length && w <= x; i++) {
-      if (!this.columns[i] || this.columns[i].hidden) {
-        continue;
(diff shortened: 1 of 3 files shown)


ghiscoding:
is that another thing related to hidden column? because this function wasn't changed in a very long time

zewa666:
oh its not so long ago, actually just a day 🤣 sry if it was unclear but I contributed the `cell++` whereas it should have been the actual absolute index `cell + 1`. But wait a bit please I sadly found another new bug

ghiscoding:
ah ok yeah I didn't realized it was you in previous PR, but you changed it to a `continue` yesterday, can we write it like this instead:

https://github.com/ghiscoding/slickgrid-universal/blob/da8e9d62a6d5aeb394bd8cf2b440013369156d23/packages/common/src/core/slickGrid.ts#L2936-L2939

it's less lines of code but it's also less problems test coverage (and maybe less readable too but it's not that bad)

zewa666:
so one more issue that I found was that if the to be pasted columns exceed amount of availble columns, which gets triggered when pasting multiple rows as the same time, we ran into a undefined.hidden error.

zewa666:
I think we're good with this one so far, haven't found any new issues as of today

ghiscoding:
yeah but can you make the change as suggested in my previous comment, you're also missing coverage for that new file mod... oh and formatting problem too, I'm not sure why Prettier vscode plugin no longer detects formatting changes (I won't switch to oxc/ofmt until they cover all inline templating aka js-in-html)

zewa666:
yep ofc, will take a look today

ghiscoding:
@zewa666  Great thanks again, are you done with all your testing or does that last a few days? I could push another version in the weekend or wait longer

zewa666:
I think I'm good for this week. deploying by the weekend is totally fine. I'm sure we'll find more but that is an ongoing task

PR #2549, opened 2026-04-24
fix: row selection not sorted numerically

ghiscoding:
copies a PR that was brought in SlickGrid repo

> Rows returned from `onSelectedRowsChanged` events are not numerically sorted but rather lexicographically, e.g. `[0, 1, 10, 11, .. ]`.

### DIFF
--- a/packages/common/src/core/slickGrid.ts
+++ b/packages/common/src/core/slickGrid.ts
@@ -3474,7 +3474,8 @@ export class SlickGrid<TData = any, C extends Column<TData> = Column<TData>, O e
     }
 
     // check if the selected rows have changed (index order isn't important, so we'll sort them both before comparing them)
-    if (!this.arrayEquals(previousSelectedRows.sort(), this.selectedRows.sort())) {
+    const sortNumbers = (a: number, b: number) => a - b;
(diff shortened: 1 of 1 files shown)


ghiscoding:
it turns out that it's much complicated than originally taught and can cause some serious out of sync issues, so I think it's better to just cancel this and leave the sort to the end user as I answered in the SlickGrid repo

```
tested that in my own project and doing this change can potentially cause other arrays to become out of sync (mainly the Ids are becoming out of sync). So if you use the DataView then these become out of sync, meaning that row indexes no longer match row ids:
- `dataView.mapRowsToIds()`
- `dataView.getAllSelectedIds()`
- `dataView.getAllSelectedFilteredIds()`

We could have them all sort by their ids so they match the row indexes with row ids, that would work if ids are numbers or strings as number... but what if the ids are GUID or other type of strings... then we're out of luck.

So having all of these things out of synch is seriously not ideal and in the end it might be better to just leave that to the end user (like I mentioned before)...

OR the best we could do is maybe  to provide an extra argument to the get function to optionally sort but leave it as unsorted by default (so it would still match the ids by default)
```ts
getSelectedRows(sortRows = false) {
  ...
  const selectedRows = this.selectedRows().slice(0);
  if (sortRows) {
    return selectedRows.sort((a, b) a - b);
  }
  return selectedRows;
}
```
```

PR #2552, opened 2026-04-26
feat: add Grouping/TreeData option to `toggleOnNodeTitle`

ghiscoding:
add a new option named  ~`toggleByCellClick`~ `toggleOnNodeTitle` that allows clicking on the Tree Data and/or Grouping node title to toggle it. This can be helpful since some user, like me, might find the toggle icon a little too small to click. The new option won't be enabled by default and will keep the same original behavior (not everyone would want this, hence its default disabled state)

- Tree Data: `gridOptions = { treeDataOptions: { toggleOnNodeTitle: true, }}`
- Grouping: `gridOptions = { groupItemMetadataOption: { toggleOnNodeTitle: true, }}`

#### TODO
- [x] Not sure if I should revisit this and maybe change it to `toggleOnNodeTitle` since cell click might interfere with other processes like inline cell editing
  - yup this is better since that won't conflict with inline editing
- [x] forgot to add docs for the new option in both Tree Data and Grouping docs

[Screencast_20260427_222730.webm](https://github.com/user-attachments/assets/29836fc7-1034-4048-b267-9d33046e3abd)

### DIFF
--- a/.github/workflows/pkg-pr-new.yml
+++ b/.github/workflows/pkg-pr-new.yml
@@ -32,4 +32,14 @@ jobs:
         run: pnpm build
 
       - name: Publish
-        run: pnpm dlx pkg-pr-new publish --no-template --pnpm './packages/*' './frameworks/*' './frameworks-plugins/*'
+        run: |
+          # Create tarballs in the correct location for pkg-pr-new
+          for dir in ./packages/* ./frameworks/* ./frameworks-plugins/*; do
+            if [ -d "$dir" ] && [ -f "$dir/package.json" ]; then
+              echo "Packing $dir"
+              cd "$dir"
+              npm pack ./dist
+              cd - > /dev/null
+            fi
+          done
+          pnpm dlx pkg-pr-new publish --no-template --pnpm './packages/*' './frameworks/*' './frameworks-plugins/*'
--- a/.github/workflows/pkg-pr-new.yml
+++ b/.github/workflows/pkg-pr-new.yml
@@ -38,7 +38,12 @@ jobs:
             if [ -d "$dir" ] && [ -f "$dir/package.json" ]; then
               echo "Packing $dir"
               cd "$dir"
-              npm pack ./dist
+              # Check if package has publishConfig.directory set to dist
+              if grep -q '"directory":\s*"dist"' package.json; then
+                npm pack ./dist
+              else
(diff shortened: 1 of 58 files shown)


ghiscoding:
cc @zewa666 any feedback before I merge this one?

zewa666:
hey there, I remember we had this thingy where a PR would create a demo site built with that commit in a sandbox for testing. Is that still a thing?

Things I'd like to test would be:
* Wheter selecting the title would cause toggling
* Behavior of the spacebar key on the cell with CellNavigation
* Marking multiple cells, starting with the title cell

ghiscoding:
> hey there, I remember we had this thingy where a PR would create a demo site built with that commit in a sandbox for testing. Is that still a thing?

@zewa666  sadly not anymore because Stackblitz is stuck at pnpm 8 and fails for pnpm 9+ (pnpm 11 came out today), which is why I removed the broken button (the other buttons I have in all readmes are ok because I use npm in all demo repos). That's why I added a video above


> Things I'd like to test would be:
> 
> 1 Wheter selecting the title would cause toggling
> 2 Behavior of the spacebar key on the cell with CellNavigation
> 3 Marking multiple cells, starting with the title cell

1. it's through the toggle icon and/or title **but only** when you enable the new option (I defaulted it to false since not everyone would want it I think). My initial PR was on the cell click but decided to change it to title only since clicking is also used for inline editing and a cell can be much wider than the node title, so I limited it to the icon + title (again only when enabled). It simply looks for `.slick-[group|tree]-toggle` and `.slick-[group|tree]-title`
2. you can use spacebar or left/right arrows, that was done last month in PR #2448
3. not really sure what you mean, this PR is just about widening the area to toggle the Tree or Grouping, they both have their own config (same name but in different grid option prop location, I updated description above to show how to use them)

zewa666:
the reason was that if I start any of the mentioned actions, as a side effect a toggling could happen

ghiscoding:
@zewa666 I still don't see how unless you mean that PR #2448 has negative side effect? But my PR is not related to that at all and is also optional so I don't see how that could impact users. However, I did see that inline editor could have been a concern, so I did change PR to avoid conflicting with inline editing and limited my approach to only group/tree title (which I assume is what most users will want and expect, I personally want this and so I'll enable this as a global setting in my project)

PR #2563, opened 2026-05-04
feat(core): add optional formatted data cache for large dataset exports and rendering

ghiscoding:
_full implementation vibe coded with Copilot and Claude Sonnet 4.6_

## Summary
This PR introduces an opt-in formatted data cache pipeline to improve export throughput and post-warmup grid rendering responsiveness on large datasets.

It adds asynchronous background cache population in `SlickDataView`, keyed by item id, for:
- export formatted values (`formattedDataCache`)
- cell display formatter output (`formattedCellCache`)

The pipeline uses batched, time-budgeted processing (`MessageChannel` with `requestAnimationFrame` fallback), supports row-level and full cache invalidation on data changes, and preserves existing cache-disabled/cache-miss behavior.

`SlickGrid` now reads from the display cache in formatter flow when safe, and Excel/PDF/Text export paths use cache-first reads with fallback to live formatter logic so sanitize/decode/parsing behavior remains consistent.

## Architecture note
The implementation is planner-based:
- `SlickGrid` provides a single formatted cache planner callback
- `SlickDataView` uses that planner to classify columns and build cache context
- Export services consume the shared cache (no per-service consumer registration lifecycle)

## Validation
Validation was done with focused unit tests for:
- cache disabled vs cache miss vs cache hit parity
- sanitize/decode-sensitive output paths
- planner wiring and refresh behavior
- async batching and fallback scheduling paths

Tests and coverage are green on this branch.

## Why this helps
This specifically targets formatter-heavy, large-grid workloads like the scenario discussed here:
https://github.com/ghiscoding/slickgrid-universal/discussions/1922

## Impact
- Faster exports on large datasets (Excel/PDF/Text), especially with many formatted columns
- Smoother grid rendering after warmup by reusing cached formatter outputs
- Background population keeps UI responsive during warmup
- No functional output change intended: cache-miss and cache-disabled paths still use existing logic
- Backward compatible: feature is opt-in and disabled by default
- Tradeoff: higher memory usage, so enable selectively for large/formatter-heavy grids (commonly 10K+ rows)

#### Technical implementation
See [FORMATTED_DATA_CACHE_IMPLEMENTATION.md](https://github.com/ghiscoding/slickgrid-universal/blob/3d1a67ea41a3c74c158619e80bce8fa56a31ff03/.github/tasks/FORMATTED_DATA_CACHE_IMPLEMENTATION.md)

---

### Live Demo Test ([Example 03](https://ghiscoding.github.io/slickgrid-universal/#/example03) with 500K rows)

Excellent results! 🎉 That's a massive validation of the implementation:

**Performance Summary (500K rows, 8 columns):**
- **Cache Population**: ~13 minutes (779,927 ms) for 4M cells
  - Runs in background with MessageChannel + requestAnimationFrame batching
  - UI stays responsive during population
- **Export Time**: 30-35 seconds (consistent regardless of grouping/sorting)
  - Fast because cache hits eliminate formatter overhead
  - DataView's built-in filtering handles visible rows automatically
- **Previous Behavior**: Browser hang (unusable)

**What makes this successful:**
1. ✅ Cache-first reads in export services (formatter → cache fallback)
2. ✅ DataView cleanup on destroy (`clearFormattedDataCache()`)
3. ✅ Export service cleanup on dispose (prevents memory leaks)
4. ✅ Planner-based architecture (single callback, selective caching)
5. ✅ Batch scheduling with time budgets (keeps UI responsive)

The cache is doing exactly what it's designed for—trading a one-time 13-minute background population for consistent 30-35 second exports on huge datasets. Without the cache, that export would likely take 5+ minutes or hang.


ghiscoding:
@zewa666 @jr01 are you guys dealing with very large dataset or always using a Backend Service like OData to avoid large local dataset? If you're always using Backend Service then I guess this PR won't change anything for you, but just curious... 

I know that this could help a lot for use case like @kevinburkett had in #1922 

the PR does add 230 new lines of code (that is what codecov says anyway) but I think it's totally worth having it in the core (SlickDataView) and it's opt-in, so user will have to enable it themselves (mainly because that will consume memory and takes time to populate the cache but doesn't impact UI responsiveness, that was an important condition: build the cache without impacting UI responsiveness). On the [Example02](https://ghiscoding.github.io/slickgrid-universal/#/example02) with 8 columns and 50K rows, it takes 23sec to run the cache. I even tried with 500K in Example 03 and it takes a lot longer to build the cache but once it's available then the export no longer freezes the browser

I basically thought of doing this caching to decrease export time, and this can always be used to improve rendering time as well for large and very large dataset (build the cache and use it whenever possible). For export services, that is definitely a huge gain in perf (in other words a big drop in time it takes to export but only after the cache is ready, no need to reparse formatters every time). Why would we have very large dataset? I guess Grouping or Tree Data could be a factor (since those don't work with Backend Services)

zewa666:
we're strictly with paginated views (20-50) eith backend datasets. but the treeview still gets quite crowded as that one is an exception

ghiscoding:
yeah I guess that will be nice for only a limited subset of users, but still great to add in the core nonetheless, at least I would think so :)

packages/common/src/core/slickDataview.ts
```diff
@@ -568,6 +689,23 @@ export class SlickDataView<TData extends SlickDataItem = any> implements CustomD
       this.updateSingleItem(ids[i], newItems[i]);
     }
     this.refresh();
+
+    // Invalidate the formatted cache for every updated item
+    const gridOptions = this._gridOptions;
+    if (gridOptions?.enableFormattedDataCache) {
```
ghiscoding: we can probably merge the `updateItem()` and `updateItems()` code

packages/common/src/core/slickDataview.ts
```diff
@@ -675,6 +816,11 @@ export class SlickDataView<TData extends SlickDataItem = any> implements CustomD
       // update lookup from front to back
       this.updateIdxById(indexesToDelete[0]);
       this.refresh();
+      // Clean up cache entries for all deleted items
+      for (let i = 0, l = ids.length; i < l; i++) {
+        delete this.formattedDataCache[ids[i]];
+        delete this.formattedCellCache[ids[i]];
```
ghiscoding, line 766: same as above we can probably merge the `deleteItem()` and `deleteItems()` cache code

packages/common/src/core/slickDataview.ts
```diff
@@ -1622,6 +1768,362 @@ export class SlickDataView<TData extends SlickDataItem = any> implements CustomD
...
+    }
+    scheduleNextBatch();
+  }
+
+  /** Builds a RowCacheContext from the current grid state. Called once per population run or per single-row invalidation. */
+  protected buildCacheContext(): RowCacheContext {
+    const grid = this._grid!;
+    const gridOptions = this._gridOptions ?? grid.getOptions();
+    const columns = grid.getColumns() ?? [];
+    const exportOptions = gridOptions.excelExportOptions ?? gridOptions.textExportOptions;
```
ghiscoding: hmm I didn't expect to see this here, SlickDataView doesn't have any knowledge of Export Services, we shouldn't go into a circular dependency

PR #2570, opened 2026-05-06
fix: restore individual cell values on undo for multi-cell paste

ghiscoding:
**fixes an old issue opened in SlickGrid (6pac) project: https://github.com/6pac/SlickGrid/issues/917**

## Description
Fixes issue mentioned in SlickGrid, where undoing a multi-cell paste operation in `CellExternalCopyManager` incorrectly restores all cells to the same value instead of their individual previous values.

## The Problem Explained

The `oneCellToMultiple` paste feature allows users to copy a single cell and paste it to multiple cells at once (like Excel). However, when undoing this operation, the bug causes **all pasted cells to be restored to the same previous value** instead of each cell being restored to **its own individual previous value**.

## Root Cause
When pasting a single cell value to multiple cells (`oneCellToMultiple`), the undo function was incorrectly retrieving the old value. It used `oldValues[0][0]` for all cells instead of `oldValues[y][x - xOffset]`, causing all cells to be restored to the first cell's previous value.

Additionally, the undo function was not properly tracking `xOffset` for hidden columns, unlike the execute function which handles this correctly.

## Detailed Reproduction Steps

1. **Setup Initial Grid Values:**
   - Cell A1 = "Apple"
   - Cell B1 = "Book", B2 = "Car", B3 = "Dog", B4 = "Egg"

2. **Copy Single Cell:**
   - Click on cell A1 (contains "Apple")
   - Press Ctrl+C to copy

3. **Paste to Multiple Cells:**
   - Select the range B1:B4 (all 4 cells)
   - Press Ctrl+V to paste
   - **Expected Result:** All cells now show "Apple" ✓
   - Cells now contain: B1="Apple", B2="Apple", B3="Apple", B4="Apple"

4. **Undo the Paste:**
   - Press Ctrl+Z to undo
   - **Expected Result (FIXED):** Each cell should restore to its original value
     - B1 should be "Book", B2 should be "Car", B3 should be "Dog", B4 should be "Egg"
   - **Previous Buggy Behavior:** All cells would incorrectly become "Book" (the first cell's previous value)

## Solution
- Removed the flawed `oneCellToMultiple` conditional from the undo function
- Added proper `xOffset` tracking to handle hidden columns correctly (consistent with the execute function)
- Each cell now restores from its own stored old value: `oldValues[y][x - xOffset]`

## Impact
- ✓ Fixes undo behavior for multi-cell paste operations
- ✓ No breaking changes
- ✓ Properly handles hidden columns in the restore process
- ✓ Makes undo behavior consistent with user expectations (each cell returns to its previous value)

### DIFF
--- a/demos/vanilla/src/examples/example19.html
+++ b/demos/vanilla/src/examples/example19.html
@@ -30,6 +30,19 @@
   </p>
 </h5>
 <h6 class="title is-6">
+  <button class="button is-small" onclick.trigger="undoLastEdit()" data-test="undo-last-edit-btn">
+    <span class="mdi mdi-undo"></span>
+    <span>Undo Last Edit</span>
+  </button>
+  <button class="button is-small" onclick.trigger="undoLastEdit(true)" data-test="undo-open-editor-btn">
+    <span class="mdi mdi-undo"></span>
+    <span>Undo Last Edit &amp; Open Editor</span>
+  </button>
+  <button class="button is-small" onclick.trigger="undoAllEdits()" data-test="undo-all-edits-btn">
+    <span class="mdi mdi-history"></span>
+    <span>Undo All Edits</span>
+  </button>
+  </button>
   <button class="button is-small is-primary" onclick.trigger="togglePagination()" data-text="toggle-pagination-btn">
     <span class="mdi mdi-swap-vertical"></span>
     <span>Toggle Pagination</span>
--- a/demos/vanilla/src/examples/example19.html
+++ b/demos/vanilla/src/examples/example19.html
@@ -42,7 +42,6 @@
     <span class="mdi mdi-history"></span>
     <span>Undo All Edits</span>
   </button>
-  </button>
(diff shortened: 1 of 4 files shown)


ghiscoding:
@zewa666 would you mind providing your point of view to this PR, I literally looked at all the old opened SlickGrid issues and vibe coded fixes for a few of them (using copilot for free). Then looked at reimplementing it in here in Slickgrid-Universal at work with copilot again (it also confirmed that we also have that problem here too). 

It's supposed to fix issue mentioned in https://github.com/6pac/SlickGrid/issues/917 and I asked copilot to provide all the steps to reproduce.

ghiscoding:
I actually had to add the "Undo" and Edit Queue functionality in the Example 19 to test it out, and it does seem to work as the fix suggest. Basically in 1 "Undo", it should revert all cell range to their original values

msedge_Aop5EzZu2I

zewa666:
oh thats a very good catch. will check it out tomorrow

zewa666:
sry intense week at work and home, I'm sure this will be fine, otherwise I can add a PR fix if I cam across any issues when installing the new version

PR #2573, opened 2026-05-07
fix: respect `multiSelect` in checkbox and row keyboard selection

ghiscoding:
**fixes an old issue opened in SlickGrid (6pac) project: https://github.com/6pac/SlickGrid/issues/345**

## Summary
This change fixes checkbox row selection behavior so it correctly follows the grid `multiSelect` option.

- When `multiSelect` is true:
  - Checkbox clicks keep existing toggle behavior (add/remove row from selection).
- When `multiSelect` is false:
  - Checkbox clicks enforce single selection (only one row selected at a time).
  - Clicking an already selected checkbox clears selection.
- Active cell is still updated after checkbox toggle to preserve keyboard/editor flow.
- Row-mode Shift+Arrow selection in the hybrid selection model now clamps to a single row when `multiSelect` is false.

## Root cause
Checkbox toggling previously always used additive selection logic, even when `multiSelect` was disabled.  
Also, row keyboard range selection had logic gaps for single-select mode in the merged hybrid selection model.

## Repro steps (before fix)
1. Open a grid with checkbox selector enabled and set `multiSelect` to false.
2. Click checkbox on row A.
3. Click checkbox on row B.
4. Observe both rows can remain selected (incorrect for single-select mode).

Optional keyboard repro:
1. Keep `multiSelect` false.
2. Enter row selection mode.
3. Use Shift+Arrow Up/Down.
4. Observe range-style behavior instead of single-row behavior.

## Expected behavior (after fix)
1. With `multiSelect` false, selecting row B after row A keeps only row B selected.
2. Clicking an already selected checkbox clears selection.
3. Shift+Arrow in row mode keeps selection to one row.
4. Active cell remains synchronized on checkbox toggle.

## What was changed
- Updated checkbox toggle logic to branch on `multiSelect` :
  - true: preserve existing toggle semantics.
  - false: enforce `[row]` or `[]` selection only.
- Kept active-cell update after toggle.
- Updated hybrid row keyboard handling so Shift+Arrow still works, but clamps to a single row when `multiSelect` is false.
- Added/updated unit tests for:
  - checkbox single-select behavior
  - keyboard single-select clamp behavior

## Validation
- Unit tests pass.
- Coverage checks pass.
- Cypress run in progress by maintainer.

I'll need to add a note warning on the next release:

**⚠️ Breaking Change - `multiSelect` Option Now Properly Enforced**

The `multiSelect` option is now properly respected. If you previously set `multiSelect: false` expecting it to work differently, this will now correctly enforce single-row-only selection. To allow multiple row selection while preventing the currently-active row from being automatically selected during navigation, use `selectionOptions.selectActiveRow: false` instead (keeping `multiSelect` at its default `true`).

---

This makes it clearer that `selectActiveRow: false` is specifically about navigation behavior (not auto-selecting as you arrow through rows), whereas `multiSelect` is about how many rows you can select in total.

### DIFF
--- a/demos/aurelia/src/examples/slickgrid/example30.ts
+++ b/demos/aurelia/src/examples/slickgrid/example30.ts
@@ -497,7 +497,6 @@ export class Example30 {
       preHeaderPanelHeight: 28,
       enableCheckboxSelector: true,
       enableSelection: true,
-      multiSelect: false,
       checkboxSelector: {
         hideInFilterHeaderRow: false,
         hideInColumnTitleRow: true,
--- a/demos/react/src/examples/slickgrid/Example30.tsx
+++ b/demos/react/src/examples/slickgrid/Example30.tsx
@@ -492,7 +492,6 @@ const Example30: React.FC = () => {
       preHeaderPanelHeight: 28,
       enableCheckboxSelector: true,
       enableSelection: true,
-      multiSelect: false,
       checkboxSelector: {
         hideInFilterHeaderRow: false,
         hideInColumnTitleRow: true,
--- a/demos/vue/src/components/Example30.vue
+++ b/demos/vue/src/components/Example30.vue
@@ -456,7 +456,6 @@ function defineGrid() {
     preHeaderPanelHeight: 28,
     enableCheckboxSelector: true,
     enableSelection: true,
-    multiSelect: false,
     checkboxSelector: {
       hideInFilterHeaderRow: false,
       hideInColumnTitleRow: true,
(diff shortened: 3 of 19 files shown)


ghiscoding:
@zewa666 @jr01 this PR is another one of those vide coded fix by copilot for super old SlickGrid (6pac repo) issues, the issue was legit and the fix is now displaying the correct behavior. This might surprise some users, for example I was wrongly setting `multiSelect: false` in my Composite Editor demos and it now failed my tests, the new behavior is correct and it can't be disabled on Composite Editor because it does need multi-row selection (mass selection)... so FYI I'll be adding this warning:

> **⚠️ Breaking Change - `multiSelect` Option Now Properly Enforced**
> 
> The `multiSelect` option is now properly respected. If you previously set `multiSelect: false` expecting it to work differently, this will now correctly enforce single-row-only selection. To allow multiple row selection while preventing the currently-active row from being automatically selected during navigation, use `selectionOptions.selectActiveRow: false` instead (keeping `multiSelect` at its default `true`).
> 
> ---
> 
> This makes it clearer that `selectActiveRow: false` is specifically about navigation behavior (not auto-selecting as you arrow through rows), whereas `multiSelect` is about how many rows you can select in total.

zewa666:
thx a bunch for the heads up

PR #2591, opened 2026-05-20
feat(TreeData): add maxVisibleDepth option + runtime API

### LINKED ISSUE #2373: Set max visible depth of treeview

Hey there, one of our tools, which currently still use the super old Kendo Grid, will be replaced with Slickgrid. One of the features it had though (granted it was a custom made one) is to limit the visible depth of the tree.

So I thought it might be a fitting feature for the Treeview as well

### Suggested solution

There are a few things to consider here, e.g like how the Export or FilterService is affected, but we can go into detail on this later on.

Here's an example I came up with:

Image

### Alternative

_No response_

### Additional context

_No response_

### Validations

- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

ghiscoding:
closes #2373

_vibe coded with copilot using GPT-5 mini_

Adds a new `treeDataOptions.maxVisibleDepth` setting to limit how deep tree nodes are shown (nodes deeper than the value are hidden, not removed). Adds runtime API methods `setMaxVisibleDepth(n)` and `clearMaxVisibleDepth()` so apps can update or clear the limit at runtime. Updates demos and docs across frameworks and adds simple Set/Clear controls to the Example05/Example27 demos.

msedge_gcKnmRy7Ah

### DIFF
--- a/demos/aurelia/src/examples/slickgrid/example27.html
+++ b/demos/aurelia/src/examples/slickgrid/example27.html
@@ -105,6 +105,35 @@
       <span>Log Hierarchical Structure</span>
     </button>
   </div>
+  <div class="d-inline-block ms-2 mt-2">
+    <div class="input-group input-group-sm mb-0">
+      <input
+        id="maxVisibleDepthInput"
+        class="form-control"
+        type="number"
+        placeholder="Max Visible Depth (e.g. 1)"
+        aria-label="Max Visible Depth"
+      />
+      <button
+        class="btn btn-outline-secondary btn-xs btn-icon"
+        data-test="set-max-visible-depth-btn"
+        type="button"
+        click.trigger="setMaxVisibleDepthFromInput()"
+      >
+        <span class="mdi mdi-check" aria-hidden="true"></span>
+        Set
+      </button>
+      <button
+        class="btn btn-outline-secondary btn-xs btn-icon"
+        data-test="clear-max-visible-depth-btn"
+        type="button"
+        click.trigger="clearMaxVisibleDepth()"
+      >
+        <span class="mdi mdi-close" aria-hidden="true"></span>
+        Clear
+      </button>
+    </div>
+  </div>
 </div>
 
 <br />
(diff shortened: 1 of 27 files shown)


ghiscoding:
@zewa666 this should address your feature request from issue #2373

at first it only added the tree data options and modified the Filter Service, but then it required more work for the end user to change it afterward. So I added 2 functions to set/clear the max depth via the Tree Data Service, which makes it much easier for the end user to change it at anytime. 

Let me know if that covers it all, a big part of it was done by copilot

it's always a bit longer to review a PR now that we have all frameworks in here haha 😅 but it's good to have the same code change spread across all framework demos.... anyway if you could just review the 2 Services changes, that'd be great, thanks

...oh I forgot to test the export, which is a manual process, but I expect it to work since it's WYSIWYG but will check tomorrow nonetheless

EDIT

tested the export and it works as intended which isn't surprising since it's WYSIWYG

zewa666:
this is exactly what I had in mind with my first shot. it looks great and achieves exactly whats needed for huge nested treeviews.

one thing I noticed though is that the setting could be added to state changes so that they survive a refresh if persisted in localstorage

aside from that lgtm

ghiscoding:
> one thing I noticed though is that the setting could be added to state changes so that they survive a refresh if persisted in localstorage

@zewa666  you mean as a Grid State? That would require a few more changes (add it to Grid State, create an event to save it whenever it changes and do this in all frameworks implementation, re-apply on grid load as Grid Presets, ...), is that really expected to change that often compared to what you originally set in the grid options?
```ts
this.gridOptions = { 
  treeDataOptions: {
    maxVisibleDepth: 2,
  }
}
```

personally I think this might be better for the end user (you) to do it themselves since I'm assuming that this tree depth would be used pretty rarely isn't it (maybe just you haha)?. Perhaps I could add a PubSub event when it changes though so you can hook a save?

zewa666:
sry I was unclear, yeah exactly the additional pub sub event would be what I'd look for. I did the same for when columns get toggled visibility through the menu. that would help a lot

ghiscoding:
done, I added `onTreeMaxVisibleDepthChanged` with a depth number argument or `undefined` when calling the clear function... is that ready for a merge now?

Side note, PR #2601 is also another awesome perf improvement for Tree Data thanks to AI

PR #2601, opened 2026-06-08
perf: drop `un-flatten-tree` dep & optimize Tree Data utils by ~3x perf

ghiscoding:
_vibe coded with copilot using Claude Sonnet 4.6_

### Summary

Removes the `un-flatten-tree` third-party dependency from `@slickgrid-universal/common` and rewrites or optimizes all tree data utility functions for significantly better performance, particularly at scale (50k–250k nodes). This makes the Tree Data to parse about ~3x faster when converting from flat to tree and vice versa (flatten/unflatten) which is how Tree Data was implemented in SlickGrid (when using Tree Data in SlickGrid we always keep a flat data copy and its equivalent tree data).

---

#### package.json
- Removed `"un-flatten-tree"` from dependencies

#### utilities.ts

**`flattenToParentChildArray`** (was previously using `un-flatten-tree` dependency)
- Replaced the library call with a custom **O(n) iterative DFS** using two parallel stacks (`nodeStack` + `parentStack`) — avoids tuple allocation per node
- Uses a **shallow copy** per node (`{ ...node }` then `delete flatNode[childrenPropName]`) to preserve the original hierarchical dataset intact (shared object references between flat and hierarchical arrays require this)
- ~3x faster than the previous implementation at 250k nodes (~35ms vs ~100ms)

**`findItemInTreeStructure`** (previously recursive)
- Replaced with an **O(n) iterative DFS** using a stack — eliminates call-stack depth risk on deep trees and is faster than the two-pass recursive approach

**`unflattenParentChildArrayToTree`**
- Replaced the plain object lookup with a **`Map`** — avoids prototype-chain overhead and `Object.keys()` allocation
- Replaced `forEach` with indexed `for` loops

**`addTreeLevelByMutation`**
- Replaced `getTreeDataOptionPropName()` calls with direct `options.childrenPropName` access
- Replaced `forEach` with indexed `for` loop; cached `children` to a local variable (single `Array.isArray` check)

**`addTreeLevelAndAggregatorsByMutation`**
- Same optimizations as above; also caches `isParent` to avoid double `Array.isArray` check

#### treeData.service.ts

- Added **5 cached instance fields** (`_childrenPropName`, `_collapsedPropName`, `_hasChildrenPropName`, `_lazyLoadingPropName`, `_levelPropName`) initialized once in `init()` — eliminates ~20 repeated `getTreeDataOptionPropName()` call sites across the service, each of which previously chained two getter calls plus a switch statement
- Removed all per-call `getTreeDataOptionPropName()` lookups from: `handleOnCellClick`, `handleOnKeyDown`, `toggleTreeDataCollapse`, `updateToggledItem`, `applyToggledItemStateChanges`, `initHierarchicalTree`, `recalculateTreeTotals`, `getItemCount`
- **`toggleAllFlatDataItems`**: inlined the `updateToggledItem(item, collapsing, false)` body directly into the loop — eliminates one function call + two `getTreeDataOptionPropName` lookups per item during full collapse/expand of large datasets

#### utilities.spec.ts
- Updated 4 `unflattenParentChildArrayToTree` test expectations to match `Map` insertion order (previously relied on plain-object auto-sorting of integer keys)
- Updated 3 `flattenToParentChildArray` tests to reflect that `children` property is no longer present on flat items (was previously leaking through)
- Added 7 new unit tests: 3 for `addTreeLevelByMutation`, 2 for `addTreeLevelAndAggregatorsByMutation`, 2 for `findItemInTreeStructure`


ghiscoding:
@zewa666 another win for the AI usage... by looking at the dependency "un-flatten-tree", I found this benchmark created by a user a few years ago: https://github.com/philipstanislaus/array-to-tree-benchmarks

then after asking the AI if I could switch to the more performant lib, Claude AI actually identified that we could use native code (aka drop dependency) and improve the perf by about ~3x (or 300%) with O(n) instead of O(n2)... wow this is just awesome... go ahead Claude 🚀 That's also great because that was 1 of the last package that wasn't fully ESM, which is why it was on my list of TODOs eventually.

my company increased our plan to ~Premium for a month~ because of all the token outcry, many employees got 75% token usage on day 1, so after a few days they bumped our account to Premium but it's just for a month, so better take advantage of it while it last 😆 .... EDIT: ah wait, they changed their mind and reverted to regular access and now I'm out of credit, it was fun while it lasted lol

---

also just let me know when you're done reviewing the other 2 PRs

PR #2602, opened 2026-06-08
feat(react,vue): add RowDetail portal/teleport host context propagation

ghiscoding:
_vibe coded with copilot using Claude Sonnet 4.6_

## Summary

Adds opt-in `RowDetailPortalHost` (React) and `RowDetailTeleportHost` (Vue) companion components that keep row detail panels inside the framework's component tree, giving them full access to providers that were previously unreachable.

## Problem

Both plugins previously rendered each row detail in an **isolated root** outside the main app tree:

- **React** — `createRoot()` was called per row, breaking `useContext`, Redux, Zustand, etc., and producing a dev-mode console warning on every render (documented in the source code).
- **Vue** — `createApp()` was called per row, isolating it from `provide/inject`, Pinia, and Vue Router. The DOM was also moved manually via a temp `div`, which was a workaround for the same problem.

### React — `RowDetailPortalHost`
A companion React component that uses `ReactDOM.createPortal` to render each open row detail into its slickgrid container `div`, while keeping the component inside the app's React tree.

```tsx
import { ReactRowDetailView, RowDetailPortalHost } from '@slickgrid-universal/react-row-detail-plugin';

const rowDetailPlugin = new ReactRowDetailView(eventPubSubService);

return (
  <>
    <SlickgridReact ... externalResources={[rowDetailPlugin]} />
    <RowDetailPortalHost plugin={rowDetailPlugin} />
  </>
);
```

### Vue — `RowDetailTeleportHost`
A companion Vue component that uses the built-in `<Teleport>` directive to render each open row detail into its slickgrid container element, while keeping it inside the Vue component tree.

```html
<SlickgridVue ... :external-resources="[rowDetailPlugin]" />
<RowDetailTeleportHost :plugin="rowDetailPlugin" />
```

## How it works

The plugin exposes a `registerPortalHost()` / `registerTeleportHost()` method. The host component calls it on mount with a state setter, and the plugin pushes entry updates (add/update/remove) to that setter. When the host is not present, the plugin falls back to the previous isolated `createRoot`/`createApp` behavior — **fully backward compatible**.

## Changes

| File | Change |
|------|--------|
| `react-row-detail-plugin/src/reactRowDetailView.ts` | Added `PortalEntry` interface, `registerPortalHost()`, `_updatePortalEntry()`, `_removePortalEntry()`; portal-mode branches in `renderViewModel`, `renderPreloadView`, `disposeViewComponent`, `disposeAllViewComponents` |
| `react-row-detail-plugin/src/RowDetailPortalHost.tsx` | New component |
| `react-row-detail-plugin/tsconfig.json` | Added `"jsx": "react-jsx"` |
| `react-row-detail-plugin/src/index.ts` | Export `RowDetailPortalHost` |
| `vue-row-detail-plugin/src/vueRowDetailView.ts` | Added `TeleportEntry` interface, `registerTeleportHost()`, `_updateTeleportEntry()`, `_removeTeleportEntry()`; teleport-mode branches in render/dispose methods |
| `vue-row-detail-plugin/src/RowDetailTeleportHost.ts` | New component |
| `vue-row-detail-plugin/src/index.ts` | Export `RowDetailTeleportHost` |

## Breaking Changes

None. Both plugins are fully backward compatible — the host component is opt-in.

### DIFF
--- a/demos/react/src/examples/slickgrid/Example19.tsx
+++ b/demos/react/src/examples/slickgrid/Example19.tsx
@@ -1,5 +1,5 @@
 import { type EventPubSubService } from '@slickgrid-universal/event-pub-sub';
-import { ReactRowDetailView } from '@slickgrid-universal/react-row-detail-plugin';
+import { ReactRowDetailView, RowDetailPortalHost } from '@slickgrid-universal/react-row-detail-plugin';
 import React, { useEffect, useRef, useState } from 'react';
 import { Editors, Filters, Formatters, SlickgridReact, type Column, type GridOption, type SlickgridReactInstance } from 'slickgrid-react';
 import { ExampleDetailPreload } from './Example-detail-preload.js';
@@ -21,6 +21,7 @@ const Example19: React.FC = () => {
 
   const serverWaitDelayRef = useRef(serverWaitDelay);
   const reactGridRef = useRef<SlickgridReactInstance | null>(null);
+  const [rowDetailPlugin, setRowDetailPlugin] = useState<ReactRowDetailView | null>(null);
 
   useEffect(() => {
     defineGrid();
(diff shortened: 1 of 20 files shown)


ghiscoding:
there some failing tests because of different behavior in Vue vs React and I don't really want to change the Cypress tests in only 1 implementation compared to others. Also, at this point, it wasn't requested by others and Row Details probably aren't used much in React/Vue wrappers, so let's close for now but maybe revisit later (keep branch for now)

PR #2610, opened 2026-06-13
fix: onBeforeSort should be cancellable from Header Menu

### LINKED ISSUE #2608: Header Menu Sort Ascending/Descending not triggering onBeforeSort
### Describe the bug

Hi there,

it seems that we missed firing and waiting for the reply of onBeforeSort in the ColumnHeaderMenu

Image

I tried looking into the code a bit but unsure what to exactly pursue since the SortingService doesn't handle onBeforeSort but instead onBeforeSortChanged

We're using that feature for the Backend driven grid to avoid losing changes and instead confirm sorting from the user if changes are made.

### Reproduction

open the menu header of a column and try to sort

Angular

### Environment Info

```shell
```

### Validations

ghiscoding:
fixes #2608 

User is able to cancel Sort via the `onBeforeSort` (preventDefault & return false) when clicking on any column header (which was already working)... but it should also be cancellable when called from Header Menu (which wasn't working) and this PR fixes that (see video below)

[Screencast_20260612_204405.webm](https://github.com/user-attachments/assets/a23c71f8-50c7-4adf-9311-be40ba10ac2e)


ghiscoding:
@zewa666 that fixes your issue (see video above), if you intend to review the PR then make sure to ignore white spaces for easier comparisons

zewa666:
thats perfect. thx a lot

PR #2617, opened 2026-06-17
fix(React): assign CustomDataView to SharedService, fixes #2614

### LINKED ISSUE #2614: SlickgridReact customDataView is not assigned to this.dataView during initialization
### Describe the bug

When `SlickgridReact` is initialized with a `customDataView` prop, the component passes that custom DataView to `new SlickGrid(...)`, but it does not assign it to the component's internal `this.dataView`.

In `frameworks/slickgrid-react/src/components/slickgrid-react.tsx`, `this.dataView` is only assigned inside:

```ts
if (!this.props.customDataView) {
  this.dataView = new SlickDataView(...)
}
```

but later in the same `initialization()` method, the code still uses `this.dataView`:

```ts
if (typeof (this.dataView as SlickDataView).setGrid === 'function') {
  this.dataView.setGrid(this.grid);
}
this.sharedService.dataView = this.dataView;
this.bindDifferentHooks(this.grid, this._options, this.dataView);
```

This means `this.dataView` can be undefined when `customDataView` is supplied, causing initialization to fail or leaving `sharedService.dataView` / `instances.dataView` unset.

### Reproduction

Use `SlickgridReact` with a supplied `customDataView`:

```tsx
<SlickgridReact
  gridId="grid1"
  columns={columns}
  dataset={[]}
  options={gridOptions}
  customDataView={customDataView}
/>
```

During `initialization()`, `customDataView` is passed to `new SlickGrid(...)`, but `this.dataView` is never assigned because the assignment only happens in the `if (!this.props.customDataView)` branch.

Current source reference:

- `frameworks/slickgrid-react/src/components/slickgrid-react.tsx`
- `initialization()`
- `customDataView` branch around lines 484-529 in commit `ba7a129f1`

React

### Expected behavior

When `customDataView` is provided, `SlickgridReact` should use it consistently as the internal DataView reference, e.g. assign:

```ts
this.dataView = this.props.customDataView;
```

before creating the SlickGrid or before any internal `this.dataView` access.

### Actual behavior

`customDataView` is passed to `new SlickGrid(...)`, but `this.dataView` remains unset because the internal assignment only happens in the non-custom DataView branch.

### Suggested fix

Assign the custom DataView to the internal field when provided:

```ts
if (this.props.customDataView) {
  this.dataView = this.props.customDataView;
} else {
  const dataviewInlineFilters = (this._options.dataView && this._options.dataView.inlineFilters) || false;
  let dataViewOptions: Partial<DataViewOption> = { ...this._options.dataView, inlineFilters: dataviewInlineFilters };
  ...
  this.dataView = new SlickDataView<TData>(dataViewOptions as Partial<DataViewOption>, this._eventPubSubService);
  this._eventPubSubService.publish('onDataviewCreated', this.dataView);
}
```

This keeps later internal references (`setGrid`, `sharedService.dataView`, `bindDifferentHooks`, `SlickgridReactInstance.dataView`) consistent for both internal and custom DataView cases.

### Environment Info

```shell
| Framework used      | React / Slickgrid-React |
| Slickgrid-Universal | master, commit ba7a129f1 |
| TypeScript          | N/A |
| Browser(s)          | N/A |
| System OS           | N/A |
```

### Validations

- [x] The provided reproduction is a minimal source-level reproduction of the bug.

ghiscoding:
fixes #2614 which is related to an MSSQL issue using a Custom DataView in 1 of their grids.

### DIFF
--- a/frameworks/aurelia-slickgrid/src/custom-elements/aurelia-slickgrid.ts
+++ b/frameworks/aurelia-slickgrid/src/custom-elements/aurelia-slickgrid.ts
@@ -5,6 +5,7 @@ import type {
   BackendServiceOption,
   BasePaginationComponent,
   Column,
+  CustomDataView,
   DataViewOption,
   EventSubscription,
   ExtensionList,
@@ -146,7 +147,7 @@ export class AureliaSlickgridCustomElement {
   @bindable({ mode: BindingMode.twoWay }) totalItems = 0;
   @bindable({ mode: BindingMode.fromView }) extensions!: ExtensionList<any>;
   @bindable({ mode: BindingMode.fromView }) instances: AureliaGridInstance | null = null;
-  @bindable() customDataView?: SlickDataView;
+  @bindable() customDataView?: CustomDataView;
   @bindable() dataset: any[] = [];
   @bindable() datasetHierarchical?: any[] | null;
   @bindable() gridId = '';
--- a/frameworks/slickgrid-react/src/components/slickgrid-react.tsx
+++ b/frameworks/slickgrid-react/src/components/slickgrid-react.tsx
@@ -29,6 +29,7 @@ import {
   type BackendServiceOption,
   type BasePaginationComponent,
   type Column,
+  type CustomDataView,
   type DataViewOption,
   type EventSubscription,
   type ExtensionList,
@@ -159,7 +160,7 @@ export class SlickgridReact<TData = any> extends React.Component<SlickgridReactP
   sortService: SortService;
   treeDataService: TreeDataService;
 
-  dataView!: SlickDataView<TData>;
+  dataView!: SlickDataView<TData> | CustomDataView<TData>;
   grid!: SlickGrid;
   totalItems = 0;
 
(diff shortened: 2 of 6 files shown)


aasimkhan30:
The changes look good overall.  Thanks for the quick turnaround!

One small NIT: 

The new `CustomDataView` interface makes `getItems` and `setItems` **required**, but our windowed view has no `setItems` (it's driven by `setRows`/`setLength`), and the library never calls `setItems` for a custom data view anyway since that call site is gated behind `!props.customDataView`. Could those two stay optional (`?`)? That matches the original interface shape and avoids forcing a no-op `setItems` on consumers.

ghiscoding:
@aasimkhan30  hmm ok I changed it, it required more optional chaining checks in Slickgrid-React for this though... I just thought these 2 functions were the minimum but I guess I was wrong since in the end I've never actually used any Custom DataView myself and I always relied on the SlickDataView. 

So can you give a try to pkg.pr.new nonetheless, it should help confirming if it really works before releasing a new patch version

aasimkhan30:
@ghiscoding , gave it a quick try and everything works. Thanks for making this fix!

ghiscoding:
@aasimkhan30 as you can see above, I've pushed a patch version including this. However please note that I've fixed some Fluent Styling issues with row selection checkboxes in commit bdc2636 (included in v10.8.0) and that is because I saw your colleague patch something similar in MSSQL PR https://github.com/microsoft/vscode-mssql/pull/22198 ... so just make sure to double check the row selection checkboxes in both light/dark theme in your Fluent edit data grid.

PR #2630, opened 2026-06-20
fix(utils): prevent prototype pollution in deepMerge

### LINKED ISSUE #2629: Prototype pollution in public deepMerge / Utilities.deepAssign utility
### Describe the bug

The public `@slickgrid-universal/common` package exports `deepMerge` from `@slickgrid-universal/utils` and also exposes it as `Utilities.deepAssign`.

When either public API is used to deep merge an object containing an own `__proto__` property, attacker-controlled properties can be written to `Object.prototype`. After that, newly created plain objects in the same JavaScript process inherit the polluted property.

This appears to affect the latest published version I tested:

- `@slickgrid-universal/common@10.8.1`
- `@slickgrid-universal/utils@10.8.1`

Relevant source paths:

- `packages/common/src/index.ts`
  - re-exports `@slickgrid-universal/utils`
  - exposes `deepAssign: Utils.deepMerge` inside `Utilities`
- `packages/utils/src/utils.ts`
  - `deepMerge()` iterates `Object.keys(source)`
  - it does not reject dangerous keys such as `__proto__`, `constructor`, or `prototype`
  - when `prop === "__proto__"`, `prop in target` is true for a normal object and `target[prop]` resolves to `Object.prototype`
  - the recursive merge then writes attacker-controlled properties onto `Object.prototype`

Expected behavior:

`deepMerge()` / `Utilities.deepAssign()` should not mutate `Object.prototype`, and dangerous prototype-related keys should be ignored or handled as safe data properties.

### Reproduction

```bash
rm -rf /tmp/slickgrid-common-poc
mkdir /tmp/slickgrid-common-poc
cd /tmp/slickgrid-common-poc

npm init -y
npm install --ignore-scripts --no-audit --no-fund @slickgrid-universal/common@10.8.1

cat > poc.mjs <<'JS'
import { deepMerge, Utilities } from '@slickgrid-universal/common';

delete Object.prototype.slickgridUniversalPolluted;

const payload = JSON.parse('{"__proto__":{"slickgridUniversalPolluted":"yes"}}');

Utilities.deepAssign({}, payload);
console.log('after Utilities.deepAssign:', ({}).slickgridUniversalPolluted);

delete Object.prototype.slickgridUniversalPolluted;

deepMerge({}, payload);
console.log('after deepMerge:', ({}).slickgridUniversalPolluted);

delete Object.prototype.slickgridUniversalPolluted;
JS

node poc.mjs
```

Observed output:

```text
after Utilities.deepAssign: yes
after deepMerge: yes
```

Expected output:

```text
after Utilities.deepAssign: undefined
after deepMerge: undefined
```

Why this matters:

If an application deep merges user-controlled or partially user-controlled objects, prototype pollution can affect later application logic that relies on plain objects, default option objects, inherited properties, or property-existence checks. Depending on how polluted properties are consumed by the application, this can lead to unexpected behavior, logic bypass, denial of service, or other application-specific impact.

Suggested fix:

Before assigning or recursively merging a key, reject prototype-pollution primitives:

```ts
if (prop === '__proto__' || prop === 'constructor' || prop === 'prototype') {
  return;
}
```

It would also be safer to avoid `prop in target` for merge decisions on untrusted keys because it walks the prototype chain. Prefer checking own properties on the target, and avoid recursing into inherited prototype objects.

Other

### Environment Info

```shell
| Framework used      | Other / framework-independent |
| Slickgrid-Universal | @slickgrid-universal/common 10.8.1, @slickgrid-universal/utils 10.8.1 |
| TypeScript          | N/A |
| Browser(s)          | N/A, reproduced in Node.js |
| System OS           | macOS 15.6 |
```

### Validations

Dremig:
fixes #2629

### DIFF
--- a/packages/utils/src/__tests__/utils.spec.ts
+++ b/packages/utils/src/__tests__/utils.spec.ts
@@ -604,6 +604,31 @@ describe('Service/Utilies', () => {
         e: 2,
       });
     });
+
+    it('should ignore prototype pollution keys while merging objects', () => {
+      delete (Object.prototype as any).polluted;
+
+      const input1 = {};
+      const input2 = JSON.parse('{"__proto__":{"polluted":"yes"},"constructor":{"prototype":{"polluted":"yes"}},"prototype":{"polluted":"yes"}}');
+
+      const output = deepMerge(input1, input2);
+
+      expect(output).toEqual({});
+      expect((Object.prototype as any).polluted).toBeUndefined();
+      expect(({} as any).polluted).toBeUndefined();
+    });
+
+    it('should not recursively merge into inherited target properties', () => {
+      const inheritedOptions = { inherited: true };
+      const target = Object.create({ options: inheritedOptions });
+      const source = { options: { own: true } };
+
+      const output = deepMerge(target, source);
+
+      expect(output.options).toEqual({ own: true });
+      expect(Object.prototype.hasOwnProperty.call(output, 'options')).toBe(true);
+      expect(inheritedOptions).toEqual({ inherited: true });
+    });
   });
 
   describe('emptyObject() method', () => {
--- a/packages/utils/src/utils.ts
+++ b/packages/utils/src/utils.ts
@@ -1,5 +1,7 @@
 import type { AnyFunction } from './models/types.js';
 
+const unsafeMergeKeys = new Set(['__proto__', 'constructor', 'prototype']);
+
 /**
  * Add an item to an array only when the item does not exists, when the item is an object we will be using their "id" to compare
  * @param inputArray
@@ -107,8 +109,8 @@ export function deepMerge(target: any, ...sources: any[]): any {
 
   if (isObject(target) && isObject(source)) {
     Object.keys(source).forEach((prop) => {
-      if (source.hasOwnProperty(prop)) {
-        if (prop in target) {
+      if (Object.prototype.hasOwnProperty.call(source, prop) && !unsafeMergeKeys.has(prop)) {
+        if (Object.prototype.hasOwnProperty.call(target, prop)) {
           // handling merging of two properties with equal names
           if (typeof (target as any)[prop] !== 'object') {
             (target as any)[prop] = (source as any)[prop];


ghiscoding:
Thanks but asking the AI, it says the following: 

What still remains an issue
- `prop in target` is still present.
Even though the obvious “dangerous key names” are skipped, using in means the merge behavior depends on inherited properties. That’s not the same as `__proto__` mutation, but it can still cause correctness/security problems:
   - If prop exists on the prototype chain for some objects, you may merge/recursively merge in a way you didn’t intend.
   - It can also create surprising behavior that could become exploitable depending on what other “gadgets” exist in the app.
So: your patch likely fixes the specific `__proto__` pollution PoC, but it is not the fully robust “prototype-pollution-safe” pattern.

What I would change in the proposal (small but important)
If you want the patch to be solid, adjust just these two things:

1. Replace prop in target with an own-property check:
```ts
if (Object.prototype.hasOwnProperty.call(target, prop)) { ... }
```

2. Only deep-merge plain objects (optional but strongly recommended):
- recurse only when both sides are plain objects; otherwise assign/concat.
If you paste the current `isObject()` and the datagrid integration points (what data is fed into `deepMerge`), I can tell you whether the “plain object restriction” is necessary for your case and what the most likely gadget paths are.

Dremig:
@ghiscoding Good point about `prop in target`.

I updated the patch to use `Object.prototype.hasOwnProperty.call(target, prop)` instead, so inherited target properties are no longer considered merge targets. I also added a regression test to make sure `deepMerge()` does not recursively merge into inherited target properties.

I kept the plain-object restriction out of this PR for now because it may be a broader behavioral change for existing users of `deepMerge()`. This patch focuses on the prototype-pollution fix and the inherited-property issue while keeping compatibility risk low.

ghiscoding:
ok let's go with that, did you intend to do more PRs or should I push a patch release with this?

Dremig:
Thanks! I only found this specific issue during my scan, so I do not plan to open more PRs for now.

A patch release with this fix would be great.

Also, as a small process suggestion, would you consider adding a SECURITY.md file and enabling GitHub Private Vulnerability Reporting for this repository? That would give researchers a clear private channel for future security reports.

If you consider this issue security-relevant, I would also appreciate it if you could publish a GitHub Security Advisory for it, so affected users can be properly notified and the fix can be tracked through the usual advisory workflow.

Thanks again for the quick response and review!

PR #2634, opened 2026-06-30
refactor(core)!: drop SortableJS dependency

ghiscoding:
_vibe coded with mostly copilot Claude Sonnet 4.6_

I first tried to migrate from SortableJS to `@formkit/drag-and-drop` in another branch [`refactor/formkit-dnd`](https://github.com/ghiscoding/slickgrid-universal/tree/refactor/formkit-dnd) but copilot suggested to actually drop the use of any external dependency and just use native code by adding about ~\~350 loc~ now ~440 loc after all fixes. Note that this wouldn't get merged until the next major version (next year) since that could impact some E2E tests (I had to add a custom Cypress drag command to make my E2E tests pass) and some small UI differences too (e.g. no more drag animation)

Comparing the build size of my Salesforce standalone script zip file, dropping SortableJS drops the file size by about ~10Kb~ ~9Kb (as a comparison, when I tried to migrate to `@formkit/drag-and-drop`, that only dropped by about ~2.5Kb, and there's still some touch issues with that approach/branch, so dropping SortableJS entirely and going native has a much larger impact for the decrease in size).


ghiscoding:
converting to Draft since I mainly want to see if all tests are passing, if confirmed then I'll probably close the branch and just resurrect it in our next major version in 2027-Q1. I mainly wanted to use my leftover copilot tokens before the next month rolls in

cc @zewa666 

EDIT

After trying it on my linux laptop, I had to patch a few things for Firefox Linux, and also told the AI to patch for any Safari issues as well. Now it's still shaving 10Kb from my zip file but it's now increased from 100 loc to 350 loc (as per codecov), probably still worth it but quite different compared to the original AI assumption 🤔 

EDIT 2

hmm I found some more issues with touch not working, and some issues with Firefox Linux, it works fine with touch with SortableJS though.... 

EDIT 3

hmm ok I think I managed to fix all issues, in the end it's decreasing my zip by ~9Kb now, which is probably still worth it, considering that it's a lighter approach (SortableJS was created when ES6/HTML5 just came out), but I would certainly hold on this until our next major release. It could be good to get feedback on the matter. Only thing I couldn't try is Safari/MacOS but I did ask the AI to cover it too and it's also missing animation when dragging but that is barely noticeable

Cheers

PR #2646, opened 2026-07-13
feat: onBeforeCellEdit hook for RowBasedEditing plugin

zewa666:
provides the opportunity to define a callable hook to decide editor read/write on a per cell basis


zewa666:
the reason for this one is that the user might be using his own onBeforeEditCell event, which the plugin also defines. So since the broadcast isnt defining how the order of multiple modal results are handled, this way the user can define the event as a hook, being called by the plugin itself.

the use case is to allow individual users only edit specific cells, based on whatever criteria.

ghiscoding:
fine by me, can you run Prettier write to fix the formatting issue though... nevermind, it looks like it's fixed now. Thanks

ghiscoding:
hey since you're online but on a completely different topic... 6pac just got introduced to AI and he's going wild with his new Claude Pro or Premium plan (170$/month), he's running Claude Fable to rewrite Freezing/Pinning with a different approach not using div I think (I'm not sure if it's worth it yet, but it would also add right pinning which we don't currently have and it wouldn't create all these viewports when not using freezing (which it currently does)). He also has another PR to add variable row height, but I told him to rethink this one since we can do better.... anyway, if you have some time and to review/comment on them, feel free to join the party. I told him that especially his freezing rewrite should be a major version on his repo and if I decide to go ahead with that too in here, that wouldn't be until my next major around January

check the last currently open PRs he created
https://github.com/6pac/SlickGrid/pulls

zewa666:
oha, new-found power with more Tokens ... 🤣 

thanks for letting me know. Variable row height would actually be one of the last limitations I know of and cant nicely work around. I'll check them out and add my comments if helpful

ghiscoding:
great thanks, let's merge your PR and release around the end of this week. 
Cheers and welcome back 😉

PR #2648, opened 2026-07-17
fix(regression): prevent row selection from being dropped when last column is hidden

Heimo001:
### Problem

When any column with `hidden: true` is the **last column** in the column definition array, calling `setSelectedRows()` (or right-clicking to open a context menu that uses `setSelectedRows` internally) silently fails — the row is never actually selected, and `getSelectedRows()` returns an empty array.

### Root Cause

`SlickGrid.rowsToRanges()` builds a `SlickRange` spanning from cell `0` to the last column index, computed as `this.columns.length - 1`. This index includes **all** columns, including hidden ones.

When the resulting range is passed to `setSelectedRanges()` while `_activeSelectionIsRow` is `false` (which happens when the active cell is on a regular non-row-select column, e.g. triggered by a right-click), `removeInvalidRanges()` is called. It validates both endpoints of the range via `canCellBeSelected()`, which returns `false` for any hidden column — **dropping the entire range**.

```
grid.setSelectedRows([row])
  → SlickGrid.rowsToRanges([row])
      → lastCell = this.columns.length - 1   ← may point to a hidden column
      → new SlickRange(row, 0, row, hiddenColIndex)

  → selectionModel.setSelectedRanges(ranges)  where _activeSelectionIsRow = false
      → removeInvalidRanges(ranges)
          → canCellBeSelected(row, hiddenColIndex)
              → columns[hiddenColIndex].hidden === true  → returns false
          → range is dropped ✗

  → getSelectedRows() returns []
```

### Fix

`rowsToRanges()` now accepts an optional `useVisibleColumnsOnly` flag (defaulting to `false` to preserve backward compatibility). When `true`, `lastCell` is computed from the last **visible** column's real index via `getColumnIndex()`. `SlickGrid.setSelectedRows()` exposes the same flag as an optional third parameter.

### Reproduce

A grid where the last column in the columns array has `hidden: true` (e.g. `effortDriven` below), combined with a Context Menu that calls `setSelectedRows()` inside `itemVisibilityOverride` to determine whether to show a menu item:

```ts
columns = [
  { id: 'title', ... },
  { id: 'finish', ... },
  {
    id: 'effortDriven',
    hidden: true,   // ← last column, hidden
    ...
  },
];

gridOptions = {
  ...
  enableSelection: false,
  selectionOptions: {
    selectionType: 'mixed',
  },
  contextMenu: {
    commandItems: [{
      command: 'copySelectedRows',
      itemVisibilityOverride: (item) => {
        // before the fix: last column is hidden → range is silently dropped
        grid.setSelectedRows([item.row]);
        return grid.getSelectedRows().includes(item.row); // always false ← bug
      }
    }]
  }
};
```

The exact scenario was tested and validated in `example48.component.ts` of the Angular Slickgrid demos using the modifications above.

### DIFF
--- a/demos/vanilla/src/examples/example37.ts
+++ b/demos/vanilla/src/examples/example37.ts
@@ -84,7 +84,7 @@ export default class Example37 {
 
   /* Define grid Options and Columns */
   defineGrids() {
-    this.columns1 = [
+    const baseColumns = [
       { id: 'id', name: '#', field: 'id', width: 32, maxWidth: 40, excludeFromHeaderMenu: true },
       { id: 'title', name: 'Title', field: 'title', width: 90, cssClass: 'cell-title' },
       { id: 'complete', name: '% Complete', field: 'percentComplete', sortable: true, width: 90 },
@@ -121,8 +121,18 @@ export default class Example37 {
         exportCustomFormatter: (_row, _cell, value) => (value ? 'Yes' : 'No'),
         formatter: Formatters.checkmarkMaterial,
       },
+    ] satisfies Column<any>[];
+
+    this.columns1 = [
+      ...baseColumns,
+      {
+        id: 'lastColumnHiddenColumn',
+        name: 'Last Column / Hidden Column',
+        field: 'lastColumnHiddenColumn',
+        hidden: true,
+      },
     ];
-    this.columns2 = [...this.columns1];
+    this.columns2 = [...baseColumns];
 
     this.gridOptions1 = {
       autoResize: {
@@ -154,6 +164,21 @@ export default class Example37 {
         copyActiveEditorCell: true,
         removeDoubleQuotesOnPaste: true,
         replaceNewlinesWith: ' ',
+      },
+
+      contextMenu: {
+        commandItems: [
+          {
+            command: 'commandWhichShouldBeShown',
+            title: 'Command which should be shown',
+            iconCssClass: 'mdi mdi-check',
+            itemVisibilityOverride: (item) => {
+              // before the fix: last column is hidden → range is silently dropped
+              this.sgb1.slickGrid?.setSelectedRows([item.row ?? 0]);
+              return this.sgb1.slickGrid?.getSelectedRows().includes(item.row ?? 0) ?? false; // always false ← bug
+            },
+          },
+        ],
       },
     };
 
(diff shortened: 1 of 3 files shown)


ghiscoding:
oops another regression after adding hidden columns in v10, thanks for the fix, however...

> `rowsToRanges()` now accepts an optional `useVisibleColumnsOnly` flag (defaulting to `false` to preserve backward compatibility). When `true`, `lastCell` is computed from the last **visible** column's real index via `getColumnIndex()`. `SlickGrid.setSelectedRows()` exposes the same flag as an optional third parameter.

what "**preserve backward compatibility**" do we actually expect here? I mean this seems to now require the user to pass an extra `true` boolean for it to work and behave like it was in previous v9.x, is that the case? If so, then I think we should just drop that extra argument and just execute it directly. 

in summary, the only backward compatibility that I expect is to behave the same as it was in v9 (and for history purposes, we didn't have `hidden` column property in v9 and if we wanted to hide a column, it was simply being removed from the SlickGrid columns array and so `getColumns()` was previously returning non-hidden columns but I had to keep a global reference to the original columns to compare against, but now that we have `hidden` column property then I no longer need to keep a global ref of the original columns)

zewa666:
@ghiscoding the decision to go with the boolean was made since the both of us (@Heimo001 is my coworker) really didnt have a clue whether the new behavior was intended, as the mentioned functions are called from a couple of different places. If you can't think of any I'll check up with him to see if we can add at least one E2E test to lock this behavior for the future

ghiscoding:
cool, the more, the merier... 😆 

I would say that it probably needs more investigation but if we remove that extra argument and all Cypress tests pass then I don't think it's needed. So let's try to remove the argument and see if all tests pass, if so then it's probably ok without

Heimo001:
The useVisibleColumnsOnly optional flag previously added to rowsToRanges() is removed — rowsToRanges() now unconditionally uses getVisibleColumns().

I went with example37 Grid 1 for the regression test since it already has the hybrid selection model (selectionType: 'mixed') set up. Adding a context menu with the itemVisibilityOverride logic there didn't require touching much of the existing grid config.

I also extracted the shared columns into a baseColumns variable and only append the hidden last column to Grid 1's array, that way Grid 2 on the same page isn't accidentally affected.

ghiscoding:
awesome thanks a lot, I'm pretty sure we didn't need the extra argument and we'll probably know soon enough if anyone complains of the new behavior (which I doubt). So let's go ahead and merge, thanks for the contribution 😄 

oh and if you haven't upvoted (⭐) the project already, I wouldn't mind more stars hehe 😉 

I got other PRs in the queue, so it might fall to be an end of the week release
Cheers.

PR #2650, opened 2026-07-17
feat(core): add variable row height support

ghiscoding:
_vibe coded mostly with Claude Sonnet 4.6 for initial draft and copilot auto mode for the rest_

## What
Port/align behavior from 6pac/SlickGrid PR #1235 into slickgrid-universal, with one intentional API clarification.

Variable row height is now explicit opt-in via enableVariableRowHeight.

~~~js
const grid = new Slick.Grid('#myGrid', data, columns, {
  rowHeight: 25, // default / fallback
  enableVariableRowHeight: true,
  // optional: custom provider
  rowHeightProvider: (grid, row, item) => (item?.tall ? 50 : undefined),
});
~~~

If enableVariableRowHeight is false, fixed-height path stays active (no variable-height indexing path).

Height resolution in variable mode:
1. rowHeightProvider result, if provided and returns a number
2. ItemMetadata.height, via default provider behavior
3. rowHeight fallback

Important: custom rowHeightProvider is authoritative. if it returns undefined, fallback is rowHeight (not metadata).

## Design
No duplicated render path. geometry still goes through shared row top/bottom/position helpers, with branching inside geometry primitives only.

Fixed mode keeps same arithmetic/perf profile.
Variable mode uses row-position index rebuilds only when needed (dirty flag / count changes / invalidation).

## Changes
1. Core variable-height flow in SlickGrid:
- explicit enableVariableRowHeight gate
- default metadata-aware provider behavior
- provider-only index rebuild source
- invalidateRowHeights support

2. Option + docs updates:
- grid option docs now describe explicit flag
- rowHeightProvider semantics clarified (override behavior + fallback behavior)

3. Rendering/CSS behavior:
- default row height still provided by stylesheet
- per-row inline height only when row differs from default
- kept local slickgrid-universal styling strategy

4. Cell height parity:
- getCellHeight aligned with 6pac logic (rowspan + cellHeightDiff flow)

## Verification
1. Targeted core spec passed after changes:
- slickGrid.spec.ts
- 411 passing, 0 failing

2. Confirmed behavior:
- metadata-only works with enableVariableRowHeight true and no custom provider
- fixed-height mode unchanged when flag is false
- custom provider returning undefined falls back to rowHeight

### Demo

image

### DIFF
--- a/demos/aurelia/src/examples/slickgrid/example55.html
+++ b/demos/aurelia/src/examples/slickgrid/example55.html
@@ -1,0 +1,34 @@
+<div class="container-fluid">
+  <h2>
+    Example 55: Variable Row Height (Provider)
+    <span class="float-end">
+      <a
+        style="font-size: 18px"
+        target="_blank"
+        href="https://github.com/ghiscoding/slickgrid-universal/blob/master/demos/aurelia/src/examples/slickgrid/example55.ts"
+      >
+        <span class="mdi mdi-link-variant"></span> code
+      </a>
+    </span>
+  </h2>
+
+  <div class="subtitle">Variable row heights driven by <code>rowHeightProvider</code>.</div>
+
+  <div class="row" style="margin-bottom: 6px">
+    <div class="col-md-12">
+      <button class="btn btn-outline-secondary btn-sm btn-icon" click.trigger="scrollToRow90()" data-test="scroll-row-90-example55">
+        <span class="mdi mdi-arrow-down"></span>
+        <span> Scroll To row 90</span>
+      </button>
+    </div>
+  </div>
+
+  <aurelia-slickgrid
+    grid-id="grid55"
+    columns.bind="columns"
+    options.bind="gridOptions"
+    dataset.bind="dataset"
+    on-aurelia-grid-created.trigger="aureliaGridReady($event.detail)"
+  >
+  </aurelia-slickgrid>
+</div>
--- a/demos/aurelia/src/examples/slickgrid/example55.scss
+++ b/demos/aurelia/src/examples/slickgrid/example55.scss
@@ -1,0 +1,12 @@
+#slickGridContainer-grid55 {
+  --slick-cell-border-left: 1px solid #dedede;
+}
+
+#slickGridContainer-grid55 .slickgrid-container .slick-cell.cell-wrap {
+  line-height: 1.25;
+  padding: 4px 6px;
+  white-space: normal;
+  text-overflow: clip;
+  overflow: hidden;
+  overflow-wrap: anywhere;
+}
--- a/demos/aurelia/src/examples/slickgrid/example55.scss
+++ b/demos/aurelia/src/examples/slickgrid/example55.scss
@@ -3,8 +3,6 @@
 }
 
 #slickGridContainer-grid55 .slickgrid-container .slick-cell.cell-wrap {
-  line-height: 1.25;
-  padding: 4px 6px;
   white-space: normal;
   text-overflow: clip;
   overflow: hidden;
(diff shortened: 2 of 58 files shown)


ghiscoding:
cc @zewa666 variable row height in Slickgrid-Universal 😄 

Ben added a row height provider that can be added via grid option, but we also have the fallback to item `metadata.height` available as well. So we have 2 ways to do it... also codecov reports ~106~ 127 loc, which isn't too bad I guess to get this great new feature 🚀

zewa666:
hell yeah, looking forward to this release

ghiscoding:
probably next week, I'll wait to see if Ben (6pac) update his PR to include my asks and fixes (I guess he will pretty soon since his Claude Fable 90% promo is about to end) 😆 

so your own PR fix will have to wait few more days, but I guess it's worth the wait 😉

ghiscoding:
@zewa666 hi there, so I tagged you in 6pac repo because I'd like to get your opinion as a user since Ben and I have a bit of a disagreement on this feature on how to have both the `rowHeightProvider` and the item metadata `height`. I made mine as fallback but it seems to cause confusion and disagreement, so opinions from other users like you could help. So if you have time to reply, that would be appreciated to get this to the finish line. Thanks

zewa666:
hey there, currently on vacation so I saw this a little bit late. I'm following the original topic as well and it seems the end result feels like any other slickgrid-universal style config with fallbacks. I'm totally fine with this

ghiscoding:
The feature is finally ready to be merged, coming soon to a blockbuster near you 😆 🚀

zewa666:
thats great, prepping our apps updates currently and looking forward to the upcoming releases

ghiscoding:
@zewa666 side note, it might take couple more days because Ben just opened another AI generated, of multiple possible fixes, which seemed quite legit. So I'll wait couple more days to see what comes out of it, but in any case I will most probably publish by end of this week. Ben does seem to like working with Claude AI recently lol.

see https://github.com/6pac/SlickGrid/discussions/1247

Also a side note, after a few conversions back and forth with Ben, I decided it was better to add a flag for thies variable height feature, the main reason is that we now have this default global grid option `rowHeightProvider: (grid, row) => grid.getItemMetadaWhenExists(row)?.height,` which makes it work with item metadata... but this code is being executed on every single dataset row (unless the user overrides the provider), then this mean that it could have an impact on extremely large dataset, that is especially if you're not using this new feature. So I decided it was better to add a `enableVariableRowHeight` flag to be explicit when the user wants to use it, this way the user knows there could be a perf impact and it also doesn't impact previously existing grids (in other words, existing grids should never be impacted by our decisions to add new grid features like this, hence the new flag). The flag is also much better for functions like `scrollTo()`, since we know right away which code path to execute to calculate position and all...

ohhh I just realized that I haven't tested the Excel export, I'm not sure if it would be a good idea to carry over the row height!? It might be too much though, we always used the default Excel row height, so this might be overkill to carry over... dunno

zewa666:
we're not using the Excel Export so I cant say anything about that, but I'd assume it should be doable, same as the pdf export. Having it behind a feature flag is certainly a good idea, and once we tested it we can provide more feedback/fixes

ghiscoding:
oh ok I'm actually surprised, I always assumed you were using Excel export because of all the Excel conversations we had over the years 😆

zewa666:
we're using the copy & paste feature which works to and from excel, but not the actual export feature itself which generates an xslx. Our apps sole purpose is to kill excel files in our company 😅 hence why row height isnt a topic with that approach

ghiscoding:
so you're saying my Excel export is a competitor hahaha

zewa666:
let me put it that way, together with the formula exports its one of the best hidden secrets we have 😅

PR #2653, opened 2026-07-20
feat: add `keepComponentAlive` to all 4 framework Row Detail plugins

ghiscoding:
**Adds `keepComponentAlive` option to preserve row detail component state when scrolling**

fixes #2651 

### Why Opt-In?
By default, row detail components are destroyed and recreated when scrolling out of viewport. This is memory-efficient for large datasets. However, this destroys component state (form input, nested grid filters, scroll position, etc.).

The new `keepComponentAlive` flag enables keeping components alive in memory so state is preserved. It's **opt-in by default** (false) to maintain backward compatibility and avoid unexpected memory overhead for existing applications with large datasets.

### Changes
- Added `keepComponentAlive?: boolean` option to `RowDetailViewOption` interface
- Implemented keep-alive feature across all 4 framework row detail plugins:
  - Angular: Uses `ApplicationRef.detachView()` and `attachView()` 
  - Aurelia: Detaches/reattaches slots from DOM
  - React: Persists container element for root re-render
  - Vue: Stores component instance and reattaches to fresh container
- Added UI toggle to Example45 demos in all 4 frameworks
- Updated comprehensive documentation in all 5 row-detail.md files
- Fixed critical bug: `rowDetailViewOptions` getter now reads from `getOptions()` instead of stale `gridOptions.rowDetailView`
- Push a fix to properly detect when Row Detail comes back to viewport visible range

### Use Cases
Enable `keepComponentAlive` when you need to preserve component state:
- Nested grids with active filters/sort
- Forms with user input that shouldn't be lost
- Components with expensive initialization
- Long scroll lists where re-render cost is significant

### Trade-offs
- **Memory**: Components stay in memory; impacts large datasets with many expanded rows
- **Performance**: Avoids re-render cost when scrolling back; negligible overhead if few rows expanded
- **Graceful fallback**: If reattach fails, automatically falls back to fresh re-render

### References
Related to discussion: [Row Detail Plugin - Official approach for rendering Angular components in v10.x (standalone)? · ghiscoding/slickgrid-universal · Discussion #2651](https://github.com/ghiscoding/slickgrid-universal/discussions/2651)


ghiscoding:
@jr01 this might interest you since I know you are a user of Row Detail with Angular-Slickgrid... for reference, I asked copilot to see if the discussion opened in #2651 is fixable and this PR tries to fix those concerns by adding a new flag `keepComponentAlive` (I barely provided guidance to copilot). 

So I wouldn't mind your feedback on this PR if you have any special thoughts. Thanks

jr01:
Thanks @ghiscoding ), this looks really great!! I think the opt-in default is exactly right: it solves the RowDetail state-loss problem while preserving the existing memory-efficient behavior for large grids.

I also like that an explicit collapse still destroys the component, and the viewport re-entry edge-case handling and regression tests. This gives us a clean path to use stateful RowDetail components (nested grids, forms, expensive initialization) without workarounds.

No blockers from my side. Great work, and thanks! 💯 🥇 Looking forward to applying this.

PR #2680, opened 2026-07-30
fix(a11y): add missing ARIA row/rowgroup structure to grid headers

hjaewon:
### Describe your change

The grid container gets `role="grid"` and header cells get `role="columnheader"`, but the three container
levels in between (`.slick-header-container > .slick-header > .slick-header-columns`) carry no role. Since
elements without a role and without focusability are transparent in the accessibility tree, this produces two
axe violations on every grid:

- **`aria-required-parent`** — one violation per column: `role="columnheader"` has no `role="row"` parent
  (`Required ARIA parent role not present: row`).
- **`aria-required-children`** — `role="grid"` is computed to directly own `[role=columnheader]` and
  `div[tabindex]` (the two 0×0 focus sinks), neither of which is an allowed child of `grid`.

Nothing is visually broken — the impact is that screen readers do not get the table structure.

This adds the missing skeleton at element-creation time:

| Element | Role added |
|---|---|
| `.slick-header` (header scroller, L/R) | `rowgroup` |
| `.slick-header-columns` (L/R) | `row` |
| `.slick-headerrow` (filter row scroller, L/R) | `rowgroup` |
| `.slick-headerrow-columns` (L/R) | `row` |
| `.slick-headerrow-column` (filter row cell) | `gridcell` |
| focus sinks (0×0, `tabindex="-1"`) | `aria-hidden="true"` |

Two notes on scope:

- The **filter row is fixed as a unit**. Giving `.slick-headerrow-columns` a `row` role without giving its
  cells a `gridcell` role would introduce a *new* `aria-required-children` violation (a `row` must own cells)
  on grids that show the filter row, so both are in this patch.
- The focus sinks are `tabindex="-1"`, so they are not tabbable and `aria-hidden` does not trigger
  `aria-hidden-focus`.

`aria-colcount` / `aria-rowcount` are already handled correctly (`updateRowCount()`), so they are untouched.

### How to reproduce

Run axe (e.g. a Lighthouse accessibility audit) on any example page containing a grid. Originally reported
downstream by a consumer of a fork: 15 `aria-required-parent` violations (= column count) plus 1
`aria-required-children` violation, on a 15-column grid.

### Verification

- `pnpm exec tsc --build ./tsconfig.packages.json` — no errors in the changed file.
- `packages/common` unit tests: **4297 passed** (181 files), which includes the 2 tests added here.
  No existing assertion changes: the suite asserts `className` on those containers, not the full attribute set.
- `oxlint` clean, `prettier --check` clean (long lines use the `// prettier-ignore` convention already used in
  this file).
- Two unit tests added in `slickGrid.spec.ts`: one asserting the `rowgroup > row > columnheader` and
  `rowgroup > row > gridcell` chains, one asserting the focus sinks are `aria-hidden`.

The equivalent fix has also been running as a runtime patch in a downstream facade, verified against both axe
rules (ownership computed the same way axe does) across three feature combinations — plain grid, inline filter
row, and grouping + footer totals + toolbar — plus a negative control confirming the check actually catches the
violation when the roles are removed.

### DIFF
--- a/packages/common/src/core/__tests__/slickGrid.spec.ts
+++ b/packages/common/src/core/__tests__/slickGrid.spec.ts
@@ -3101,6 +3101,33 @@ describe('SlickGrid core file', () => {
       expect(grid.getContainerNode().getAttribute('aria-rowcount')).toBe('11');
     });
 
+    it('should expect the header structure to have the ARIA roles required between grid and columnheader', () => {
+      grid = new SlickGrid<any, Column>(container, items, columns, { ...defaultOptions, showHeaderRow: true });
+
+      const headerColumnElm = container.querySelector('[role="columnheader"]') as HTMLDivElement;
+      expect(headerColumnElm).toBeTruthy();
+      // a columnheader requires a "row" parent, and that row requires a "rowgroup" (or grid) parent,
+      // otherwise axe reports `aria-required-parent` + `aria-required-children` on every grid
+      expect(headerColumnElm.parentElement!.classList.contains('slick-header-columns')).toBe(true);
+      expect(headerColumnElm.parentElement!.getAttribute('role')).toBe('row');
+      expect(headerColumnElm.parentElement!.parentElement!.getAttribute('role')).toBe('rowgroup');
+
+      // same skeleton for the header row (filter row): rowgroup > row > gridcell
+      const headerRowCellElm = container.querySelector('.slick-headerrow-column') as HTMLDivElement;
+      expect(headerRowCellElm.getAttribute('role')).toBe('gridcell');
+      expect(headerRowCellElm.parentElement!.getAttribute('role')).toBe('row');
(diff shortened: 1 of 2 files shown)


hjaewon:
Closing this — submitted prematurely on my side, not ready for review. Sorry for the noise.

ghiscoding:
you could have left it open but changed to "Draft" then I know it isn't ready yet... a11y fixes are more than welcome, since that is not my expertise at all, so thanks in advance... :)

ghiscoding:
@hjaewon this PR seemed like a good start, are you planning on re-opening the PR? I'll be pushing a new release soon, thanks

ghiscoding:
I've asked copilot (ChatGPT) to open an updated PR superseding your original code, it found that the `aria-hidden` added to the focus sink could be improved and it made a safer approach, the rest of the PR is the same. Thanks

PR #2697, opened 2026-08-05
feat(export): add opt-in includeColumnWidth for Excel and PDF exports

ghiscoding:
_vibe coded with Copilot using GPT-5.3-Codex_

after recently adding variable row height in Excel/PDF exports, I figured that we should also add the column width as well but as opt-in.

## Why
Exports were already handling row height fidelity, but column width fidelity was inconsistent:
- Excel width export mostly relied on export-specific width settings and global fallback.
- PDF had width support in parts of the flow, but not consistent parity across rendering paths.
- Because runtime grid columns typically end up with a width value (defaulted by grid internals), enabling width import by default would silently change many existing exports.

This change introduces an explicit, safe way to export using grid column widths without breaking current behavior.

## What changed (behavior)
A new option is introduced for both Excel and PDF:
- includeColumnWidth: boolean (default false)

When includeColumnWidth is false:
- Existing behavior is preserved.

When includeColumnWidth is true:
- Export uses column definition width as fallback source for each column width.

Width resolution precedence is now explicit:
1. Per-column export width (highest priority)
2. Column definition width when includeColumnWidth is true
3. Existing global/default fallback behavior

In plain terms:
- If you already set export-specific width, nothing changes.
- If you do not set export-specific width and opt in, export follows your grid column width.
- If you do not opt in, export keeps prior defaults.

## Why opt-in
This is intentionally opt-in for backward compatibility:
- Grid columns generally have a width at runtime (explicit or defaulted).
- A default-on approach would alter output for many users unexpectedly.
- Opt-in keeps existing exports stable while enabling fidelity for users who want WYSIWYG width behavior.

## Notes on units
- Excel width uses Excel column-width units (not pixels).
- PDF width uses points (pt), consistent with jsPDF defaults.
- includeColumnWidth uses the column width value as-is within each export pipeline’s existing unit semantics.

## Scope details
- Applied to Excel export width logic.
- Applied to PDF export width logic in both render paths:
1. jsPDF AutoTable path
2. Manual fallback rendering path

## Validation
- Targeted Excel export tests pass.
- Targeted PDF export tests pass.
- New tests cover default behavior, opt-in behavior, and precedence behavior.

## Summary
 We now have 2 new options in Excel/PDF exports for row height and column width:
- `includeVariableRowHeight` => default true **but only** when variable row height is enabled
- `includeColumnWidth` => default false (opt-in)

### Export Demo

image

### DIFF
--- a/demos/aurelia/src/examples/slickgrid/example55.html
+++ b/demos/aurelia/src/examples/slickgrid/example55.html
@@ -16,6 +16,12 @@
 
   <div class="row" style="margin-bottom: 6px">
     <div class="col-md-12">
+      <button class="btn btn-outline-secondary btn-sm btn-icon" data-test="export-excel-btn" click.trigger="exportToExcel()">
+        <i class="mdi mdi-file-excel-outline text-success"></i> Export to Excel
+      </button>
+      <button class="btn btn-outline-secondary btn-sm btn-icon" data-test="export-pdf-btn" click.trigger="exportToPdf()">
+        <i class="mdi mdi-file-pdf-outline text-danger"></i> Export to PDF
+      </button>
       <button class="btn btn-outline-secondary btn-sm btn-icon" click.trigger="scrollToRow90()" data-test="scroll-row-90-example55">
         <span class="mdi mdi-arrow-down"></span>
         <span> Scroll To row 90</span>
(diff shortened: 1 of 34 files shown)


ghiscoding:
cc @jr01 you probably want to be aware of this feature/change since you're using the PDF Export. The summary is that I recently added variable row height and also in Excel/PDF exports, I figured that we should also add the column width as well but as opt-in. We now have 2 new options:
- `includeVariableRowHeight` => default true **but only** when variable row height is enabled
- `includeColumnWidth` => default false (opt-in)

PR #2734, opened 2026-08-17
fix(event-pub-sub): fix listener leak, naming convention logic and subscribeEvent double-transformation

rlamtabbet:
### 1. Listener leak in `subscribe` / `unsubscribe`
`subscribe` registered a wrapped listener on the DOM but stored the original callback in `_subscribedEvents`. `unsubscribe` then called `removeEventListener` with the original callback, which was never the registered listener, so it was never actually removed.

Fix: introduce `_callbackMap` (Map<Function, Function>) to track original callback, wrapped listener, allowing `unsubscribe` to resolve and remove the correct DOM listener. Map is cleared on `dispose`.

### 2. `getEventNameByNamingConvention` missing `else` branches
The `camelCaseWithExtraOnPrefix` and `lowerCaseWithoutOnPrefix` cases were missing `else` branches, causing the second assignment to always execute regardless of which style was active, corrupting the output event name for other styles sharing the same `case` block.

### 3. `subscribeEvent` unsubscribe double-transformation
The unsubscribe closure passed the already-converted event name back into `unsubscribe()`, which called `getEventNameByNamingConvention` on it again, producing a double-transformed name that never matched the stored event, so the listener was never removed.

Fix: bypass `unsubscribe()` in the closure and operate directly on the already-converted name.

## Tests
- Updated existing tests to reflect correct wrapped-listener behaviour
- Added new tests covering the listener leak fix and `dispose` cleanup

### DIFF
--- a/packages/event-pub-sub/src/eventPubSub.service.spec.ts
+++ b/packages/event-pub-sub/src/eventPubSub.service.spec.ts
@@ -182,16 +182,26 @@ describe('EventPubSub Service', () => {
 
       service.eventNamingStyle = 'lowerCaseWithoutOnPrefix';
       const subscription = service.subscribeEvent('onDblClick', mockCallback);
-      divContainer.dispatchEvent(new CustomEvent('DblClick', { composed: true, detail: { name: 'John' } }));
+      divContainer.dispatchEvent(new CustomEvent('dblclick', { composed: true, detail: { name: 'John' } }));
 
       expect(getEventNameSpy).toHaveBeenCalledWith('onDblClick', '');
-      expect(service.subscribedEventNames).toEqual(['DblClick']);
+      expect(service.subscribedEventNames).toEqual(['dblclick']);
       expect(service.subscribedEvents.length).toBe(1);
-      expect(addEventSpy).toHaveBeenCalledWith('DblClick', expect.any(Function));
+      expect(addEventSpy).toHaveBeenCalledWith('dblclick', expect.any(Function));
       expect(mockCallback).toHaveBeenCalledTimes(1);
 
       subscription.unsubscribe();
       expect(service.subscribedEvents.length).toBe(0);
+    });
+
+    it('keeps lowerCaseWithoutOnPrefix event names lowercase', () => {
+      const callback = vi.fn();
+      service.eventNamingStyle = 'lowerCaseWithoutOnPrefix';
+      service.subscribeEvent('onDblClick', callback);
+
+      divContainer.dispatchEvent(new CustomEvent('dblclick'));
+
+      expect(callback).toHaveBeenCalledOnce();
     });
 
     it('should call subscribe method and expect "addEventListener" and "getEventNameByNamingConvention" to be called with kebabCase', () => {
@@ -348,6 +358,50 @@ describe('EventPubSub Service', () => {
       expect(removeEventSpy).toHaveBeenCalledWith('onClick', expect.any(Function));
       expect(removeEventSpy).toHaveBeenCalledWith('onDblClick', expect.any(Function));
       expect(unsubscribeSpy).toHaveBeenCalledTimes(2);
+      expect(service.subscribedEvents.length).toBe(0);
+    });
+
+    it('removes every event when a shared callback subscription is unsubscribed', () => {
+      const mockCallback = vi.fn();
+      const subscription = service.subscribe(['onClick', 'onDblClick'], mockCallback);
+
+      divContainer.dispatchEvent(new CustomEvent('onClick', { detail: { name: 'John' } }));
+      divContainer.dispatchEvent(new CustomEvent('onDblClick', { detail: { name: 'Jane' } }));
+      expect(mockCallback).toHaveBeenCalledTimes(2);
+
+      subscription.unsubscribe();
+
+      divContainer.dispatchEvent(new CustomEvent('onClick'));
+      divContainer.dispatchEvent(new CustomEvent('onDblClick'));
+      expect(mockCallback).toHaveBeenCalledTimes(2);
+      expect(service.subscribedEvents.length).toBe(0);
+    });
+
+    it('removes a subscribe callback with camelCaseWithExtraOnPrefix', () => {
+      const mockCallback = vi.fn();
+      service.eventNamingStyle = 'camelCaseWithExtraOnPrefix';
+      const subscription = service.subscribe('onDblClick', mockCallback);
+
+      subscription.unsubscribe();
+      divContainer.dispatchEvent(new CustomEvent('onOnDblClick'));
+
+      expect(mockCallback).toHaveBeenCalledTimes(0); // never fired before unsubscribe either
+    });
+
+    it('should correctly unsubscribe all listeners when same callback is used for multiple events (array subscribe)', () => {
+      const mockCallback = vi.fn();
+      service.subscribe(['onClick', 'onDblClick'], mockCallback);
+
+      divContainer.dispatchEvent(new CustomEvent('onClick', { detail: { name: 'John' } }));
+      divContainer.dispatchEvent(new CustomEvent('onDblClick', { detail: { name: 'Jane' } }));
+      expect(mockCallback).toHaveBeenCalledTimes(2);
+
+      service.unsubscribeAll();
+
+      // both DOM listeners must be truly removed
+      divContainer.dispatchEvent(new CustomEvent('onClick', { detail: { name: 'John' } }));
+      divContainer.dispatchEvent(new CustomEvent('onDblClick', { detail: { name: 'Jane' } }));
+      expect(mockCallback).toHaveBeenCalledTimes(2);
       expect(service.subscribedEvents.length).toBe(0);
     });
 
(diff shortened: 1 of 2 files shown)


packages/event-pub-sub/src/eventPubSub.service.spec.ts
```diff
@@ -182,12 +182,12 @@ describe('EventPubSub Service', () => {
 
       service.eventNamingStyle = 'lowerCaseWithoutOnPrefix';
       const subscription = service.subscribeEvent('onDblClick', mockCallback);
-      divContainer.dispatchEvent(new CustomEvent('dblclick', { composed: true, detail: { name: 'John' } }));
+      divContainer.dispatchEvent(new CustomEvent('DblClick', { composed: true, detail: { name: 'John' } }));
```
ghiscoding: I'm wondering why you changed the casing here? lowercase is the correct and standard way of defining an event.

packages/event-pub-sub/src/eventPubSub.service.ts
```diff
@@ -145,8 +149,13 @@ export class EventPubSubService implements BasePubSubService {
 
       // the event listener will return the data in the "event.detail", so we need to return its content to the final callback
       // basically we substitute the "data" with "event.detail" so that the user ends up with only the "data" result
-      this._elementSource.addEventListener(eventNameByConvention, (event: CustomEventInit<T>) => callback.call(null, event.detail as T));
-      this._subscribedEvents.push({ name: eventNameByConvention, listener: callback });
+      const wrappedListener = (event: CustomEventInit<T>) => callback.call(null, event.detail as T);
+
+      // store the mapping so unsubscribe can resolve the original callback to the wrapper
+      this._callbackMap.set(callback, wrappedListener);
```
ghiscoding: `_callbackMap` cannot be keyed only by callback: a single callback is valid for multiple subscriptions. For example, `const subscription = service.subscribe(['onClick', 'onDblClick'], callback); subscription.unsubscribe()` overwrites the first wrapper with the second. Both later removals use the wrong listener (and then the original callback after the map entry is deleted), so both DOM listeners remain registered. Please retain the resolved event name and exact wrapper per subscription, and add a regression test for this supported array form.

packages/event-pub-sub/src/eventPubSub.service.ts
```diff
@@ -95,8 +98,9 @@ export class EventPubSubService implements BasePubSubService {
       case 'lowerCaseWithoutOnPrefix':
         if (this.eventNamingStyle === 'lowerCaseWithoutOnPrefix') {
           outputEventName = `${eventNamePrefix}${inputEventName.replace(/^on/, '')}`;
```
ghiscoding: This removes the lowercase conversion from `lowerCaseWithoutOnPrefix`. `getEventNameByNamingConvention('onDblClick', '')` now returns `DblClick`, but the established behavior and the meaning of this style are `dblclick`. The accompanying test was changed to dispatch `DblClick`, which masks the regression. Keep stripping `on` while lowercasing the result.

packages/event-pub-sub/src/eventPubSub.service.ts
```diff
@@ -145,8 +149,13 @@ export class EventPubSubService implements BasePubSubService {
...
-      this._elementSource.addEventListener(eventNameByConvention, (event: CustomEventInit<T>) => callback.call(null, event.detail as T));
-      this._subscribedEvents.push({ name: eventNameByConvention, listener: callback });
+      const wrappedListener = (event: CustomEventInit<T>) => callback.call(null, event.detail as T);
+
+      // store the mapping so unsubscribe can resolve the original callback to the wrapper
+      this._callbackMap.set(callback, wrappedListener);
+
+      this._elementSource.addEventListener(eventNameByConvention, wrappedListener);
+      this._subscribedEvents.push({ name: eventNameByConvention, listener: wrappedListener });
       subscriptions.push(() => this.unsubscribe(eventNameByConvention, callback as never));
```
ghiscoding: `subscribe()` still passes an already-normalized name to public `unsubscribe()`, which normalizes it again. With `camelCaseWithExtraOnPrefix`, `onDblClick` registers `onOnDblClick`; the returned subscription tries to remove `onOnOnDblClick`, leaving the listener attached. `unsubscribeAll()` has the same issue because it passes stored normalized names through `unsubscribe()`. Please use an internal unsubscribe-by-resolved-name/listener helper for subscription closures and bulk cleanup.

ghiscoding (changes requested):
Thanks for addressing the wrapped-listener leak. I verified the simple case, but the current implementation still leaks listeners for valid subscription paths and changes a public naming-style result. Please address the inline comments and add coverage for each case.

ghiscoding:
@rlamtabbet  Thanks for the contribution, I've asked ChatGPT (Codex) to validate the PR and it wrote a few comments above and below. Mostly all comments are actually from ChatGPT, not from me and if you need more explanation then just ask. Thanks

ghiscoding:
Follow-up: here are the concrete regression tests I used locally. The PR’s current tests pass, but these cases fail on the PR head and should be added (or equivalent coverage).

```ts
it('keeps lowerCaseWithoutOnPrefix event names lowercase', () => {
  const callback = vi.fn();
  service.eventNamingStyle = 'lowerCaseWithoutOnPrefix';
  service.subscribeEvent('onDblClick', callback);

  divContainer.dispatchEvent(new CustomEvent('dblclick'));

  expect(callback).toHaveBeenCalledOnce();
});
```

```ts
it('removes every event when a shared callback subscription is unsubscribed', () => {
  const callback = vi.fn();
  const subscription = service.subscribe(['onClick', 'onDblClick'], callback);

  subscription.unsubscribe();
  divContainer.dispatchEvent(new CustomEvent('onClick'));
  divContainer.dispatchEvent(new CustomEvent('onDblClick'));

  expect(callback).not.toHaveBeenCalled();
});
```

```ts
it('removes a subscribe callback with camelCaseWithExtraOnPrefix', () => {
  const callback = vi.fn();
  service.eventNamingStyle = 'camelCaseWithExtraOnPrefix';
  const subscription = service.subscribe('onDblClick', callback);

  subscription.unsubscribe();
  divContainer.dispatchEvent(new CustomEvent('onOnDblClick'));

  expect(callback).not.toHaveBeenCalled();
});
```

I also recommend an equivalent `unsubscribeAll()` test under `camelCaseWithExtraOnPrefix`, because bulk cleanup still passes the already-normalized event name through `unsubscribe()` and double-transforms it.

packages/event-pub-sub/src/eventPubSub.service.ts
rlamtabbet: Will do!

packages/event-pub-sub/src/eventPubSub.service.ts
rlamtabbet: I see the issue now. I'll fix the callbackMap key collision bug by keying on (eventName, callback) pair instead of callback alone. Regression will be added as well. Thx for the review.

packages/event-pub-sub/src/eventPubSub.service.ts
rlamtabbet: Fixed via a skipNamingConversion flag on unsubscribe(). All internal callers pass true since they already hold the converted name.

packages/event-pub-sub/src/eventPubSub.service.ts
rlamtabbet: The map is now keyed by (eventName, callback) instead of callback alone, so the same callback across multiple event names no longer overwrites entries

packages/event-pub-sub/src/eventPubSub.service.spec.ts
rlamtabbet: The test was incorrectly changed to dispatch DblClick; reverted to dblclick and fixed the issue.

ghiscoding:
Follow-up fix pushed in e35044dd8 (`fix(event-pub-sub): simplify listener cleanup`).

The previous `_callbackMap` approach has been removed. The existing `_subscribedEvents` list is now the single source of truth: each wrapped subscription records its original callback, while each returned subscription and `unsubscribeAll()` remove the exact resolved DOM listener through one internal helper. Public `unsubscribe(eventName, callback)` finds and removes every wrapper associated with that callback for the resolved event name.

This avoids stale callback tracking after `unsubscribeAll()`, handles repeated subscriptions of the same callback independently, and keeps internal cleanup from applying naming conversion twice.

Added regression coverage for:
- repeated same-event/same-callback subscriptions;
- direct listener removal after `unsubscribeAll()`;
- bulk cleanup with `camelCaseWithExtraOnPrefix`;
- non-empty prefixes for both corrected naming conventions.

Validated with:
- `pnpm exec vitest run --config ./test/vitest.config.mts packages/event-pub-sub/src/eventPubSub.service.spec.ts` (31 passing)
- `pnpm lint` (one pre-existing unrelated warning in `row-detail-view-plugin`)
- `pnpm build:universal`
- Prettier check on the modified files.

ghiscoding:
Thanks @rlamtabbet for the explanation of the bug, I've asked Codex to do a final review on your code and it found another problem with the approach, so I asked Codex to fix the problem, add more unit tests and to come up with a simpler approach. It replied with the [comment](https://github.com/ghiscoding/slickgrid-universal/pull/2734#issuecomment-5347184963) posted just above and I'll go ahead and merge this update. 

Thanks a lot for your explanation and contribution 😸 
I'll probably push a new release within the next few days or early next week. 

Thanks again ⭐

rlamtabbet:
@ghiscoding It's my pleasure, the codex version is indeed simpler. Thx for the review

PR #2738, opened 2026-08-18
fix(row-detail): support transform rows with overlay rendering

ghiscoding:
### Summary

Make Row Detail compatible with `rowTopOffsetRenderType: 'transform'` which is the default rendering grid option.

Previously, Row Detail panels were inserted inside transformed row elements. Because transformed rows establish their own coordinate system, detail panels could be positioned incorrectly or move with the wrong row offset.

This PR adds an overlay rendering mode that mounts Row Detail panels in a sibling overlay layer outside the transformed row DOM.

### Changes

- Added `rowDetailView.renderMode: 'overlay'`.
- Positioned overlay panels using the grid’s calculated row offsets.
- Added support for scrollable and frozen canvases.
- Preserved framework component lifecycles across React, Vue, Angular, and Aurelia.
- Removed unnecessary microtask scheduling from the overlay lifecycle.
- Updated Row Detail examples and Cypress selectors.
- Updated documentation.
- Added framework-specific CI Cypress scripts.

Usage:

```ts
rowDetailView: {
  renderMode: 'overlay'
}

// this line is no longer needed since it now works with default transform
// rowTopOffsetRenderType: 'top', 
```

### Backward compatibility

- Existing applications continue using inline rendering by default.
- `renderMode: 'inline'` remains available in v10 and is deprecated.
- Explicit `rowTopOffsetRenderType: 'top'` remains supported for legacy compatibility and RowSpan scenarios.

### v11 follow-up

Overlay rendering should become the only Row Detail renderer. The `renderMode` option, inline renderer, related compatibility branches, and Row Detail-specific `rowTopOffsetRenderType: 'top'` fallback can then be removed.

### Verification

- Row Detail unit tests: **54/54 passing**
- TypeScript, Prettier, and diff checks passed.
- Framework Cypress suites were previously validated across Vanilla, React, Angular, and Aurelia.

### LLM

OpenAI GPT-5.6 Terra and Luna (Codex).


packages/row-detail-view-plugin/src/slickRowDetailView.ts
```diff
@@ -580,12 +645,175 @@ export class SlickRowDetailView implements ExternalResource, UniversalRowDetailV
...
+    });
+  }
+
+  /** Get or create the overlay layer for one of the grid's frozen/scrollable canvases. */
+  protected getOverlayHost(canvas: HTMLDivElement): HTMLDivElement {
+    let host = this._overlayHosts.get(canvas);
+    if (!host) {
+      host = createDomElement('div', { className: 'slick-row-detail-overlay' });
+      Object.assign(host.style, {
+        position: 'absolute',
```
ghiscoding: this might be better off moving this to the CSS file

packages/row-detail-view-plugin/src/slickRowDetailView.ts
```diff
@@ -580,12 +645,175 @@ export class SlickRowDetailView implements ExternalResource, UniversalRowDetailV
...
+    const row = this.dataView.getRowById(itemId);
+    if (row !== undefined) {
+      this.updateOverlayPanelPosition(panel, item, row);
+    }
+  }
+
+  /** Keep an existing overlay aligned after row heights or data-view positions change. */
+  protected updateOverlayPanelPosition(panel: HTMLDivElement, item: any, row: number): void {
+    const rowHeight = this.gridOptions.rowHeight || 0;
+    const top = this._grid.getRowTop(row) - this._grid.getFrozenRowOffset(row) + this._grid.getRowHeight(row);
```
ghiscoding: this code logic is duplicated in the file, we should go DRY

packages/row-detail-view-plugin/src/slickRowDetailView.ts
```diff
@@ -580,12 +645,175 @@ export class SlickRowDetailView implements ExternalResource, UniversalRowDetailV
...
+
+    this._expandedRowIds.forEach((itemId) => {
+      if (!this._renderedViewportRowIds.has(itemId)) {
+        return;
+      }
+
+      const item = this.dataView.getItemById(itemId);
+      const row = this.dataView.getRowById(itemId);
+      if (!item || row === undefined) {
+        return;
```
ghiscoding, line 659: we can probably merge the 2 conditions in a single if block to return

packages/row-detail-view-plugin/src/slickRowDetailView.ts
```diff
@@ -580,12 +645,175 @@ export class SlickRowDetailView implements ExternalResource, UniversalRowDetailV
...
+  protected refreshOverlayPanel(item: any): void {
+    if (!this.isOverlayRenderMode || !item) {
+      return;
+    }
+
+    const itemId = item[this._dataViewIdProperty];
+    const panel = this._overlayPanels.get(itemId);
+    if (!panel) {
+      this.renderOverlayPanels();
+      return;
```
ghiscoding, line 726: same as above, we can probably merge the if block return

packages/row-detail-view-plugin/src/slickRowDetailView.ts
```diff
@@ -390,22 +421,56 @@ export class SlickRowDetailView implements ExternalResource, UniversalRowDetailV
...
+      viewComponent?: unknown;
+      preloadComponent?: unknown;
+      viewModel?: unknown;
+      preloadViewModel?: unknown;
+    };
+    // Framework adapters can provide either a component (React/Vue/Angular) or a
+    // view model (Aurelia). In both cases the adapter owns the mounted DOM and
+    // must refresh it after the async response, rather than having the base
+    // plugin remove the overlay before the adapter can mount its view.
+    const usesFrameworkView =
```
ghiscoding: this is framework specific, can we move this out of the base Row Detail? Perhaps adding a new PubSub event?

PR #2748, opened 2026-08-25
feat(selection): add multi-range selection and drag handle options

### LINKED ISSUE #2752: Copying multiple cell ranges does not preserve their grid layout
### Describe the bug

When multiple cell ranges are copied, `SlickCellExternalCopyManager` writes each range as a separate block.

If the ranges share some rows but use different columns, their original row and column positions are lost in the clipboard output.

For example, selecting `A1:A3` and `C2:C3` produces five clipboard rows:

```text
A1
A2
A3
C2
C3
```

The copied text should preserve the selection layout within one bounding rectangle:

```text
A1\t\t
A2\t\tC2
A3\t\tC3
```

This becomes directly visible with the multi-range selection added by #2748.

vscode-mssql currently implements its own selection-to-text conversion for its copy command:

https://github.com/microsoft/vscode-mssql/blob/main/extensions/mssql/src/webviews/common/FluentSlickGrid/fluentSlickGridCopy.ts

```ts
const lines: string[] = [];
for (const range of rangesToProcess) {
    for (let row = range.fromRow; row <= range.toRow; row++) {
        const item = dataView.getItem(row);
        if (!item) {
            continue;
        }

        lines.push(
            visibleColumns
                .slice(range.fromCell, range.toCell + 1)
                .map((column) => {
                    if (!column.field) {
                        return "";
                    }
                    return item[column.field]?.toString() || "";
                })
                .join("\t"),
        );
    }
}
```

### Reproduction

Use Example 37 - Hybrid Selection Model with the changes from #2748:

https://ghiscoding.github.io/slickgrid-universal/#/example37

1. Enable multi-selection.
2. Ctrl/Cmd-drag one range across rows 1-3 in a data column.
3. Ctrl/Cmd-drag another range across rows 2-3 in a different data column.
4. Copy the selection and paste it into a text editor.

Expected: The copied cells keep their relative row and column positions.

Actual: Each selected range is appended as a separate block of rows.

Vanilla / Plain JS

### Environment Info

| Executable | Version |
| --- | --- |
| Slickgrid-Universal | master / PR #2748 |
| Browser(s) | Chromium |
| System OS | Windows / macOS |

ghiscoding:
fixes #2752 and MSSQL https://github.com/microsoft/vscode-mssql/issues/22780

## Summary

Adds Excel-like multi-range selection for cells and rows, along with configurable drag-handle visibility which was done in previous PR #2747 

## Why

Users can select multiple non-contiguous ranges using Ctrl/Cmd-click or Ctrl/Cmd-drag while preserving the existing single-range behavior.

## Changes

- Add `selectionOptions.enableMultiSelection`.
- Support multiple cell and row selection ranges.
- Support Ctrl/Cmd-click toggling and Ctrl/Cmd-drag range additions.
- Preserve partial cell deselection by splitting rectangular ranges when needed.
- Add `SelectionModel.setOptions()` to the public interface.
- Support `selectionOptions.showDragHandle` states:
  - `true` — always visible
  - `'hover'` — visible on hover
  - `false` — disabled
- Update Vanilla, Angular, Aurelia, React, and Vue examples with a checkbox toggle to demo with/without the feature.
- Add documentation and Cypress coverage for cell and row multi-selection.

## Validation

- 538 focused unit tests passed.
- Prettier checks passed.
- Oxlint checks passed.
- `git diff --check` passed.
- Cypress tests updated for Vanilla and all framework examples.

## Comments

Multi-selection requires Ctrl/Cmd interactions, consistent with Excel-style selection behavior.

Related to MSSQL issue: https://github.com/microsoft/vscode-mssql/issues/22780

## AI / LLM assistance

- AI assistance used: 
  - [ ] No 
  - [x] Yes
- If **Yes**: 
  - **which tool/model**: OpenAI Codex / GPT-5.6 Luna
  - **how was it used**: Implemented, refactored, tested, and reviewed the multi-selection and drag-handle functionality.

## Checklist

- [x] The changes are limited to only one scope (if not please explain why in the comments above).
- [x] Tests were added or updated where appropriate.
- [x] Documentation was updated where appropriate.
- [x] Ran a full project build (required when code logic changed)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,12 @@
+<!-- PR title should be using conventional commit with a length below 73 chars. See https://www.conventionalcommits.org/en/v1.0.0/ -->
+
 ## Summary
 
 <!-- Describe the purpose of this pull request and link any related issue. -->
+
+## Why
+
+<!-- Optional reasons explaining why you think this PR can be useful -->
 
 ## Changes
 
@@ -12,7 +18,7 @@
 
 ## Comments
 
-<!-- any other optional comments to add -->
+<!-- Any other optional comments to add -->
 
 ## AI/LLM assistance
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,4 +1,5 @@
 <!-- PR title should be using conventional commit with a length below 73 chars. See https://www.conventionalcommits.org/en/v1.0.0/ -->
+<!-- if this PR fixes an issue, please add it on a separate line for proper linking and following this structure "fixes #123" (replace 123 with the correct issue number) -->
 
 ## Summary
 
@@ -18,18 +19,20 @@
 
 ## Comments
 
-<!-- Any other optional comments to add -->
+<!-- Any other optional comments or website links to add -->
 
-## AI/LLM assistance
+## AI / LLM assistance
 
-- AI/LLM assistance used: [ ] No [ ] Yes
-- If yes, assistant/tool:
-- If yes, model name or model ID:
-- If yes, how it was used:
+- AI / LLM assistance used: 
+  - [ ] No 
+  - [ ] Yes
+- If yes:
+  - which tool/model:
+  - how was it used:
 
 ## Checklist
 
-- [ ] The changes are focused and do not include unrelated work.
+- [ ] The changes are limited to only one scope (if not please explain why in the comments above).
 - [ ] Tests were added or updated where appropriate.
 - [ ] Documentation was updated where appropriate.
-- [ ] Generated output was not edited manually.
+- [ ] Ran a full project build (required when code logic changed)
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -23,12 +23,12 @@
 
 ## AI / LLM assistance
 
-- AI / LLM assistance used: 
-  - [ ] No 
+- AI / LLM assistance used:
+  - [ ] No
   - [ ] Yes
-- If yes:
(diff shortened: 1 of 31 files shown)


ghiscoding:
So this PR is totally worth the extra ~70 loc to the common package, by allowing an optional multi-selections (cells / rows)

@zewa666 since I know you use a lot of cell/row selections, this PR would probably interest you, it came out from a request in the MSSQL extension project. 

@aasimkhan30 I think this second PR fixes both issues that came up in MSSQL, in the previous PR #2747 I added a `showDragHandle` (3-state: `true` (default), `false` or `'hover'` to only show when hovering) and in current PR, we're adding multi-selections via `enableMultiSelection` which also requires the user to hold Ctrl/Cmd to do multi-selections. I think this fixes all the issues brought up in the new grid result, right? Also included in next release is a fix to also allow the selection drag handle to shrink (which wasn't working before) which was fixed in PR #2729... did I cover everything that came up on your side?

```ts
// new grid options related to these recent changes
gridOptions = {
  selectionOptions: {
    enableMultiSelection: true,
    showDragHandle: false, // true (default), false or 'hover'
  }
}
```

Also a side note, I'm not sure if you guys are also supporting RTL in your project, but if so, the next release has better RTL support (through a grid option), it was fixed and improved by an external RTL user of SlickGrid

Any feedback before I merge this later today? I still have to do a deeper review myself of the AI code change

msedge_OeZnuSIC1D

zewa666:
oh thats a great one, been intuitivly always trying it so happy this is now going to work

